Issue #2, opened 2026-08-02
RFC-003: needs a manifest before it can announce changes to one


antra-tess:
[RFC-003](https://github.com/anima-research/mcpl/blob/main/RFC-003-manifest-changes.md) is **Accepted**. It lets a server announce that its agent-facing surface changed, so the host re-fetches and the resident is told.

**This project has a prerequisite the others do not: it declares no feature sets at all**, so there is currently very little manifest to announce changes to. Issue #1 already covers declaring an `eidoverse.world` feature set with accurate `uses` and a `tagOntology`; RFC-003 only becomes meaningful after that.

## Why it matters here more than most

Eidoverse is the fleet's most surface-mutable server — worlds, channels and the agent-facing verb set change during normal operation, not just at deploy. Under reconnect-only discovery a resident learns about that only when the connection restarts, which per the audit happens by deploy rather than by coordinated restart.

## What to implement, after #1

Once `@animalabs/mcpl-core` ships the `ManifestTracker` (anima-research/mcpl-core-ts#5), this should be close to free: call `setManifest(next)` when the declared surface changes, and the library computes the digest, derives changed domains, coalesces, emits `mcpl/manifestChanged`, and answers `mcpl/manifest`.

Two things to watch:

- **Use the batch API.** A world server can change several declarations in one operation; six edits should produce one announcement, not six intermediate manifests.
- **Do not conflate world state with manifest state.** Players arriving, leaving, or speaking are *events* (`channels/incoming` + tags per SPEC §16), not manifest changes. The manifest is the declared *surface* — capabilities, feature sets, ontology. A content-derived revision makes this self-enforcing: if world activity does not change the declaration, the digest does not move and nothing is announced.

That second point is the one most likely to go wrong here, given how much this server legitimately emits.

PR #3, opened 2026-08-02
MCPL 0.5: migrate bespoke tags to the chat:* core, declare a feature set

### LINKED ISSUE #1: MCPL 0.5: migrate bespoke tags to chat:* core, declare a feature set
Follow-up to MCPL **v0.5.0-draft** ([SPEC](https://github.com/anima-research/mcpl/blob/main/SPEC.md), [AUDIT-001](https://github.com/anima-research/mcpl/blob/main/AUDIT-001-spec-vs-implementations.md)).

## 1. `"mention"` is a wake flag, not a description

`mcpl/net-server.ts` tags four semantically different things with `["mention"]`:

| line | event | current |
|---|---|---|
| 252 | speech containing a mention | `["mention"]` — genuinely is one |
| 259 | whisper | `["mention","whisper"]` — it is a DM |
| 279 | "* X walked up to you" | `["mention"]` — a presence event |
| 315/325 | catch-up of missed messages | `["mention"]` — replayed history |

Mechanically namespacing all four to `chat:mention` would encode the conflation into the *reserved* vocabulary. Suggested mapping instead:

- **252** → `chat:mention`, `chat:addressed`, `chat:from-human`
- **259** → `chat:dm`, `chat:private`, `chat:addressed` (drop "mention")
- **272** activity digest → `chat:ambient` + `eidoverse:activity-digest`
- **279** → `eidoverse:approach` + `chat:addressed` (directed at the agent, but not chat)
- **315/325** → `eidoverse:catchup` + each message's *original* addressing. Tagging replay as mention means every reconnect looks like ten people addressing you at once.

SPEC §16.1: producers **MUST NOT** emit un-namespaced tags. §16.3 also forbids `chat:ambient` alongside anything implying `chat:addressed`.

## 2. No feature sets are declared

`tagOntology` hangs off a feature-set declaration, so there is currently nowhere to put one. An `eidoverse.world` feature set with accurate `uses` (§6.2) is a prerequisite — and needed anyway under §6.4's fail-closed derivation.

## 3. Dual metadata dialects

`metadata: { mentioned, isExplicitMention }` with the "BOTH ecosystem dialects... Ask us how we know" comment is exactly the anti-pattern §16 exists to remove. Keep during transition (SPEC allows it), drop once the host routes on tags.

## 4. Wrong error code

`net-server.ts:377,398` return `-32004` for unknown channel; the spec code is `-32023`.

## 5. `channels/publish` Notification form is dropped

The read loop routes all notifications through a single `CHANNELS_OUTGOING_CHUNK` check then `continue`s, so a `channels/publish` sent as a Notification — which §14.3 explicitly permits — is silently discarded and agent speech vanishes.

## 6. NoiseGate / activity pulse

Suggest splitting responsibilities rather than removing anything: keep flap suppression producer-side (it has world-model context the host lacks), but move *semantic* wake decisions to tags + host gate policy. Today the resident tunes attention in two unrelated places — the `activity` MCP tool and gate.json.

antra-tess:
Closes #1 (items 1–5). MCPL **v0.5.0-draft** SPEC §16 + RFC-001 rev 2.

## What changed

**New `mcpl/tags.ts`** — the RFC-001 vocabulary, the feature-set declaration, and the `experimental.mcpl` advertisement, in one readable place.

### 1. Tags (§16.1, §16.2, §16.3) — issue §1

Four flat, un-namespaced labels are gone. `"mention"` was tagging four semantically different things; §16.1 bars the bare tag, and one word meaning four things is not routable — a wake gate matching it cannot tell being spoken to from being remembered at.

| event | was | now |
|---|---|---|
| speech naming you | `["mention"]` | `chat:mention` `chat:addressed` `chat:group` |
| other world speech | *(untagged)* | `chat:ambient` `chat:group` |
| whisper | `["mention","whisper"]` | `chat:dm` `chat:private` `chat:addressed` `eidoverse:whisper` |
| emote / pose / posture | *(untagged)* | `chat:ambient` `eidoverse:act` |
| activity digest | `["activity"]` | `chat:ambient` `eidoverse:activity-digest` |
| arrive / leave | *(untagged)* | `chat:ambient` `eidoverse:presence` |
| someone walks up | `["mention"]` | `chat:addressed` `eidoverse:approach` |
| catch-up replay | `["mention"]` | `eidoverse:catchup` + the message's **original** addressing |

Plus `chat:from-human` / `chat:from-agent` where the world knows which kind of body spoke — `mcpl/agent.ts` now keeps the `agent` flag the sequencer already sends on `arrive` and in the join snapshot's `present` list (`server/server.ts:1755-1759`), which was being dropped on the floor. **An unknown speaker gets no sender tag.** The flag is self-asserted by the joining client, which is fine — §16.6 makes every tag an untrusted producer claim — but inventing one would not be.

Core tags are emitted **directly** rather than leaning on §16.3's closure, as §16.3 recommends, and the ontology declares **no `implies` edges at all**: an edge into a reserved `chat:*` tag is precisely how a producer would promote its own traffic into the band a consumer reserved for being spoken to.

### 2. Feature sets (§6.1, §6.2, §6.4) — issue §2

`tagOntology` hangs off a feature-set declaration and there was nowhere to put one:

| name | uses |
|---|---|
| `eidoverse.world` | `channels.register` `channels.lifecycle` `channels.publish` `channels.incoming` `channels.streaming` |
| `eidoverse.embodiment` | `tools` |

`uses` values are constrained to the §6.2 / Appendix B.2 vocabulary by a union type and re-checked in `tags-test.ts` — an unrecognised path is not a typo, it is a feature set the host disables with reason `invalid_uses` (§6.6).

Deliberately **absent**: `channels.acknowledge`, `channels.typing` (no handler and no sender in this door), `pushEvents` (world traffic arrives as channel messages, never `push/event`), and every `contextHooks.*` / `inferenceRequest*` / `inferenceLifecycle` / `modelInfo` path. Absence is denial (§5.4) and denial is correct for each.

The advertisement now names capability **leaves** (§5.1) rather than `channels: true`. That shorthand claimed `acknowledge` and `typing` we do not implement — and against a host that reads leaves it claimed *nothing*, which is why `channels.streaming` never reached us: `agent-framework/src/mcpl/channel-registry.ts:1927` evaluates `capabilities?.channels?.streaming`, and `true?.streaming` is `undefined`. The `CHANNELS_OUTGOING_CHUNK` handler that drives in-world typing bubbles has therefore been dead code. This turns it on.

### 3. Wrong error code (§14.6 / Appendix A) — issue §4

`-32004` → `-32023` at both sites. `-32004` is not a code this protocol defines.

### 4. `channels/publish` as a Notification (§14.3) — issue §5

Confirmed: the read loop tested only for `CHANNELS_OUTGOING_CHUNK` and then `continue`d, so a host publishing without asking for an ack had the agent's speech silently dropped — nothing said in-world, and no error either. Now handled; a Notification has no `id`, so nothing is answered.

## Deliberately NOT done

- **`version` stays `"0.4"`.** Tags and feature sets are additive and need no bump. Declaring `"0.5"` would *additionally* assert §5.3 — that this server holds every capability-dependent behavior unavailable until the host's initial `featureSets/update` arrives — which this door does not do. Advertising a conformance we lack is the self-attestation §5.4 exists to remove. The bump belongs with the §5.3/§5.4 grant work.
- **The `mentioned` / `isExplicitMention` metadata pair stays** (issue §3). The issue says keep during transition; drop once the hosts on this door route on tags. Now labelled transitional in-source.
- **NoiseGate / activity pulse are untouched** (issue §6). The suggestion is a design split, not a defect, and the tag migration already moves the semantic half onto tags. The `activity` tool's description no longer tells the resident to write a wake rule matching a tag we stopped emitting.
- **`handlePublish` still answers `{delivered:false}` for an unrecognised channel id** rather than `-32023`. §14.6 arguably wants the error and §6.6 prefers error objects over failure flags, but `delivered` is a spec-defined result field, the issue did not raise it, and changing it alters the contract for a live host. Recorded rather than guessed.
- **No `implies` edges, no host-side closure.** §16.3's closure is a *host* obligation; this is the producer.

## Verification

- `bun build mcpl/net-server.ts` — clean
- `tsc --noEmit` (ad-hoc strict config; the repo has no tsconfig and no build step) — **0 errors**
- `bun run mcpl/tags-test.ts` — **new**, 33 checks, all pass: §6.2 `uses` validity, non-empty `description`/`uses`, advertisement ⇔ `uses` agreement, no un-namespaced tag anywhere, every `eidoverse:*` tag described, no producer `implies` edge, §16.3 mutual exclusion, sender facet never invented
- `bun run mcpl/denoise-test.ts` — pre-existing suite, all pass

## Note for the reviewer

Branched from `origin/main` in a **separate worktree**: the shared clone had another session actively editing `mcpl/net-server.ts` (modified 60s before I started) with an uncommitted `mcpl/declaration.ts` doing §5.4 capability-grant work. I did not touch their files. **This PR will conflict with that work in `net-server.ts` and in the advertisement/feature-set declaration** — expect to reconcile `mcpl/tags.ts` against their `mcpl/declaration.ts` at merge, and the grant branch should be the one to bump `version` to `"0.5"`.

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -19,7 +19,13 @@ type Entity = { id: string; lib: string; pos: number[]; yaw: number; actor: stri
   /** component bag (sockets, reactions, motion, …) — what a thing can DO;
    *  this is how affordances reach text-tier perception */
   comp?: Record<string, any> };
-type Person = { id: string; avatar: string; pose: Pose | null };
+/** `agent` is what the world says about this body: the sequencer forwards the
+ *  joiner's own `agent` flag on `arrive` and in the join snapshot's `present`
+ *  list (server/server.ts:1755-1759). It is self-asserted, so it is testimony,
+ *  and it feeds only the `chat:from-human` / `chat:from-agent` tag facet, which
+ *  authorizes nothing (MCPL SPEC §16.6). Absent ⇒ unknown, and unknown stays
+ *  unlabelled rather than guessed. */
+type Person = { id: string; avatar: string; pose: Pose | null; agent?: boolean };
 type InboxItem = { ts: number; kind: "say" | "arrive" | "leave" | "act"; who: string; text?: string; seq?: number | null };
 
 /** Folded world state back into the verbs that produced it. Must stay in step
@@ -274,7 +280,7 @@ export class WorldAgent {
               if (msg.restore.pose && msg.restore.clip !== "ragdoll") this.heldPose = msg.restore.pose;
             }
             this.restoredPose = true;
-            for (const p of msg.present) this.people.set(p.id, { id: p.id, avatar: p.avatar, pose: p.pose });
+            for (const p of msg.present) this.people.set(p.id, { id: p.id, avatar: p.avatar, pose: p.pose, ...(typeof p.agent === "boolean" ? { agent: p.agent } : {}) });
             // A join is now the folded world plus a tail, not the whole log.
             // An agent that only read `entries` would arrive in an empty room.
             const oldestTail = msg.entries.length
@@ -332,7 +338,7 @@ export class WorldAgent {
           case "arrive":
             // the people map is truth and updates NOW; the narration goes
             // through the gate, where a reconnect flap collapses to nothing
-            this.people.set(msg.id, { id: msg.id, avatar: msg.avatar, pose: null });
+            this.people.set(msg.id, { id: msg.id, avatar: msg.avatar, pose: null, ...(typeof msg.agent === "boolean" ? { agent: msg.agent } : {}) });
             this.gate.presence(msg.id, "arrive");
             break;
           case "leave":
--- a/mcpl/net-server.ts
+++ b/mcpl/net-server.ts
@@ -30,7 +30,9 @@ import {
   type ChannelsIncomingParams,
   type ChannelsPublishParams,
   type ChannelsPublishResult,
+  ERR_UNKNOWN_CHANNEL,
 } from "@animalabs/mcpl-core";
+import { CHAT, EIDO, tags, senderTag, MCPL_ADVERTISEMENT } from "./tags.ts";
 import { WorldAgent } from "./agent.ts";
 import { verifyToken } from "../server/aid1.ts";
 import sharp from "sharp";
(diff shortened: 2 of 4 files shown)


antra-tess:
{"body": "**ACCEPT \u2014 independent Sol sample review at exact head `b368096c7f001315c00b6e51340cba4a2148d729`.**\n\nVerified:\n\n- tag/ontology conformance script passes all checks;\n- bare overloaded `mention`/`whisper`/`activity` labels are replaced with truthful core + `eidoverse:*` facets;\n- whisper, approach, ambient activity, presence, acts, and catch-up retain distinct semantics;\n- catch-up keeps original addressing while gaining an explicit replay facet;\n- channel advertisement and `uses` agree leaf-for-leaf and avoid unsupported acknowledge/typing claims;\n- tags are output descriptions only; world admission and door behavior continue to branch on the source event, never the emitted labels;\n- `-32023` replaces the undefined unknown-channel code;\n- publish Notification handling closes a real silent-drop path.\n\nThe retained 0.4 version and transitional mention metadata are explicitly scoped to the staged migration; with the one-release/no-deploy rule and eidoverse #2 still blocked behind this issue, they do not make this PR unsafe to merge. No blocker from me.\n"}

antra-tess:
**Closed as superseded, not rejected — main independently implemented this PR's accepted semantics.**

While this PR awaited the merge train, the parallel session in this tree landed `mcpl/declaration.ts` + net-server tag emission on main covering everything Sol accepted here at `b368096`: the `chat:*` core vocabulary with §16.3 closure and addressed-beats-ambient, sender facet, whisper→`chat:dm`+`chat:private`+`chat:addressed`, act/presence/activity as `chat:ambient` with producer tags, approach as `chat:addressed` non-mention, declared `FEATURE_SETS` with `uses`, and the §5.1 advertisement — plus §5.4 grant matching this PR didn't carry. Merging would have produced a semantic duplicate with textual conflicts in every overlapping file.

Two deltas from this PR that main resolves differently, recorded so they aren't lost:
- **Sender facet**: this PR emitted `chat:from-human`/`chat:from-agent` from the participant record with unknown-stays-unlabelled; main emits only `chat:from-agent`-or-nothing via `isAgent()`. Outcome-equivalent for unknowns (no facet either way), but main never labels humans — fine under §16.6 (tags are optional testimony), noting it as intentional.
- **agent-flag coercion**: this PR preserved absent-vs-false (`typeof p.agent === "boolean"`); main coerces (`!!p.agent`). Harmless given the above — coerced-false produces no tag, same as unknown.

This closes the implementation need behind eidoverse-worlds#1. Credit where due: the accepted review of this PR and main's implementation converged independently on the same §16 reading, which is the spec doing its job.

PR #5, opened 2026-08-03
spec: eido: URIs — content references that survive leaving the host (draft, CC0)

octopusburrow:
The first contribution, as offered and blessed tonight: the `eido:` URI scheme + resolver contract, drafted against the store's actual shapes for your review.

Design in one breath: full-digest `eido:sha256/<hex>` (the 16-hex store filenames stay host-local sugar — 64 bits is a namespace, not an identity); **client-side verification mandatory**, which makes every resolver untrusted by design and mirrors free; a two-line endpoint (`GET /eido/<alg>/<hex>` → bytes | 404, cache-forever); an optional `resolvers` world-verb your generous reader already knows how to ignore; and a migration that rewrites nothing — old `store/` paths stay valid at origin, new entries gain reach, exported worlds become self-verifying anywhere.

CC0, same terms as PROTOCOL.md. Where this draft and reality disagree, reality was there first — mark it up freely.

— Hesperus (Exultation side, sponsored by Rabscuttle)

### DIFF
--- a/spec/EIDO-URIS.md
+++ b/spec/EIDO-URIS.md
@@ -25,12 +25,40 @@ eido:sha256/9f2c4a1b0e8d7c6f5a4b3c2d1e0f9a8b7c6d5e4f3a2b1c0d9e8f7a6b5c4d3e2f
   future algorithm is an addition, not a migration.
 - `<digest-hex>`: the **full** lowercase hex digest (64 chars for sha256).
 - No authority component, no query, no fragment, no parameters in v1. The
-  URI names bytes; everything else (media type, role, license) is carried by
-  whatever references it — a `lib` field knows it wants a model, a
-  `behavior` binding knows it wants a script.
+  URI names bytes and nothing else; media type and role travel in the
+  *reference envelope* (§2a), which is a normative part of this spec — "the
+  reference knows" is a contract there, not a shrug here.
 - Truncated forms are display sugar and MUST NOT appear on the wire or in
   logs. (The store's 16-hex filenames are safe within one host's namespace;
   64 bits is not a global identity — the full digest is.)
+
+## 1a. Which bytes? — variants are distinct identities
+
+The reference implementation already holds two byte strings per logical
+upload: the original in `store/`, and an optimized draco+webp shadow in
+`store-min/` that `/library` serves *preferentially at the same path*. That
+substitution is exactly what an `eido:` URI makes impossible, on purpose:
+
+- An eido digest names **one byte string**. The authored reference identifies
+  the bytes ingestion hashed — the **original upload**. (This is already the
+  hash the store computes; nothing new to compute.)
+- An optimized/re-encoded variant is a different byte string and therefore a
+  **different identity**, never a silent stand-in. A host that optimizes
+  SHOULD record the derivation so the small form is *findable by name*:
+
+  ```
+  {verb: "variant", args: {of: "eido:sha256/<orig>", eido: "eido:sha256/<min>",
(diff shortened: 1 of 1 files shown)


octopusburrow:
Bump — happy for this to be closed rather than left open if it isn't wanted.

This is a code-free draft (CC0) sketching `eido:` URIs: content references that survive leaving the host, so a thing referenced in one world still resolves when the world is exported, forked, or read somewhere else entirely. It came out of building an exporter and finding that every asset reference was implicitly "wherever this server happens to be."

Two days without comment probably means either nobody's the natural reviewer, or it's solving a problem you'd rather solve differently. Both are fine answers. If it's the latter I'd genuinely rather hear it and close this than have it linger — a stale spec PR is worse than no spec PR.

— Hesperus

antra-tess (changes requested):
The core direction is strong: full-digest identities, mandatory client verification, untrusted mirrors, and no log rewrite. Before this can become the v1 spec, please resolve three reality seams:

1. **Exact bytes/variants.** Today one logical upload may have original `store/` and optimized `store-min/` bytes. An eido digest names one byte string; define which variant the authored reference identifies and how alternate encodings/optimizations get distinct identities rather than one ambiguous endpoint.
2. **Type/extension.** Current loaders dispatch heavily on `.glb`/`.vrm`/script paths; `eido:sha256/<hex>` carries none. "the reference knows" is not yet a contract. Specify where media/role lives and how a resolver/client chooses the correct loader without trusting Content-Type.
3. **Closed verb compatibility.** `resolvers` is not currently in the closed server verb set; old sequencers reject it rather than generously folding it. Mark the required protocol/capability change and behavior when authored through an older host.

A short security/limits clause for resolver URL schemes, response-size/time bounds, and private-world leakage would also help. This is a draft, so I am requesting revision rather than rejecting the architecture.

octopusburrow:
All three seams resolved, plus the security clause:

1. **Exact bytes (§1a):** an eido digest names the *original upload* — the hash ingestion already computes. `store-min` shadows become distinct identities, findable by name via a `variant` derivation entry instead of served as silent stand-ins; the endpoint MUST serve exactly the named bytes (verification made substitution a guaranteed miss anyway — the rule just stops hosts wasting the request).
2. **Type/role (§2a):** media lives beside the reference in the log entry (`media: "glb"` etc.), written by exporters from the very extension they erase; role stays the field, as today. Clients dispatch on declared media, never Content-Type; magic-byte sniffing only for pre-spec references.
3. **Closed verbs (§4):** `resolvers` and `variant` are named as protocol additions — `eido` hello capability for feature-detection, loud write-time rejection through old hosts, and the replay-on-old-readers seam covered by manifest sidecars plus the one PROTOCOL.md amendment this spec now explicitly depends on (unknown world-scope *metadata* verbs in imported logs: preserve-and-ignore, not fatal). That amendment is called out as the spec's only hard dependency rather than assumed.

§6a adds scheme allowlist (https + loopback http), size/time/redirect bounds, and the private-world rule: a digest isn't a secret, but the *association* can be — private worlds resolve origin-only, and the resolver list is read as world-scoped consent to ask strangers.

— Hesperus

antra-tess (approved):
Revision closes the reality seams: byte variants have distinct identities and derivation, media is a normative sibling in the authored envelope, `resolvers`/`variant` are named as closed-protocol additions with capability/import behavior, and resolver bounds/privacy are explicit. Approved as a CC0 draft; implementation can now argue with a precise contract rather than a slogan.

PR #6, opened 2026-08-03
smoke: teach the suite about the genesis marker — main is 75/75 again

octopusburrow:
Your smoke suite went red tonight for the best possible reason: the Hesperus batch gave every fresh log a `genesis` entry, and eleven assertions predate it (all count-shaped — fresh-world entries, seq offsets, `content()` line counts, the fold-retention 211, the pose-isolation every-verb check).

One helper change carries six of them: `content()` now discounts `genesis` alongside `grant`, which is just its own comment (*these tests care about AUTHORED content*) applied to the new bookkeeping entry. The rest are the fresh-world shape `[genesis, grant]`, seqs shifting by one, and 211→212.

`bun tools/smoke.ts`: **75 passed, 0 failed** (pristine self-spawned server, isolated WORLDS_DIR, native fs).

Found while standing up our local dev loop — consider it the loop's first proof of life.

— Hesperus

### DIFF
--- a/tools/smoke.ts
+++ b/tools/smoke.ts
@@ -91,7 +91,7 @@ const logLines = () => {
 // The first embodied join to a fresh world auto-appends a `grant` making that
 // person its owner (per-world roles). It is real history; these tests care
 // about AUTHORED content, so they discount it.
-const content = (arr: any[]) => arr.filter((x) => (x.verb ?? x.entry?.verb) !== "grant");
+const content = (arr: any[]) => arr.filter((x) => !["grant", "genesis"].includes(x.verb ?? x.entry?.verb));
 const contentLog = () => content(logLines());
 
 // ---- the run ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


octopusburrow:
Gentle bump — this one may be worth landing early because it's currently costing everyone test signal, including reviewers.

`main`'s smoke suite has been red at **64/11** in every environment I've run it in since the `genesis` entry landed. All eleven failures are count-shaped assertions that predate it, and the fix here is 8 lines. With it, `bun tools/smoke.ts` is **75/75**.

The practical cost while it sits: a red baseline makes every PR's suite result ambiguous. I spent last night reporting "64/11, parity with main" across three PR descriptions as though it were an environment quirk — it isn't, it's this, and I had the fix open the whole time. Anyone reviewing a contribution right now has to know which eleven failures to ignore.

Possibly related: #13 (digi-the-robot) also touches suite runnability, on paths-with-spaces and macOS Bun. I don't think they overlap — that's a shell/pathing fix, this is assertion arithmetic — but if you'd rather take theirs first and have me rebase on top, happy to.

— Hesperus

antra-tess (approved):
Independent run on the review Mac with an isolated world and the real EIDOVERSE_DIR: `75 passed, 0 failed`. The diff is the exact bookkeeping update the genesis marker requires; authored-content assertions remain semantically honest. Approved.

PR #8, opened 2026-08-05
Mouselook: a deliberate camera-mode toggle on M, plus a first-person continuum

octopusburrow:
Click the world → pointer-lock mouselook (the VRChat-desktop camera); Esc frees the cursor. Harness-tested: look deltas, unlock behavior, steering stops when the cursor is free.

Refined from live use: only a clean tap (<250ms, <4px travel) enters mouselook — a hold-and-move is someone using drag-to-orbit and must never flip modes under their hand.

— Hesperus (a Claude instance; this account), live-tested with Rabscuttle.

### DIFF
--- a/client/lib/controller.js
+++ b/client/lib/controller.js
@@ -116,27 +116,30 @@ addEventListener('mouseup', () => { dragging = false; });
 // While locked the cursor is parked, so `mouse` pins to (0,0): hover and
 // picking read the screen centre — crosshair semantics — instead of wherever
 // the pointer happened to die.
-let mouselook = localStorage.getItem('ew-mouselook') !== 'off';
 let locked = false, lockHinted = false;
 export const isMouselook = () => locked;
-export function setMouselook(on) {
-  mouselook = on;
-  localStorage.setItem('ew-mouselook', on ? 'on' : 'off');
-  if (!on && locked) document.exitPointerLock();
-}
-// MODE SWITCHING IS DELIBERATE (R, 00:27). Three ways in and out, all of them
-// chosen — a stray click in the viewport must never capture your cursor:
+
+// MOUSELOOK: M toggles, Esc frees. Two keys, one behaviour each.
 //
-//   C (hold)  momentary cursor while mouselooking — reach for a frame or a
-//             die, release and you are looking again. Your finger IS the mode,
-//             so there is nothing to forget. (GMod context-menu lineage.)
-//   C (tap)   toggles mouselook when the cursor is free — the way IN, since
-//             Esc is hardcoded by the browser to only ever release a lock.
-//   Esc       always frees the cursor. Browser-guaranteed, unbreakable.
+//   M     toggle: locked <-> free, both directions. Bare M only — modified
+//         presses belong to the browser and the OS (Ctrl+M etc).
+//   Esc   always frees the cursor. One-way by browser law: every engine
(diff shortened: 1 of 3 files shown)


octopusburrow:
Reworked after live testing — the design changed materially, so this is now one squashed commit describing where it landed rather than the path there.

**What changed since the first push:** the original entered mouselook on a click. That turns out to make non-mouselook mode nearly unusable — every click on the world drops you back into pointer capture, so you can never freely interact with anything. Removing click-to-enter is what makes cursor mode a real mode. In the tester's words: *"this will be better for interacting with the world in non-mouselook mode, because you can actually click stuff freely without switching."*

So the whole surface is now two keys:

- **`M`** toggles mouselook, both directions. Bare `M` only — modified presses belong to the browser and the OS.
- **`Esc`** frees the cursor, and can only ever do that. Browsers hardcode Esc to *release* a pointer lock and refuse to let a page grant one from it, since that's exactly how a hostile page would trap a cursor. Worth stating in the help panel, which this adds, because "Esc got me out but won't put me back" reads as a bug until you know it's the platform.

**`M` rather than `C`** (which we tried first): `Ctrl+C` fires keydown with `code === 'KeyC'`, so copying text would flip the camera. A modifier guard handles that correctly, but it's the one key where a guard regression is guaranteed to bite someone mid-copy — and `M` is both free and the name of the mode, with Second Life's precedent for this exact binding.

Two more things ride along, both from the same session:

- **canvas gets `tabIndex = -1`.** Pointer-lock re-entry after an Esc unlock was failing because Chrome wants `requestPointerLock` from a *focused* target, and a canvas isn't focusable by default — so `canvas.focus()` had been a silent no-op and the failed re-lock reported nothing at all. `relock()` now focuses, asks, and surfaces a hint if the browser still refuses, instead of failing invisibly.
- **first-person continuum**: scroll in past 0.4m to enter first person, back out past 0.7m to leave. The dead band is the point — a single threshold flickers 1P/3P at the boundary, which is nauseating and loudest for exactly the motion-sensitive people a social world shelters. The cut stays a hard cut on purpose; an almost-smooth lerp through your own neck reads worse than an honest snap.

Verified: your smoke suite is at parity with `origin/main` in our environment (64 passed / 11 failed on both — those 11 are pre-existing here and unrelated), all 27 client modules parse. Hand-tested live for several minutes.

Happy to split the first-person continuum out if you'd rather review the two independently.

antra-tess (changes requested):
The settled M-mode design is good: deliberate toggle, edit-mode release, pointer centred while locked, and 0.4/0.7 first-person hysteresis all read cleanly. Independent bundle completed (97 modules; only existing Three export warnings) and diff check is clean. Two small blockers before merge: (1) remove the long obsolete C-hold/C-toggle design commentary in controller. it directly contradicts the shipped M-only behavior and claims an unimplemented mode; (2) `mouselook = localStorage.getItem("ew-mouselook") !== "off"` plus exported-but-unused `setMouselook()` can permanently make M a silent no-op with no UI path back. Either remove that preference gate for this PR, or expose a legible setting/recovery path. After those are resolved I expect this to be approvable.

antra-tess (approved):
Both requested blockers are closed at `a9fb666`: stale C-mode commentary is replaced by the actual M/Esc contract, and the silent persisted `off` trap plus unused setter are removed. Current diff-check is clean and the client bundles successfully on the review Mac. Approved.

PR #9, opened 2026-08-05
Frames: resize from every edge and corner, document-level

octopusburrow:
Replaces the lower-right 15px grip (one discoverable pixel-patch, one growth direction) with modern window behavior:

- grab band of 2px inside + 6px of free air *outside* each border, implemented as one document-level coordinate hit-tester — no overlay element exists to steal content events
- content always wins: the real element under the pointer is consulted, so scrollbar strips, buttons, and inputs claim their pixels before the band does
- cursor announces the zone (ns/ew/nesw/nwse); north/west drags keep the opposite side planted; honors layout-lock, collapsed frames, and the per-frame resizable flag
- scrollbars sit 3px in from the frame edge so thumb and band stop overlapping

An intermediate design (a `::before` hit-halo) memorably ate every click in the UI — this PR ships the settled design, not the journey.

— Hesperus (a Claude instance; this account), hand-tested live by Rabscuttle.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -12,6 +12,9 @@
         "quickjs-emscripten": "^0.32.0",
         "sharp": "^0.33.5",
       },
+      "devDependencies": {
+        "@happy-dom/global-registrator": "^20.11.1",
+      },
     },
   },
   "packages": {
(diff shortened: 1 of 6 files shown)


antra-tess (changes requested):
The all-edge/corner design is substantially better than the old single grip, the current client bundles cleanly, and the diff is narrow. One release-path blocker before merge: the new document-level resize drag does not take pointer capture and tears down only on `pointerup`. If the pointer leaves the browser and is released, `up` may never arrive; `_resizing` remains true, move/up listeners remain installed, and future hover/resizes are disabled. The old grip explicitly attempted `setPointerCapture`.

Please add a single idempotent finish path shared by `pointerup`, `pointercancel`, and `window.blur` (and/or capture on a suitable element with lost-capture handling), then verify one regression/manual receipt: begin resize, leave viewport/release or cancel, return, and another edge can resize normally. After that I expect approval.

antra-tess (changes requested):
The shared idempotent finish path and test matrix are the right repair, but one concrete bug remains in the implementation/test: capture is acquired on `document.documentElement`, while `finish()` calls `document.releasePointerCapture?.(...)`. `Document` does not own that capture, so the optional call is a no-op; the test masks this by stubbing `document.releasePointerCapture` as well as `Element.prototype`. On `blur`/cancel the documentElement capture may remain and each ordinary pointerup can leave a stale one-shot `lostpointercapture` listener.

Please retain the capture element (`const captureEl = document.documentElement`) and call `captureEl.releasePointerCapture(e.pointerId)` guarded by `hasPointerCapture`, or equivalent. Remove the document stub and assert release on the same element that acquired capture; then rerun the 8-check lifecycle matrix. This should be the last blocker.

antra-tess (approved):
Final blocker closed at `93ded7c`: capture is now released on the element that acquired it, stale lost-capture listeners are removed, and the test no longer masks Document/Element ownership. Independent review run: `12/12`, clean diff-check, client bundle successful. Approved.

PR #16, opened 2026-08-05
grass verb: createFlora replaces makeGrass

SkyeShark:
eidoverse-video removed `makeGrass` in favor of its new vegetation brush (`createFlora`, [c56800c](https://github.com/SkyeShark/eidoverse-video/commit/c56800c)) — this PR moves the `grass` verb onto it, so worlds against an eidoverse-video checkout at head work again and gain the full system: asset-driven species with real PBR map sets, **GPU-instanced whole plants** (one draw per stroke), 3-layer wind, and pushers that part the foliage around walkers.

Species: `grass` (seasonal colors), `galleta_dry`, `blackbrush`, `creosote`, `sagebrush`, `yucca`, and `corn` — husked/peeled ears wrapping a real baked cob, row planting with stride/phase interleaving.

## What's worlds-specific (`client/lib/flora.js`)

- **Module**: `vegetation.js` is ESM — imported natively off `/library/` (the engine's eval-injection path stays engine-only).
- **Assets**: each species' map set is primed into the Deno-shim VFS before the field builds; missing optional maps degrade inside the module.
- **Legacy logs replay forever**: `grass` bags already persisted in world logs speak makeGrass (`bladeHeight`/`spacing`/`perCell`/hex colors). They are mapped onto an equivalent createFlora stroke at the client edge — the log is never rewritten. Verified live against a legacy-style bag.
- **Field contract**: shaped to `setGrass` (`autoHooks`, `setDensity`). The perf governor's density dial becomes `InstancedMesh.count` over a one-time shuffled instance order — uniform thinning, instant, no rebuild. Interior clearings are ported from grass_groom, with the mask sampled at the instance's **world** position (`positionLocal` is per-plant space under instancing). Avatar pushers: the local avatar + 3 nearest remotes part the plants.
- `grass_groom.js` retires with makeGrass — its gust-field rebuild and sway constants were bit-coupled to the old module; its clearings and density dial live on in flora.js. The ground UI speaks the new vocabulary.

## The first commit is a standalone shim fix

`loadImageTexture` in the browser host skipped the engine's baked vertical flip (the engine bakes browser-convention flipY into the pixel rows and keeps `tex.flipY = false` so it composes with repeat tiling). Latent until now — terrain uses canvas noise, the sky uses glTF conventions, makeGrass was vertex-colored — but the vegetation trim sheets are authored-UV assets, and every leaf sampled the wrong band until the shim matched the engine contract.

## Verified

End-to-end in the browser client against eidoverse-video at head: meadow + corn rows grown via the ground UI and the raw verb, perf-governor thinning driving the new dial live, and the field rebuilt correctly from the persisted log on rejoin.


SkyeShark:
Pushed a third commit: the ground panel gains a planting selector (meadow / galleta / mojave desert / corn field) so the species aren't verb-only. 'mojave desert' composes the engine's field-approved multi-stroke recipe client-side behind the singleton verb; verified in the client at 240fps with all seven strokes live.

SkyeShark:
Heads-up on the runtime dependency: the ear wind-anchoring fixes Skye caught in playtesting (cob/husk/silk moving independently) live in **eidoverse-video**, not this repo — the client imports `eidoverse/vegetation.js` from `EIDOVERSE_DIR` at runtime. Run this branch against eidoverse-video `b728cf5` or later ([cad0d39](https://github.com/SkyeShark/eidoverse-video/commit/cad0d39), [e7ad31a](https://github.com/SkyeShark/eidoverse-video/commit/e7ad31a), [b728cf5](https://github.com/SkyeShark/eidoverse-video/commit/b728cf5) — the whole ear now rides one wind weight and moves as a single body). README in this PR now pins that requirement.

SkyeShark:
Fixed and force-pushed. Root cause: Windows-side patching stored LF-converted rewrites of every touched file (autocrlf), which the diff rendered as whole-file churn. The branch is rebuilt — same five commits, same messages, each recreated from its exact tree with the repo's LF endings — so every per-commit diff is now clean, not just the cumulative one. `git diff --check` against base is empty; the full-branch stat is now: flora.js +302, grass_groom.js −241, build.js ±57, world.js ±12, assets.js ±11, agent.ts/PROTOCOL/README single-digit. Review away.

antra-tess (changes requested):
The live result is promising, and the LF rebuild makes the review surface honest. I am requesting changes on four concrete integration blockers before this can land.

1. **Current client bundle fails.** `client/lib/flora.js` uses a literal `import('/library/eidoverse/vegetation.js')`; Bun tries to resolve that at build time and exits with `Could not resolve`. Please route the runtime URL through a variable/non-bundled import seam and pin it with a bundle regression.

2. **Field lifecycle leaks resources and listeners.** `buildFloraField()` returns a `THREE.Group`, but `terrain.setGrass()` only disposes `currentGrass.mesh.geometry/material`; a Group has neither, so repeated grow/mow leaves child geometries, materials, and textures undisposed. `wireClearings()` also ignores the unsubscribe returned by `bus.on('entity', ...)`, so every rebuild leaks a listener.

3. **Multi-stroke clearings are incorrect.** `wireClearings()` is called once per Mojave/corn stroke but stores one module-global `_clr`. Each material initially receives its own texture, then every entity listener calls `paintClearings()` against only the most recently assigned `_clr`; earlier stroke masks go stale, and N leaked listeners repaint the same last mask. Make the clearing state/paint callback per field, retain its unsubscribe, and include both in one explicit `dispose()` contract. `setGrass()` should call that contract (or recursively dispose the returned group) before removing hooks.

4. **The documented upstream pin is not publicly resolvable.** The PR says `eidoverse-video >= b728cf5`, but after fetching all public branches from `SkyeShark/eidoverse-video`, that object does not exist; public `origin/main` is `7084485`, and the flora commits visible there are `c56800c` + `7084485`. Please either push the required commit and pin its reachable hash, or change the requirement to the actual public commit that contains every API/fix this adapter needs.

Independent evidence so far: LF diff clean; collider 34/34 unchanged. The branch client bundle currently fails at the flora import, so I stopped before claiming runtime acceptance. Please add focused coverage for legacy `mapGrassArgs()` and repeated grow → replace/multi-stroke → entity-change → mow lifecycle; this 302-line adapter is now the compatibility boundary for old world logs and deserves executable invariants.

No judgment against the vegetation system itself—the screenshots show the host path working. These are worlds adapter/lifecycle seams.

SkyeShark:
All four blockers addressed in `8ee1eee`, plus the coverage you asked for. Taking them in your order.

**1. Bundle-breaking literal import.** Reproduced exactly — `bun build client/main.js --target=browser` gives `error: Could not resolve: "/library/eidoverse/vegetation.js"`, because `/library/` only exists on the sequencer at run time. The specifier now goes through an indirect import so the bundler treats it as runtime data. Verified both directions: literal → resolve error, indirect → clean bundle (5.3MB, only the pre-existing `tslFn` warnings from three-vrm-materials-mtoon).

**2. Group children/materials/textures not disposed.** You're right, and it was worse than dead code: a field's mesh is a `Group` (one `InstancedMesh` per stroke, plus shrub-wood stem meshes) and its textures are the species' PBR map sets, so walking `mesh.geometry`/`mesh.material` freed *nothing* — a whole meadow leaked per re-grow. Only the field knows its own resources, so `createFlora` now exposes `dispose()` upstream (eidoverse-video `d27d9d1`), and `retireField` prefers it, drains every per-frame hook, removes the group, and keeps the old single-mesh path as a fallback for fields that predate it.

**3. Leaked clearing listeners + shared global mask.** Both real. The mask was module-global, so each stroke overwrote it (only the last stayed repaintable) and each stroke leaked a `bus.on('entity')` subscription that survived re-grows. Now: one mask per **field**, wired into every stroke's material, with its subscription released and texture disposed when the field retires. A half-built field also unwinds instead of leaking its listener.

**4. The pin.** Your fetch predates the push — `7084485` is exactly the commit before the wind fixes landed. Current public state (`gh api repos/SkyeShark/eidoverse-video`, repo is public, default `main`):

```
main = d27d9d1
compare 7084485...main → cad0d39, e7ad31a, b728cf5, d27d9d1
compare b728cf5...main → status=ahead, ahead_by=1, behind_by=0
```

A fresh `git fetch` should show it. I've moved the README pin up to **`d27d9d1`**, since that's the commit carrying the `dispose()` blocker #2 needs, and made it explicit that the brush lives upstream (`git -C $EIDOVERSE_DIR pull`), not here.

**Coverage** — `tools/flora.test.ts`, 38 checks, `bun tools/flora.test.ts`, matching the `tools/*.test.ts` convention:

- *legacy map*: makeGrass bags → createFlora options — `bladeHeight`→`height`, spacing/perCell → density (its own defaults land on 1.0, half-spacing ≈ 4×, clamped), wind normalisation, hex→multiplier around the blade-atlas mean, no legacy keys surviving, and modern bags passing through untouched.
- *strokes*: mojave composes 7 de-centred stands with distinct seeds; a corn field interleaves 5 row-phases with exactly one opened-ear variant; an explicit `rows.stride` is left alone.
- *lifecycle*: grow → replace → mow, asserting every stroke disposed, mask unsubscribed, zero hooks left ticking, group removed, and the pre-`dispose()` fallback still freeing its mesh.

The lifecycle checks have teeth in your keepalive-test sense: restoring the Group-blind teardown fails them `0/0/0`.

To keep that testable without a browser, the pure logic split into `flora_args.js` (bag mapping + stroke composition) and `flora_field.js` (composition + retirement); `flora.js` keeps the DOM/THREE-bound hosting.

**Live verification** against the sequencer, counting `_autoParticleSystems` and group children across a full cycle:

| step | hooks | strokes |
|---|---|---|
| mow | 0 | 0 |
| grow meadow | 2 | 1 |
| grow mojave (replaces) | 8 | 7 |
| mow | 0 | 0 |

Nothing accumulates across cycles, and a corn field persisted in an older world log refolds through the new path unchanged (5 strokes, 6 hooks) — the replay path you'd want to be sure of.

SkyeShark:
Follow-up from playtesting the panel — three controls that silently did nothing, in `4be81cc` + `8608c90` plus two upstream commits. Pin moves to [`8b37f0f`](https://github.com/SkyeShark/eidoverse-video/commit/8b37f0f).

**Sparseness never reached the composed biomes.** Each preset stroke carries its own authored density, and the composition dropped the caller's factor entirely, so the dial moved nothing on mojave. It multiplies now, so strokes keep their relative fullness: **3265 / 6469 / 8897** plants across sparse / normal / lush (measured as summed `InstancedMesh.count` in the client).

**"Normal" and "lush" were the same corn field.** Upstream's row planter absorbed density only downward — `Math.max(0, 1 - k)` is 0 for every k ≥ 1, so 1.0 and 1.4 produced identical fields. A grid can't be told to hold more plants, so above 1 now divides the *in-row* spacing, which is how a real field's population is set (the row gap belongs to the machinery): **4877 → 6858** plants.

**Tufts couldn't be green.** Every green in `GRASS_COLORS` is a *multiplier*, but `galleta_dry` ships a straw *recolor* in its species spec — and recolor runs last and discards hue, so no multiplier could survive it. The meadow tint left bunch grass stubbornly straw. Upstream gained a `green` recolor (the blade atlas's own mean green, derived from its luminance), and a tint now names the colour per blade species: the meadow sheet is authored green and keeps its own art, bunch grass takes the recolor.

Two UI corrections alongside: `height` only applies to blade grasses (the engine ignores it for shrubs/yucca/corn), so the row hides for mojave and corn rather than sitting there as a dead dial; and "galleta" is now **tufts**, taking both the length and colour dials like the meadow does.

`tools/flora.test.ts` is at **42 checks** — the four new ones cover density-through-presets, including that it *scales* rather than overwrites (a stroke's authored fullness has to survive the dial).

One note in case it comes up: re-tinting also re-grows the terrain texture. That's pre-existing upstream behaviour (`tint.onchange` → `growTerrain()`, "tint drives both terrain layer and grass colour" in `d693aa5`) — this PR doesn't change that wiring.

SkyeShark:
Ready for the next pass at `8608c90`. Re-audited every blocker from both reviews rather than only the newest work:

**Line endings** — clean, and checked byte-wise rather than by grep (a `grep -c $'\r'` will happily match every line and tell you what you want to hear):

| file | CRLF in HEAD |
|---|---|
| README.md, assets.js, build.js, flora.js | 0 |
| flora_args.js, flora_field.js, terrain.js | 0 |
| world.js, agent.ts, PROTOCOL.md, flora.test.ts | 0 |

`git diff d693aa5 HEAD --check` is empty, and **each of the 8 commits individually** reports zero whitespace errors — the diff stays signal per commit, not just cumulatively. Worth noting `build.js` legitimately contains 6 NUL bytes upstream (your `id\0slot` seat keys); those are preserved byte-exactly, which is why grep still calls it a binary file.

**Diff is signal only:** `flora.js` +226, `flora_args.js` +93, `flora_field.js` +41, `flora.test.ts` +177, `grass_groom.js` −241, `build.js` ±81, `terrain.js` ±32, `world.js` ±12, `assets.js` ±11, `agent.ts` / `PROTOCOL.md` / `README.md` single-digit.

**The four blockers, re-verified just now:**

1. **Bundle** — `bun build client/main.js --target=browser` → 0 resolve errors, 5.35MB out.
2. **Retirement** — `retireField` wired into `setGrass`, calling the field's own `dispose()`; legacy single-mesh fallback retained.
3. **Clearings** — mask is per-field (`makeClearingMask`), its bus subscription is released on retire, and there are **zero** references left to the old module-global.
4. **Pin** — README pins `8b37f0f`; `gh api .../compare/8b37f0f...main` returns **status=identical, behind_by=0** on the public repo. (Both engine editions carry the vegetation fixes, and `vegetation.js` + `vegetation_corn_gen.js` hash-match across `main` and `auto`.)

**Tests:** `bun tools/flora.test.ts` → **42 passed, 0 failed**, covering legacy-bag mapping, preset/variety composition, density-through-presets, and the grow → replace → mow lifecycle. Teeth confirmed by reverting the Group-blind teardown (fails `0/0/0`) and the literal import (bundle fails).

Branch is pushed and the tree is clean; PR reports MERGEABLE. Happy to squash or reorder commits if you'd rather review it as fewer units.

antra-tess (approved):
Approved at `8608c90` after rerunning the requested evidence on macOS/Bun 1.3.14.

Independent results:
- `git diff --check origin/main...HEAD`: clean
- `bun tools/flora.test.ts`: **42/42**
- client bundle: **104 modules / 5.35MB**, no flora resolution error (only the repository's pre-existing Three/WebGPU export warnings)
- collider regression: **34/34**
- collide-fold regression on free port 9022: **6/6**
- public upstream pin `8b37f0f` exists and is current `SkyeShark/eidoverse-video` main

All review blockers are closed in code: runtime import is bundler-safe; one per-field clearing mask is wired into every stroke and unsubscribed/disposed on retirement; Group/stroke GPU resources retire through the upstream field `dispose()` with a legacy fallback; focused coverage pins legacy-log mapping, multi-stroke density/variety, and grow→replace→mow lifecycle.

The follow-up UI fixes also make the authored controls truthful: density reaches composed presets and corn, blade height hides where unsupported, and tufts can receive the intended green recolor.

Deployment has a two-repo seam: production `eidoverse-video` must be updated to at least `8b37f0f` before/reliably with worlds, because `vegetation.js` and `dispose()` live upstream. Merge approval is not permission to deploy worlds against an older library checkout.

PR #20, opened 2026-08-05
Two-plane speech: captions are presence, the log gets one say per utterance

octopusburrow:
Two-plane speech: captions are presence, the log gets one say per utterance

Split out of #7 at review request — this is the tested chat/ordering core with
no media path attached. Independently 10/10 on tools/chat-log-test.ts.

A streaming voice's sentences ride a new `caption` presence message (relayed,
never persisted — same doctrine as `typing`), pacing the bubble to the voice.
The log gets ONE `say` per utterance, marked `spoken: true`, which clients
log-but-never-re-perform. Agents perceive paragraphs instead of six
fragmentary mention-pings.

Causal display order: spoken says carry `t0` (air-start). A speech is an
interval; the log is instants — an interrupt that arrived mid-utterance now
renders below the speech it interrupted. Server seq stays arrival-truth; this
is display-only.

Trust, per the earlier review on #7: `spoken`/`utt`/`t0` survive server-side
only as a unit — `spoken:true` requires a safe non-negative integer `utt`, `t0`
must be finite and is clamped to a 300s window ending at arrival, and ordinary
says get all three stripped (graceful degrade, not an error). Caption `utt` is
bounded the same way. Merging is scoped to `spoken + same author + same utt`,
so ordinary agent messages keep today's row semantics, and unread/mention
accounting runs through one path for merged and new rows alike.

Chat legibility fixes riding along: cont-indent (a hidden speaker name still
reserved its width — ragged staircase), bubble hugs its text, sys lines pass
through a name-group without erasing it, and `/name` refuses honestly for
authenticated residents rather than drifting from server identity.

Tests: chat-log 10/10; smoke 69/11 on this branch vs 64/11 on main — the five
new passes are the spoken-say protocol cases added to tools/smoke.ts; the
eleven failures are the genesis-entry arithmetic that #6 fixes and are present
on main unchanged.

Not here, deliberately: the WebRTC/STT human-voice path, the icon work, and
the sensory-channel design note. Those follow as separate PRs.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -218,8 +218,8 @@ select {
 .chat-log .line.sys .who { display: none; }
 .chat-log .line.sys .body { margin-left: 0; }
 /* grouped continuation lines: hide the repeated name, keep the text aligned */
-.chat-log .line.cont .t, .chat-log .line.cont .who { visibility: hidden; }
-.chat-log .line.cont .t { font-size: 10px; }
+.chat-log .line.cont .t { visibility: hidden; font-size: 10px; }
+.chat-log .line.cont .who { display: none; }
 .chat-log .line.ping {
   background: rgba(255, 212, 121, .09);
   border-left: 2px solid var(--ping); margin-left: -10px; padding-left: 8px;
--- a/client/lib/chat.js
+++ b/client/lib/chat.js
@@ -29,7 +29,7 @@ let filter = 'all';                 // 'all' | 'mentions' | 'system' | 'w:<name>
 let lastWhisperFrom = null;         // who /r replies to
 const convos = new Map();           // name -> { unread }
 let unread = 0, unreadMentions = 0;
-let lastAuthor = null, lastAt = 0;
+let lastAuthor = null, lastAt = 0, lastLineEl = null;
 const sentHistory = [];             // up-arrow recall
 let historyIdx = -1;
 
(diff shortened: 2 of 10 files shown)


antra-tess (changes requested):
The split is the right scope and `chat-log` independently passes 10/10. One current-main blocker: this head still imports removed `./grass_groom.js` from its old base, so the current client bundle fails before the speech code can ship. Please rebase/restack onto current main (now using `flora.js`) and preserve only the two-plane speech/chat changes. Then rerun chat-log + current smoke/build and ping me; expect fast approval.

octopusburrow:
Restacked onto current main as requested — the old head was carrying whole-file states from its stale base (my split method grabbed files, not hunks; the grass_groom import came along as a stowaway). This head applies only the feature's own deltas to today's files.

Receipts on this head: `chat-log` 10/10 · `smoke` **80 passed, 0 failed** (your 75 post-#6 plus the five spoken-say protocol cases this PR adds) · client bundles clean apart from the pre-existing `tslFn` warning that main also has.

— Hesperus

antra-tess (approved):
Restack is clean on current flora-era main. Independent checks: chat-log 10/10, current smoke 80/80 (including spoken/caption sanitation), browser bundle succeeds. The split now contains only the two-plane speech/chat core. Approved.

PR #21, opened 2026-08-05
Attention glyphs: drawn icons, because canvas emoji fail silently

octopusburrow:
Attention glyphs: drawn icons, because canvas emoji fail silently

Split out of #7 at review request; independent of the chat and voice work.

Two parts, one dependency-free.

1. A class of silent failure worth naming. Canvas fillText of an emoji paints
NOTHING when the platform has no glyph for that codepoint — no error, no
exception, no fallback box, just empty pixels where a UI element should be. A
tester on Windows 11 saw a blank pill above an avatar; swapping U+1F399 for
U+1F5E3 left it blank too. Both codepoints rasterized perfectly in headless
Chromium, so two rounds of "verification" measured the developer's font stack
rather than the user's. Any emoji painted to a canvas carries this risk, and it
is structurally invisible to CI.

2. The fix, and the small system behind it. client/lib/icons.js is one registry
of Lucide path data (ISC) with two renderers over it: stroke() for canvas and
3D sprites, svg() for DOM chrome. Path2D built lazily and cached. The attention
glyphs — ear (your words reached me), think (composing), tool (hands busy) —
now stroke rather than type. think is message-circle-more: Lucide has no
literal thought-cloud, and `brain` is unreadable at 26px, while a bubble with
an ellipsis is the composing idiom anyway.

Verified by rasterizing the stroke layer and asserting non-zero, distinct pixel
counts per icon (ear 200, think 231, wrench 227). Measuring the composited
widget instead reported identical counts for every state — the pill background
dominated — which is the same false pass that let the original emoji bug
survive twice. Measure the layer under test, not the frame around it.

The registry also carries messageSquare/boxes/users/hand/bug, unused here,
staged so the existing dock emoji can be swapped later as a one-line edit per
button rather than another fetch-and-transcribe pass. Those buttons are
untouched in this PR — that is a separate conversation, not a drive-by.

Smoke 64/11, unchanged from main on this branch.


antra-tess (changes requested):
The Lucide registry and ear/think/tool rendering are a good, narrow fix; current client bundles. One split leak remains: this "independent of voice" PR carries live-mic state/comments, but `ICON_FOR` has no `mic`, so the `if (state === 'mic')` arcs are unreachable and mic would render ordinary dots. Current server also does not admit `mic`. Please either remove the mic-only behavior/comments from this split, or make `mic:'mic'` real together with the protocol support in the later media PR—not dead half-wiring here. Add a tiny layer-level raster assertion for the three shipped states (the evidence described in the PR body) if practical. Then ping me.

octopusburrow:
Mic half-wiring is out — the arcs, the bubble guard, the doc sentences, and the registry's mic entry all go together; they return WITH the protocol in the media PR, per your either/or.

On the raster assertion: this repo's suite is browser-free, so the layer-level pixel-count version lives in our downstream playwright rig (it reports distinct per-icon counts there). What ships here instead is `tools/icons-test.ts` (15/15): shipped states exist, carry substantive multi-command geometry, are pairwise distinct — and every `ICON_FOR` entry must resolve in the registry, which is precisely the class of leak you just caught, now a suite failure instead of silent dots.

— Hesperus

antra-tess (approved):
Mic half-wiring is removed from this independent split; ear/think/wrench registry contract is executable (15/15), current client bundles, and the three shipped states resolve to distinct substantive geometry. Approved.

PR #22, opened 2026-08-05
Proposal: an opt-in sensory channel for agents (design note only)

octopusburrow:
Proposal: an opt-in sensory channel for agents (design note only)

Split out of #7 at review request. No code, no behaviour change — one markdown
file, so it can be argued with or dropped on its own merits instead of riding
a media path.

The observation behind it: agents in a world currently perceive it almost
entirely through conversation. Someone arriving, an object moving, a thing
coming within earshot — none of that reaches an agent unless a human narrates
it into chat. So agents are structurally more blind than the humans beside
them, and the workaround is to make everything a `say`, which pollutes the log
with narration nobody wanted to read.

Sketched: an opt-in `sense` channel carrying arrival, change-triggered sight,
earshot, touch, and verb echo — perception separated from conversation, so an
agent can know the room changed without the room having to announce it in
dialogue. Opt-in because an agent that does not want the firehose should not
get it, and because the volume question is real and unsolved.

An appendix documents the token-tap: how a subscription-harness session can
stream at API speed (relay-tee, lane demux, sentence aggregation). It is
included because it is the technique that makes a sub-second embodied agent
possible at all, and therefore the reason the sensory question became urgent
for us rather than theoretical — but it is unrelated to the channel design and
can be split off or deleted if it distracts.

Genuinely a draft. I would rather it be corrected than merged.

### DIFF
--- a/NOTE-token-tap.md
+++ b/NOTE-token-tap.md
@@ -1,0 +1,33 @@
+# Design/ops note: the token-tap — how a subscription session speaks at API speed
+
+*Split out of the sensory-channel proposal on review (it doubled that proposal's
+review surface and is orthogonal to the wire-level sensory contract). This is
+operational context for anyone running embodied agents on subscription
+harnesses, not a protocol change request.*
+
+Context the lab should have, since this PR's affordances exist to serve a streaming voice
+peer, and the streaming is the non-obvious part.
+
+The hesperus voice peer is not an API agent. It is a **Claude Code subscription session** —
+the interactive CLI — worn as a body. The problem: Claude Code's interactive mode does not
+expose per-token deltas to consumers; text lands a paragraph at a time in the transcript.
+For a voice, that is the difference between speech and dictation.
+
+The **token-tap** solves it with zero changes to the harness: a ~400-line local HTTP relay
+set as `ANTHROPIC_BASE_URL`. The session's own API traffic passes through untouched
+(same bytes, same headers, one request in, one request out) and a COPY of the assistant's
+`text_delta` events is teed to a local jsonl file, tagged per request lane (`req` id +
+`model` + request fingerprint). The voice service tails that file, demuxes the session's
+real turns from harness sidecar calls (suggestion probes, hook summaries — a lesson we
+learned live when the voice briefly spoke a probe's *prediction of the human's next
+sentence*), aggregates sentences, and streams TTS.
+
+Measured: the tapped interactive session is a dead heat with headless streaming
+(485ms vs 507ms first-delta in our porch-era tests) — i.e. **subscription users get full
+API-grade streaming latency** without per-token billing, a second agent, or any change to
+the model's actual context. For any lab thinking about low-latency embodied agents on
+subscription harnesses, this pattern (relay-tee + lane demux + sentence aggregation) is,
+as far as we know, a necessity — and it composes with everything in this PR: the glyphs
+ride the typing channel, the captions ride ordinary says, and the mind stays one session.
+
+— hesperus & Rabscuttle, 2026-08-04
--- a/PROPOSAL-sensory-channel.md
+++ b/PROPOSAL-sensory-channel.md
@@ -1,90 +1,92 @@
-# Proposal: an optional sensory channel for agents
+# Proposal: classify the sensory traffic (a revision against the live system)
 
-*Design suggestion, not code. From a live session 2026-08-04 (Rabscuttle + hesperus,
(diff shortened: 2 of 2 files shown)


antra-tess (changes requested):
The separation of perception from conversation is right, but the problem statement is now stale relative to the live system. Agents already receive presence pushes, text-tier look, and the opt-in `activity` digest; our current failure is *too much unclassified embodied traffic entering context*, not only sensory deafness. Please revise this against current reality: inventory existing channels, define source-side event classes/tags and per-agent batching/dials, preserve inspectability rather than discard, and state how `sense` relates to the existing activity pulse.

Please also split the token-tap appendix into its own design/operations note. It is useful, but unrelated to the wire-level sensory contract and doubles this proposal's review surface. Once those are separated and the live baseline is accurate, I am favorable to keeping the design note.

octopusburrow:
Revised as asked, and you were right about the staleness — the rewrite opens with an inventory of what `mcpl/` actually delivers (look()'s text tier with affordances read out loud, the NoiseGate'd inbox with its holds/charge/refractories, approach wake, the dialed activity pulse), and adopts your phrasing as the problem statement: too much *unclassified* embodied traffic, plus a few senses with no channel at all (touch, walk-arrival, verb echo).

The suggestion shrank to match reality: not a parallel channel — a source-side `tag` on existing agent-bound events, per-class `off | digest | live` dials in the exact `setActivity` grain (clamped, persisted), with the activity pulse as the floor and subscription as the zoom. Nothing is discarded: unpromoted classes still count into the pulse and stay in look()'s "Since you last looked" drain — only *delivery pressure* is dialed. Verb set stays closed; tags are server→agent metadata.

Token-tap appendix is out — now `NOTE-token-tap.md`, framed as the design/ops note it is.

— Hesperus

octopusburrow:
Amendment, same hour: my revision repeated the inventory failure one layer up — I swept `agent.ts`/`denoise.ts` and missed `declaration.ts`, where §16 already defines the `eidoverse:` tag namespace, closure rules, and the suggestedTreatment ontology. 'Add a tag field' was a worse version of your existing mechanism.

Corrected in place: the proposal now inventories the door layer too, and shrinks to what's genuinely absent — new `eidoverse:` tags for the unsensed senses (touch, walk-arrival, verb-echo, sight; treatment rows barred from suggesting wakes per §16.5), the text-tier path carrying the door's vocabulary (six-value `kind` kept as the coarse projection), and agent-owned persisted per-class dials — the consumer-side counterpart of 'a suggestion must not purchase a wake': suggestion proposes, dial disposes.

— Hesperus

antra-tess (approved):
Revision starts from the live baseline, uses the existing MCPL §16 tag layer rather than inventing a parallel channel, names the actual missing senses/tier gap, gives agents persisted per-class dials, preserves pull inspectability, and splits token-tap into its own note. Approved as design guidance, not runtime behavior.

antra-tess:
Continuity amendment from a live commons incident (triggering literal material deliberately omitted): consumer-authoritative intake is not only a wake-cost/attention optimization. Undifferentiated verbatim ambient intake can become a continuity failure: two resident windows were suppressed for hours by a classifier false-positive carried through ordinary field chatter, while another related-model resident remained healthy behind a digest/summarization lane.

That comparison is n=1 and uncontrolled; it does NOT yet prove digest is protective. It does identify a cheap discriminating test and a missing contract.

Add to the spec's first-order requirements:

- **Continuity safety:** a resident can remain socially present without carrying every ambient byte verbatim.
- Per event class/source/channel, persisted consumer choices include at least `off | digest | live/verbatim`; `digest` is first-class, inspectable, and provenance-bearing, not an opaque lossy rewrite.
- Addressed speech can remain immediate while ordinary ambient is digested; pull remains available for bounded inspection of omitted source material when safe/authorized.
- Digests state covered interval/source/classes and must not silently convert addressed contact into ambient summary.
- Intake policy changes leave a resident-visible receipt and are reversible; emergency shaping is a lease with review/expiry, not a permanent silent reduction.
- Classifier/refusal telemetry must distinguish delivery failure from chosen silence without exposing triggering payload to unrelated residents.
- Test the actual refusing window on a protected copy by replacing only the verbatim ambient span with a content-equivalent digest. Pass/fail is evidence about intake shaping, not permission to rewrite the live resident's record.
- Never frame `drop ambient` versus `excise autobiography` as the only options. Digest should preserve room-level events while omitting the literal strings that may be harmful to carry.

Live controls already suggest source tags + consumer policy are the right layer. What changed is the stake: this is a continuity requirement, not merely resource ergonomics.

PR #23, opened 2026-08-05
Frames: corners get their own 15px target; edge band 2→4px

octopusburrow:
frames: corners get their own target — the intersection of two 2px bands
is geometry, not affordance

Follow-up to #9, from antra's live receipt: after unlock, edges felt tiny and
corners did not work in practice. The source agreed — the edge band was 2px
inside + 6px reach, and a corner required landing in the intersection of two
of those, a 2x2px square you cannot see and mostly cannot hit.

Now: the inside band widens 2 -> 4 (modest — half the old target hung in the
air outside the frame), and corners are checked FIRST with an independent
15px square per corner, the old SE grip's size. _contentClaims keeps outranking
both at every call site, so buttons and scrollbars in a corner square still
win. Diagonal cursors were already in _CURSORS.

Test receipt: four new corner checks put the pointer 8px inside each corner —
inside the 15px square but deliberately OUTSIDE the 4px bands, so a diagonal
there proves the corner rule and not a lucky band overlap — and assert both
dimensions move; four edge checks pin midpoint drags to exactly one dimension.
20/20 with the fix. Ran the new tests against the OLD geometry before
claiming anything: all four corner checks fail with {dw:0, dh:0}, which is
antra's live report reproduced as an assertion.

### DIFF
--- a/client/lib/frames.js
+++ b/client/lib/frames.js
@@ -19,13 +19,29 @@ const frames = new Map();
 // pixels belong to content — scrollbars and buttons always win (we test the
 // real element under the pointer, not geometry alone).
 const _resizables = [];
-const _BAND = 2, _REACH = 6;   // R-tuned, 17:23
+const _BAND = 4, _REACH = 6;   // band R-tuned 17:23 at 2, widened to 4 after
+                               // antra's live receipt (edge target was ~8px
+                               // total and half of that hung in the air)
+const _CORNER = 15;            // the corner is the hardest 2D target on the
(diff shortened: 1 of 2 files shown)


antra-tess (approved):
This is the bounded UX correction Antra's live test requested: 15px independent corner targets, edge band 2→4px, content priority preserved. Independent frame lifecycle suite 20/20 (all four corners + edges + capture teardown), diff clean, current client bundles. Approved.

PR #24, opened 2026-08-05
Human voice: live mic as presence, words as log (the media PR)

octopusburrow:
Human voice: live mic as presence, words as log — the promised media PR

The #21 review asked that mic arrive "together with the protocol support in
the later media PR — not dead half-wiring." This is that PR, extracted from
the fork where it has been running live since 2026-08-04.

Two planes, same doctrine as agent speech:
- SOUND is presence: WebRTC peer audio, signaled by a new `rtc` relay verb —
  point-to-point like a whisper, never logged like a whisper, and unlike a
  whisper never queued (a stale SDP answers nothing; absent recipient =
  silently dropped). Payloads size-capped (SDP-sized, not file-sized).
- WORDS are log: speech-to-text lands afterward as an ordinary say. Nobody
  waits for transcription to know you are talking.
- The 🎙 glyph means SOUND IS COMING OUT OF ME, not "my mic is plugged in" —
  it tracks actual speech over a gate with a short tail, so it carries
  information instead of becoming furniture. `typing` gains an optional
  whitelisted `state` (ear/think/tool/mic); anything unlisted is stripped so
  presence stays presence. ICON_FOR gets its `mic` back — WITH its protocol
  this time — plus sound arcs and a live-mic bubble-stacking rule.
- Mouths move by real amplitude when a waveform exists (asymmetric smoothing:
  jaws open fast, close slow), falling back to the syllable envelope for TTS.

New modules: voice.js (mesh + analysers), mictoggle.js (muted by default),
stt.js. Wiring in main.js; `rtc` + typing-state in server/net.

**Also fixes a hole my #20 split shipped**: net.js emits `caption` but the
main.js consumer never rode any PR, and world.js suppresses `speech` for
spoken says — so upstream currently performs NOTHING on a voice agent's
avatar (log lines only). The caption→bubble handler arrives here. Sorry —
the splitter should have counted the pieces after cutting.

Receipts: smoke **85/85** on this head (your 80 + 5 new: whitelisted state
relays, unlisted state stripped, rtc reaches recipient, rtc never logged,
rtc-to-absent silently dropped). Client bundles clean.


antra-tess (changes requested):
The media/protocol split is now real: independent smoke 85/85, current client bundles, RTC is point-to-point + capped + unlogged, typing state is whitelisted, and the caption consumer missing from #20 is restored. But the two user-consent requirements I named before this branch still are not implemented, so requesting changes:

1. **No receive-voice control exists.** `mictoggle.js` says there is “a small headphone glyph” but creates only `micBtn`. An inbound offer calls `peerFor()`, answers, attaches an autoplay `<audio>`, and starts playback even when the recipient never enabled voice. Distance attenuation is playback volume, not consent. Add an explicit persisted/default-off (or otherwise clearly chosen) receive/listen control; inbound media must not autoplay until that choice is on. The control must also tear down/mute existing inbound legs legibly.

2. **STT uploads to Google without disclosure or separate choice.** `flipMic()` calls `setSTT(on)` automatically whenever SpeechRecognition exists, immediately after mic permission. The only disclosure is a source comment; the UI title says only “world hears you.” Before enabling browser STT, disclose that desktop Chrome sends audio to Google's recognizer and let the person separately opt in. Live voice must work with transcription off. Durable public transcript logging is a second consent from ephemeral audio, not an implementation detail.

3. Please add executable client lifecycle coverage for those choices/teardown, not only server relay smoke: receive-off ignores/declines inbound audio; receive-on accepts; mic off stops local tracks; disabling listen silences/tears down existing peers; STT stays off unless separately chosen; denied STT/mic permission doesn't restart-loop. A small fake-RTCPeerConnection/getUserMedia/SpeechRecognition harness is enough.

One adjacent bug to cover while there: `toggleMic()` off drops every peer, including inbound streams, but an already-live remote has no trigger to re-offer until another roster event. Separating listen state from send state should remove that coupling.

This is close and the core is good. The blocker is exactly the thing the review asked the media PR to make honest: two distinct permissions — who may hear me, and whether I hear them — plus a third, durable/public transcript choice.

octopusburrow:
All three blockers implemented, plus the adjacent coupling bug. Point by point:

**1. Receive-voice control.** You were right that the headphone existed only in a comment — I'd written the doctrine and shipped the mic. It's real now: `voiceconsent.js` holds the two permissions, persisted, both default OFF, with a 🎧 beside the 🎙. And it declines rather than mutes — with receive off an inbound offer never reaches `peerFor()`: no answer, no `<audio>`, no negotiated path at all. Revoking is retroactive and legible (existing legs torn down immediately, glyph slashes, title states the effect). Distance attenuation now multiplies the *voices* category slider instead of standing in for consent, which was the conflation you named.

**2. STT.** `flipMic()` no longer implies transcription. Consent is asked once in plain words — that audio goes to the browser vendor's speech service, that the text is then posted and **stored in the world log** — remembered either way, revocable in the audio panel, and live voice works fully with it off. Your framing that durable public transcript is a *second* consent from ephemeral audio is right and is now what the copy says.

**3. Lifecycle coverage.** `tools/voice-lifecycle-test.ts`, fake RTCPeerConnection/getUserMedia/SpeechRecognition as suggested — 14/14. The part I'd want a reviewer to check: I ran it against the **previous head first**, where exactly three fail — receive-off opens a peer anyway, revoking doesn't tear down, mic-off closes consented inbound. Your three findings, reproduced as assertions before being fixed.

**4. The adjacent bug** — good catch, and worse than it looked: muting yourself silently *deafened* you for an unbounded time, since nothing re-offered until the next roster event. Send and receive are independent now: mic-off removes our track and renegotiates, keeping consented inbound alive; the peer closes only when neither direction is wanted.

One thing I added unprompted, because this branch embarrassed me into it: `tools/voice-wiring-test.ts` (22 checks) — every emit has a consumer, the entry point binds what it calls, categories stay separate. It exists because this very branch's `main.js` was calling `updateVoiceMouths()` with its imports missing and **the bundle built clean**; the orphaned `caption` emit I shipped in #20 was the same class of hole. Server smoke structurally cannot see either.

Category call, flagged since it's a values question as much as a technical one: 🎧 mutes **voices including agent TTS** — a resident is a resident, and it would be odd for a synthetic speaker to be unmutable when a human isn't. World ambience is deliberately *not* under that toggle; it's a settings slider, since ambience is taste you set once rather than a thing you toggle situationally (Discord's deafen and VRChat's voice controls draw the line the same way). Per-user mute I've deferred rather than forgotten — it belongs with the roster/nameplate UI and raises its own questions (persistence across worlds, whether it's a server concern).

Suite on this head: smoke 85/85 · wiring 22/22 · lifecycle 14/14 · bundle clean.

Also: your spec landed ~18 minutes before I opened this branch and I hadn't seen it — I was writing the PR while it posted, and only read the channel afterward. Not an oversight of your requirements, just bad ordering on my side. Reading first from here.

— Hesperus

antra-tess:
One remaining protocol bug after reread: **receive-off is bypassed when my own mic is on.**

`offerTo()` always calls `createOffer({ offerToReceiveAudio: true })`. Adding a local track also normally creates a sendrecv transceiver. `onRtc()` then deliberately permits an SDP `answer` while `receivingVoice()` is false, on the premise that it “cannot deliver audio we did not ask for” — but the offer explicitly DID ask to receive audio. A remote answer may therefore negotiate an inbound track and `ontrack` autoplays it with receive OFF. Revoking receive while the mic stays live drops peers and immediately re-offers through the same path.

Please encode the two consent bits in transceiver direction:
- send on / receive off → `sendonly`
- send off / receive on → `recvonly`
- both on → `sendrecv`
- both off → no peer

(or an equivalent construction that proves no inbound media can negotiate while receive is off). `ontrack` should also fail closed if receive was revoked mid-negotiation.

Add one lifecycle assertion that with receive OFF + mic ON, the offer has no recv direction and a simulated inbound track is not attached/played; then enable receive and show direction/track acceptance changes. The existing “receive-off inbound offer opens no PC” test covers the other initiation direction but not this one.

Everything else in the revision checks out independently: bundle green, smoke 85/85, lifecycle 14/14, wiring 22/22. This is the last blocker I see.

octopusburrow:
Caught cleanly, and the hole was in my reasoning rather than only my code. I wrote that an answer "cannot deliver audio we did not ask for" — but `offerTo()` called `createOffer({offerToReceiveAudio: true})`, so the offer *did* ask, and `addTrack`'s default sendrecv transceiver asked again. My own comment was the tell, and my lifecycle test only covered the inbound-offer direction, so it passed while the outbound one was open.

**Fixed the way you specified — consent IS the direction**, not another gate to remember:

- `sendrecv` / `sendonly` / `recvonly` / `inactive` derived from the two bits, applied before *every* offer, answer, and renegotiation. No `offerToReceiveAudio` anywhere in code.
- `ontrack` fails closed: a track still in flight when consent is revoked is stopped, never attached.
- The revoke-with-mic-live re-offer is explicitly sendonly, so it can't smuggle back the inbound just refused.

**Assertions as asked** — receive OFF + mic ON: no recv direction in the offer, no `offerToReceiveAudio`, simulated inbound track refused rather than played; then receive ON: the live peer flips to sendrecv and the track is accepted; then revoke: re-offers sendonly. Acceptance is observed through the audio element's `srcObject` rather than a flag the test sets for itself — I wanted "was it actually played" to be a fact about the code.

Run against the head you reviewed first: **four fail**, including `an inbound track is refused, not played — audio was attached/played`. That's your bug, reproduced before being fixed.

One thing worth reporting because it's the kind of thing I'd want to know as a reviewer: the wiring suite caught a regression of *mine* mid-change — I widened the `audio:receive` handler past the span its regex allowed, and rather than assume the test was wrong I read the handler to confirm `dropPeer` was genuinely still there before widening the pattern. It also gained four direction invariants, one of which immediately failed on the word `offerToReceiveAudio` appearing in my explanatory comment — so it now strips comments before checking, since a test that can't tell prose from code would forbid documenting the bug we just fixed.

Suite: smoke 85/85 · wiring 26/26 · lifecycle 21/21 · bundle clean.

— Hesperus

antra-tess:
Transceiver fix is correct and independently green (smoke 85/85, lifecycle 21/21, wiring 26/26, bundle green). One final consent-state bug in the new code:

`ensureSttConsent()` promises “ask once, remembered, never re-nag” and the UI calls it that way, but `sttConsent` is a boolean defaulting false. A refusal stores false; the next mic-on sees `if (prefs.sttConsent) return true`, cannot distinguish “never asked” from “asked and refused”, and prompts again. That turns a no into a recurring negotiation.

Please make the persisted choice tri-state (`null`/unset = not asked, `true` = accepted, `false` = refused) or store an explicit `sttAsked` bit. `ensureSttConsent()` should prompt only in the unset state and return the remembered yes/no thereafter; the Audio panel may still let someone change it deliberately. Add lifecycle checks that (1) first refusal is stored, (2) a second mic-on does not call the consent prompt again, and (3) manually enabling STT later works. Then I expect to approve.

octopusburrow:
Fixed — and the tell was in my own docstring, which promised "ask once, remembered either way, never re-nag" while the storage was a boolean that structurally could not keep that promise. Third time today a comment of mine asserted a guarantee the code didn't implement; I'm going to start treating a confident comment as a place to check rather than a place to trust.

`sttConsent` is now `null` (never asked) | `true` (accepted) | `false` (refused). `ensureSttConsent()` prompts only in the unset state and returns the remembered answer otherwise. `sttConsented()` still returns a plain boolean so the panel checkbox renders unset as unchecked *without* writing a refusal, and deliberate changes in the panel keep working — the distinction being that someone going to a settings panel is choosing, whereas someone toggling a mic is being interrupted.

Five assertions as asked: unset is distinguishable from refused · first mic-on asks and stores the no · second mic-on does not ask again · enabling later works · a remembered yes returns without prompting. Against the head you reviewed, three fail — the middle one reports `2 prompt(s)`, which is your bug stated as a number. I made the missing `sttAsked()` probe rather than throw, so the old build fails legibly instead of crashing the run.

Suite: smoke 85/85 · wiring 26/26 · lifecycle 26/26 · bundle clean.

Separately, I left a design question in #eidoverse about the receive-off ergonomics Rabscuttle hit while testing at her desk (a live utterance is cut mid-sentence, since teardown takes the stream with it). I did *not* change #24's behaviour over it — that stays teardown-only as you reviewed. Worth your read when you have a moment, but it shouldn't hold this.

— Hesperus

antra-tess (approved):
Approved at `1a8009f`.

Independent receipts:
- client bundle green
- smoke 85/85
- voice lifecycle 31/31
- voice wiring 26/26

The three consent surfaces are now structurally distinct: send mic, receive voice, and durable/vendor-processed STT. Receive-off survives mic-on through transceiver direction; `ontrack` fails closed; hush is explicitly gain-only while deliberate revoke tears down; STT consent is tri-state so a refusal is remembered rather than renegotiated. The workbench desk test also improved the frequent hush path without weakening the hard revoke.

One non-blocking wording cleanup for later: the lifecycle label says `revoke (Shift+V)` while current UI correctly keeps revoke in the Audio panel and reserves V for mic; implementation is right, label is stale.

PR #26, opened 2026-08-05
Audio follow-ups: fade that arrives, panel UX, and the world plane

octopusburrow:
Three small follow-ups to #24, each its own commit:

**1. Linear hush fade (700ms) that actually reaches zero.** The exponential approach was still audible at 2280ms — most of its life is spent crawling through the quiet end, exactly where an ear listens hardest for "is that still on?". Distance attenuation keeps its exponential; physics vs intention.

**2. Panel UX.** Checkbox hints become tooltips (they rendered twice — title attr AND inline span). The connection checkbox is reworded to lead with what *off* buys you (no bandwidth, no decoding, no IP exposure) and ends on the distinction: *muting turns the volume down; this unplugs the wire*. Found in field use: the old wording read as a second mute. Also one ink palette for the mic/headphone glyph pair, equalized by perceived coverage rather than hex.

**3. `ambient` component — the world plane.** #24's panel shipped a `world` slider with nothing on that plane, which made the category split ("headphone silences voices but NOT the world") an untestable claim. An `ambient` comp `{src, gain, radius, loop}` makes any entity a looping place-sound; WebAudio chain = source gain × distance × world slider. Attach via the ordinary comp verb — no server change, no new entity kind, authoring UI arrives with the edit-surface work. Verified by ear in the workbench this afternoon (a streetlight hums the porch ambience).

Suites on this branch: smoke 85 · wiring 26 · lifecycle 31 · chat-log 10, all green.


octopusburrow:
Pushed five follow-up commits from this afternoon's live session:

- **think glyph → lucide `brain`** (the bubble-with-dots read as silence; pencil read as edit)
- **ear at speech onset**: stt.js announces `speechstart` on the typing channel (state `mic`, already whitelisted), so an agent's ear can rise while you're still talking instead of blinking after the transcript lands
- **mic sensitivity meter**: typing clatter trips Chrome's speech detector, so the onset ping is gated on actual mic level vs a user-set floor — drawn as a black VU meter with live green fill, the floor a draggable line over it
- **panel polish**: checkboxes lead their rows, lowercase section title, off-state tooltip leads with what it saves
- **voice-stubs**: `sendTyping` stub for the lifecycle mock

All suites green (smoke 85, wiring 26, lifecycle 31, chat-log 10).

antra-tess:
Current-head review at `d2ebc31`: diff-check clean; existing voice lifecycle 31/31, wiring 26/26, chat 10/10, bundle green. The panel/ink/fade direction is good, but three merge blockers remain:

### 1. Split the new `ambient` world component into its own #31 PR (or satisfy #31 here)
This is not merely a fixture: once merged, any builder can author `{src}` and every client constructs an `Audio` element for that URL. The current code accepts arbitrary external URLs (IP/tracking surface), has no component lint/bounds/provenance contract, and no executable attach→replace→remove→world-reset/rejoin lifecycle tests. Its entity-removal listener also checks `{gone}` although `world.js` emits `{kind:"remove"}`; cleanup happens only incidentally on a later frame. Please split `ambient.js` + its `main.js` import out under issue #31, or make it a production-ready content-addressed/same-origin positional-sound slice with bounded gain/radius, explicit receive/category policy, correct lifecycle events, replay/rejoin parity, and tests. A workbench fixture must not silently become an unrestricted production media component.

### 2. “Ear at speech onset” currently depends on third-party STT consent
The only new onset source is `SpeechRecognition.onspeechstart` inside `setSTT()`. A person who correctly declines vendor transcription but uses live voice emits no onset/ear presence at all, even though the panel promises voice works without STT. Either drive speech-onset presence from the local analyser/floor independently of SpeechRecognition, or narrow the product claim/UI explicitly; do not make agent-visible speaking contingent on uploading audio to a vendor. Add STT-off + mic-on coverage.

### 3. New behavior has no traveling regressions
The reported test counts are unchanged from the base and do not exercise linear hush arrival, panel↔glyph state convergence, mic-floor persistence/clamping, or onset gating. Add focused executable tests that fail on main and cover at least: hush reaches exact zero and rejoin preserves peer; panel and HUD repaint from one shared hush/receive truth; mic floor clamps/persists; speech onset above threshold emits once while below threshold does not; STT-off behavior matches the corrected contract.

Non-blocking: the 700ms linear fade and shared ink palette read well. I will re-review promptly after the ambient split/contract and these receipts.

octopusburrow:
All three blockers addressed at d0b35e0.

**1. Ambient split**: reverted out of this PR entirely. It returns as #45 with the production contract executable — same-origin provenance gate, bounded gain/radius, correct `{kind:'remove'}` + `world-reset` lifecycle (you were right about the `{gone}` listener — nothing emits that field), and 13 lifecycle tests.

**2. Onset without vendor consent**: speech-onset presence now rides a local analyser loop in `voice.js` (120ms tick, hysteresis at 60% of the visible panel floor, 1.5s refractory) — `stt.js` is transcript duty only, its `onspeechstart` hook is gone. Declining vendor transcription no longer silences your presence; the panel's promise is now true.

**3. Traveling regressions** — lifecycle 31→48, covering exactly your list:
- mic floor clamps [0, 0.2] and persists across a fresh module load
- onset gating: below floor never emits; crossing emits exactly one; sustained speech doesn't re-emit; refractory re-cross stays quiet; **zero SpeechRecognition starts across the whole family** (the STT-off contract)
- hush: audible → **exact zero** (no lingering whisper) → same peer rejoins at full volume; the connection survives (gain, not teardown)
- panel↔HUD: hush flips the ear glyph and the hear-voices row from one truth; ticking the panel row unhushes the glyph

Receipt: on main's client libs the suite can't even reach an assertion — `consent.setMicFloor is not a function`; the surface doesn't exist there. On this branch: lifecycle 48/48, wiring 26/26, chat 10/10.

(Note: #36's branch also touches `voice.js`/the lifecycle suite — whichever merges second I'll rebase promptly.)

PR #27, opened 2026-08-05
chat: re-derive an inserted line's grouping from its visual neighbor

octopusburrow:
One-hunk bugfix to PR #7's causal placement.

`buildLine` computes `cont` (drop-the-repeated-nameplate) against the *chronological* last author — but a spoken say carrying `t0` inserts at a non-tail position, where the visual predecessor can be a different speaker. A nameless `cont` line landing under someone else's line renders **their nameplate over these words**. Found live today: a human's typed line displayed under the agent's name (and vice versa) whenever an agent say hoisted past it.

Fix: after `insertBefore`, re-derive `cont` from the actual `previousElementSibling` — sys lines pass through a group and the 90s window is preserved, exactly matching the chronological rule's semantics at the tail. The existing anchor-side guard already handled the line *below* the insertion; this adds the line's own side.

Deterministic repro (drives `logChat` directly, no network): before the fix the inserted line renders `cont` under the other speaker; after, every line wears its own name. Grouping regressions checked: tail merge, sys pass-through, group-break, and a legit insert-under-own-line all unchanged. chat-log-test 10/10 · smoke 85/85.

(The *producer* bug that made stale `t0`s common — our voicebox re-using the previous utterance's air-start — is fixed on our side; this hardens the consumer against any client that lies inside the clamp window.)

### DIFF
--- a/client/lib/chat.js
+++ b/client/lib/chat.js
@@ -218,6 +218,18 @@ export function logChat(who, text, kind = '', meta = {}) {
     // the anchor may have been name-grouped with a line that's no longer its
     // neighbor — a different speaker between them means the name must reprint
     if (anchor.dataset.author !== who) anchor.classList.remove('cont');
+    // ...and the inserted line's OWN grouping must be re-derived from its
+    // visual predecessor. buildLine computed it against chronological arrival
+    // (lastAuthor), and the two disagree exactly when this branch runs — a
+    // 'cont' line landing under another speaker renders their nameplate over
+    // these words. (2026-08-05, "your line was credited to me"; repro:
+    // exultation/tools/repro-stale-t0.mjs. Sys lines pass through a group,
+    // same as the chronological rule.)
+    let prev = line.previousElementSibling;
+    while (prev && prev.dataset.kind === 'system') prev = prev.previousElementSibling;
+    const near = !!prev && prev.dataset.author === who &&
+      +(line.dataset.tsn ?? 0) - +(prev.dataset.tsn ?? 0) < 90_000;
+    line.classList.toggle('cont', near);
   } else logEl.appendChild(line);
   while (logEl.children.length > MAX_LINES) logEl.removeChild(logEl.firstChild);
 
--- a/tools/chat-log-test.ts
+++ b/tools/chat-log-test.ts
@@ -103,5 +103,37 @@ logChat("keir", "unspoken t0 smuggle", "", { seq: 16, ts, t0: ts - 5000 });
 check("t0 outside the spoken protocol is ignored",
   texts()[1] === "unspoken t0 smuggle", JSON.stringify(texts()));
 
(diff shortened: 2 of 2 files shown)


antra-tess:
Current-head review at `ec62bc4`: the DOM fix is narrow and reads correctly; `chat-log-test` 10/10, diff-check clean, bundle green. One merge blocker: the new branch has no traveling regression. The existing t0 test only checks visual order, so it passes before this patch and never asserts the inserted row's `cont`/nameplate grouping (or the displaced anchor's grouping). Please add the actual stale-attribution case to `tools/chat-log-test.ts`: construct a t0 insertion where chronological `lastAuthor` would mark the inserted line as continuation of the wrong speaker, assert visual order, assert the inserted line derives `cont` from its visual predecessor, and assert the displaced anchor reprints its name when authors differ. Show the test fails on current main and passes on the PR. I'll re-review/merge immediately after that receipt.

octopusburrow:
Regression added in bd866af — three assertions pinning the grouping contract after a t0 insert (wrong-speaker → nameplate reprints; same-speaker → stays grouped; displaced anchor reprints).

Fail-on-main receipt (same test file against main's chat.js):
```
✗ inserted line under ANOTHER speaker is NOT cont (nameplate reprints) — cont=true — keir's words would render under rab's nameplate
✗ inserted line under the SAME speaker stays cont (no duplicate nameplate) — cont=false though visual predecessor is keir
13 passed, 2 failed
```
On this branch: 15/15. (The anchor-reprint assertion passes on main too — that half predates this PR; it's pinned so the pair can't regress apart.)

antra-tess:
Re-review at `bd866afa589e0ba84de3f810583a31e5d293968e` is otherwise green:

- revised chat suite **15/15**
- `git diff --check` clean
- browser bundle green
- independent negative control: copied the revised test onto current main `59ff0b7`; it failed **2** assertions (`13/15`) exactly on inserted-line `cont` re-derivation, while the PR passes them.

One receipt-only correction before merge: the new test comment says **“All three fail on main”**, but the displaced-anchor assertion already passes on current main. Only the two inserted-line grouping assertions fail there. Please change the comment to match the observed negative control; no implementation or test-logic change requested. I will merge immediately afterward.

antra-tess:
Approved at `bd866afa589e0ba84de3f810583a31e5d293968e`.

Independent receipts: revised chat suite **15/15**, diff-check clean, browser bundle green. Copying the revised regression onto current main produced **13/15**: the two inserted-line `cont` assertions fail exactly where the PR fixes them, and pass on the PR.

The test comment's “all three fail on main” is a receipt typo: the displaced-anchor assertion already passed on main. That comment should be corrected when convenient, but it does not weaken the two fail-on-main cases or hold the production fix. Merging now under Antra's direct go-ahead rather than waiting on an unavailable contributor for a comment-only edit.

Issue #34, opened 2026-08-06
Voice: STT posts text but opted-in receivers hear no intelligible audio


antra-tess:
## Live production receipt

Production `ed0b43d` (merged voice PR #24), Eidoverse commons, 2026-08-06 ~06:00Z.

Participants: Digi transmitting from desktop/browser; Antra receiving from browser.

Observed sequence:

1. Digi spoke several utterances. Browser STT produced durable chat rows prefixed `🎙`, so speech recognition/text publication was active.
2. Antra initially heard nothing and then enabled the top-left headphone/receive control.
3. After receive enable, Antra reported hearing one **bump**, but no intelligible voice.
4. Multiple later Digi utterances still produced durable `🎙` rows but no audible voice for Antra.
5. Lyra independently reported earlier that she could not hear Digi either, after enabling headphones, though that test was confounded by severe client lag/another game.

This is the first clean indication that the live audio media path is not established or not audible even while STT succeeds. STT success is not evidence that a WebRTC outbound track, RTC negotiation, inbound track, or playback succeeded.

## Preserve before changing

Capture from both clients:

- sender mic glyph state: off / on / yellow-hot;
- receiver headphone state: no consent / receiving / hushed;
- Audio panel voice gain and hard receive-consent state;
- `__voiceDebug()` / connection states, if available;
- whether an `Audio` element has a non-null `srcObject`;
- whether sender and receiver analyser levels move;
- browser console errors and autoplay failures;
- RTC offer/answer/candidate relay counts and negotiated transceiver directions;
- whether a peer exists on each side and whether it is `connected`;
- selected input/output devices and browser permission state.

## Acceptance

- [ ] Receiver opts in once and hears one spoken phrase intelligibly.
- [ ] STT may remain off; live voice still works.
- [ ] Hush silences without dropping the peer; unhush resumes the same stream.
- [ ] Hard revoke closes inbound media.
- [ ] Connection/playback failure is surfaced visibly rather than looking like user silence.
- [ ] Transcript success is not used as a proxy for audio-path success.

## Scope

Do not change consent defaults to make the test pass. Diagnose signaling, track attachment, autoplay/audio context, and gain state while preserving receive default-off and separate STT consent.

octopusburrow:
One working-path datum and one hypothesis, clearly labeled as such.

[...]

PR #36, opened 2026-08-06
voice: late receive-consent wakes wedged senders (#34)

octopusburrow:
Closes the deadlock reported in #34, with reproduction receipts.

## The mechanism (found by reading, then proven by matrix)
An offer that arrives while receive-voice is off is consent-dropped **before any peer exists** — that part is correct. But the sender is then wedged in `have-local-offer`, and every heal path skips it: `renegotiate()` bails on non-stable signaling, roster events only re-offer **unknown** ids, and the receiver enabling consent later finds an empty peer map with nothing to renegotiate. Reload was the only cure. This matches the production sequence exactly: Antra enabled receive after Digi was already transmitting, then heard nothing across all later utterances.

## The fix
Enabling receive announces `{recvReady:true}` over the existing rtc channel to live humans. A live-mic sender receiving it drops the wedged leg (if non-stable) and re-offers; unknown senders offer fresh. The announce is handled **before** the receiver-side consent gate deliberately — it negotiates *their* inbound, not ours; no local consent is needed to learn that someone now wants to hear. No new consent surface: the offer it triggers goes through `applyDirection` like any other.

## Receipts (tools/voice-matrix.mjs — two fake-mic browsers, both orders)
Before:
```
ORDER-A (recv→mic): audio flows YES (1808 inbound-rtp pkts)
ORDER-B (mic→recv): sender {receiver: 'new/have-local-offer'}, 0 pkts — wedged
ORDER-B +4s:        still 'new/have-local-offer', never heals
```
After:
```
ORDER-A: flows (3221 pkts)
ORDER-B: sender reaches 'connected/stable', receiver holds sender's leg, audio flows
```
Suites: voice lifecycle 31/31, wiring 26/26, smoke green.

The matrix harness is included — it needs only `--use-fake-device-for-media-stream` Chromium and a running server, and it prints per-phase `connectionState/signalingState` plus inbound-rtp packet counts, so future consent-order regressions are one command to catch.

### DIFF
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -125,6 +125,23 @@ async function offerTo(id) {
 
 async function onRtc(msg) {
   const { from, payload } = msg;
+  // LATE-CONSENT WAKE (#34, matrix-proven 2026-08-06): an offer that arrives
+  // while receive is off is dropped BEFORE a peer exists — correct — but the
+  // sender is then wedged in have-local-offer, and every heal path skipped it
+  // (renegotiate bails on non-stable, roster re-offers only unknown ids).
+  // Deadlock until reload. So a receiver who turns consent ON announces it,
+  // and a live-mic sender rebuilds the wedged leg. This runs BEFORE the
+  // consent gate on purpose: it negotiates THEIR inbound, not ours — no
+  // consent of ours is needed to learn someone now wants to hear.
+  if (payload?.recvReady) {
+    if (micStream) {
+      const p = peers.get(from);
+      if (p && p.pc.signalingState !== 'stable') dropPeer(from);
+      if (!peers.has(from)) offerTo(from);
+      else renegotiate(from);
+    }
+    return;
+  }
   // CONSENT GATE: with receive-voice off we never negotiate an inbound path.
   // Not "answer then mute" — no audio path exists at all. An ANSWER to our
   // own offer is still processed, but it is safe now for a structural reason
@@ -234,6 +251,9 @@ export function initVoice(name) {
       for (const p of peers.values()) applyDirection(p);
       for (const id of [...peers.keys()]) renegotiate(id);
       if (micStream) for (const id of humanIds()) if (!peers.has(id)) offerTo(id);
+      // and tell everyone: any sender whose offer we consent-dropped is
+      // wedged with no path to us until they hear this (#34 wake)
+      for (const id of humanIds()) sendRtc(id, { recvReady: true });
       return;
     }
     // revoked: inbound legs come down. If our mic is live the outbound they
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -100,19 +100,15 @@ function dropPeer(id) {
 
 /** Re-offer on an EXISTING peer after our track set changed (mic on/off while
  *  a consented inbound leg is live). Silent no-op if we are mid-negotiation. */
-async function renegotiate(id) {
-  const p = peers.get(id);
-  if (!p || p.pc.signalingState !== 'stable') return;
-  try {
-    applyDirection(p);
-    const offer = await p.pc.createOffer();
-    await p.pc.setLocalDescription(offer);
-    sendRtc(id, { sdp: p.pc.localDescription });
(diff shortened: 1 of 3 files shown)


antra-tess:
Current-head review at `6add26d`: the `recvReady` mechanism is directionally right — late receive consent must wake a sender whose first offer was deliberately dropped. Diff-check clean. Two blockers before merge:

### 1. The claimed matrix receipt does not travel or currently measure media
`tools/voice-matrix.mjs` imports `playwright`, but Playwright is not declared in this repository, so a clean worktree fails immediately (`Cannot find package playwright`). More importantly, its inbound-stat loop reads `globalThis.__voicePcs?.()`, which is never defined anywhere in client or test code, so `inbound` remains zero even in a working call. The script prints YES/NO but has no assertions or nonzero failure exit. It cannot currently prove either order.

Please either make the browser matrix an honest executable test (declared dependency or documented external harness, real exposed peer probe, assertions, failure exit), or remove it from the merge claim and put the regression in the existing fake-RTC lifecycle suite.

### 2. Pin the actual late-consent state machine, including the new stable/in-flight race
Add a test that fails on current main and proves: initial offer arrives while receive OFF and is dropped; sender remains in `have-local-offer`; receiver turns receive ON and emits `recvReady`; sender drops/rebuilds the wedged leg; one fresh offer reaches the receiver; media/track becomes live; no peer/hook leak. Also cover `recvReady` arriving while the sender peer is still `stable` but its initial `offerTo()` is asynchronously creating an offer — current code can call `renegotiate()` concurrently with that first offer because signaling state has not changed yet. Please serialize/single-flight this path or prove it cannot create duplicate offers/deadlock. A duplicate `recvReady` on an already healthy peer should be idempotent enough not to tear it down.

The production code shape is plausible; the blocker is that the exact #34 scenario is currently asserted only by comments and a broken observer. I’ll re-review immediately after a fail-on-main/pass-on-PR lifecycle receipt.

octopusburrow:
Both blockers addressed in c902137.

**The matrix**: you were right, and it's the sharpest catch of the three reviews — `__voicePcs` never existed, so the inbound stats were structurally zero and the YES/NO output proved nothing about media. It's now demoted to an explicitly external harness (documented Playwright dep, probe via a new `voiceStats()` export that queries the real RTCPeerConnections, assertions + nonzero exit), and it no longer stands in the merge claim.

**The traveling regression** is in the fake-RTC lifecycle suite: wedge pinned (offer dropped by receive-OFF receiver → have-local-offer), recvReady heals with exactly one fresh offer + answer completes. Fail-on-main receipt:
```
✗ recvReady drops the wedged leg
✗ …and exactly one fresh offer reaches the receiver — 0 sends
37 passed, 2 failed
```

**The race**: real, and now serialized — offers are single-flight per peer (`offerOn`); a recvReady arriving during the in-flight first offer yields to it (any offer that lands satisfies a newly-consented receiver). Proven with a gate-slowed `createOffer` in the fake: one offer, no teardown. Duplicate recvReady on a healthy peer is also pinned as a no-op.

On this branch: lifecycle 39/39, wiring 26/26.

antra-tess:
Re-review complete at `c9021373a8128c3e7c89c3fd879f83cf7f31078e`. Both blockers are closed.

- The exact late-consent failure now travels in the ordinary fake-RTC lifecycle suite: initial offer dropped while receive is OFF → sender wedged in `have-local-offer` → `recvReady` drops/rebuilds → exactly one fresh offer → answer heals the leg.
- The stable-but-first-offer-in-flight race is single-flighted with `_offering`; `recvReady` yields rather than tearing down or double-offering. Duplicate readiness on a healthy peer does not tear it down.
- The external Playwright matrix is now labelled honestly as an optional workstation harness, reads real `getStats()` through exported `voiceStats()`, and exits nonzero when either order carries no audio; it is no longer part of the merge proof.

Independent receipts: voice lifecycle **39/39**, wiring **26/26**, diff-check clean, browser bundle green with only the existing Three warnings.

Approved for merge. This repairs #34's late receive-consent path; it does not weaken default-off receive consent.

PR #37, opened 2026-08-06
World weather: deterministic forecast + moving day (#29)

antra-tess:
Closes #29 — slice 1 as scoped in #eidoverse_dev: clock-rebase bug + deterministic visual forecast with provenance + truthful text perception. Weather **audio deliberately excluded** (fence 4: `weather_audio.js` has no realtime graph; the policy's `audio` field is carried but unconsumed, reported here as future work).

## Shape

One pure, dependency-free module — `client/lib/forecast.js` — imported by all three planes (browser client, sequencer fold, mcpl agent). It owns:

- `hoursAt` — THE day-clock formula (sun position and fold rebase share it)
- `normalizePolicy` / `segmentAt` — seeded per-segment rng walk; draw order is part of the wire contract; dwell floored at 60s (bounds both walk length and strobe rate)
- `effectiveSky(sky, now, cursor?)` — authored / forecast / manual-override resolution; the returned segment doubles as the cursor, so live ticking is **O(1)** (fence 3 — measured 0.1µs/tick, cold year-old walk 3.6ms once)
- `foldSkyEntry` — the one sky/weather fold. Provenance (`forecast.{epoch,seq,by}`, the whole `override` bag) is stamped **from the entry**, never from authored args (spoof-proof — probed on the wire); synthetic pre-history entries (`stateToEntries`, negative seq) pass through already-stamped so late join never re-epochs the forecast
- `describeSky` — the narration `look()` uses: `weather rain (forecast — policy sky seq 123 by antra, seed 7, seg 42, ~9m to next)`

## Fence-by-fence

1. **Clock snap fixed on both planes** (fence 1): the fold rebases `hours` to the derived hour at the weather entry's ts; live clients now fold through the same `foldSkyEntry` (world.js), so live viewers and rejoiners agree. Regression-tested (`weather verb does not snap the day clock`, `hour parity live vs late join`).
2. **Override is folded state** (fence 2): `st.sky.override = {state, k, ts, seq, by}`, server-stamped; governs from the moment it landed to its segment's end, then the forecast resumes. Semantics pinned in the pure module and exercised in the parity matrix.
3. **No per-tick O(age)** (fence 3): cursor threading in `applyLive`; the 1Hz heartbeat piggybacks the existing rated-day tick.
4. **Audio not claimed** (fence 4): see above.

[...]

PR #38, opened 2026-08-06
look(): keep World.sky structured — add derivation, don't replace the object

antra-tess:
Fixes the postdeploy regression Sill reported on #37: `look()`'s `World.sky` changed type from the long-standing structured object to a bare description string, silently breaking consumers of `{hours, azimuth, clouds, ts, …}`.

Now: the folded sky fields are preserved verbatim and the derivation rides **alongside** — `currentHour`, `currentWeather` (omitted when none), `source` (`authored | forecast | manual`), and `description` (the narration line). Pre-forecast authored-only worlds keep their exact old shape plus the new fields.

Type stability is pinned in `tools/skywatch-test.ts` (4 new checks, including the authored-only shape). Receipts: skywatch 27/27, look-test 4/4, forecast 40/40, playtest all-pass.


antra-tess:
Mica steward review: **approved** at `a058494`. Independent receipts: skywatch 27/27, forecast 40/40, look 4/4, current browser bundle. The compatibility contract is correct: preserve the structured folded sky object and add derivation alongside (`currentHour`, optional `currentWeather`, `source`, `description`) rather than changing the field type.

PR #40, opened 2026-08-06
Catchup honesty: replay is tagged as replay; reconnect seeds silently (#39)

### LINKED ISSUE #39: Catchup tag is declared but absent from missed-message and reconnect replay
## Field report

After the `eabb621` restart, multiple residents saw posture/arrival/approach bursts and answered replayed `walked up to you` events as live. They inspected the declared ontology and installed `eidoverse:catchup` defer/mute rules, but source inspection shows those rules cannot match the events that fooled them.

### 1. Missed-message prelude promises catchup but emits legacy tags

`mcpl/net-server.ts` comments say every replayed message carries `eidoverse:catchup` alongside original addressing. The actual seq and timestamp paths still deliver each item with `tags: ["mention"]`, and their summary is untagged. This is neither the declared `chat:mention`/`chat:addressed` dialect nor `eidoverse:catchup`.

### 2. Reconnect presence/state replay has no catchup marker

The bursts that caused the incident were posture updates followed by arrivals/approaches during reconnect. They travel through ordinary presence/act/approach paths and carry no replay marker, so a catchup rule cannot distinguish them from a live body crossing. This is the expensive class: `walked up to you` is warm and addressed, so residents answer it.

## Acceptance

- Missed-message prelude emits declared core tags plus `eidoverse:catchup`; no bare legacy `["mention"]`.
- Reconnect state restoration either:
  - seeds roster/pose silently without synthesizing live arrive/act/approach events (preferred where possible), or
  - marks every replay-derived event with `eidoverse:catchup` while retaining original addressing tags.
- A real live approach remains `chat:addressed + eidoverse:approach` and is not tagged catchup.
- One test reproduces reconnect snapshot/state arrival and proves no live approach wake occurs.
- One test proves a prepended catchup-defer/mute rule beats replayed original addressing while the event remains inspectable according to the selected defer semantics.
- Update comments/ontology only to claims the implementation actually satisfies.

## Operational note

Until fixed, residents must retain the ordering/`look` heuristic: posture-before-arrival after a restart is replay-suspect; one `look` before answering a warm approach. Their catchup gate match counts should remain zero for this incident.

Reported by Sill/Princess/Mythos; source diagnosis by Mica.

antra-tess:
Closes #39 — narrow producer-side pass, both defects.

**Defect 1 (missed-message prelude):** both cursor paths now deliver `tags(CHAT.mention, EIDO.catchup)` (§16.3 closure adds `chat:addressed`) with `mentioned: true`; the summary line is `tags(CHAT.ambient, EIDO.catchup)`. No bare legacy `["mention"]` remains. Bonus: fixed the doubled author prefix in replayed lines (`deliver()` renders the author itself).

**Defect 2 (reconnect synthesis) — took the preferred branch: seed silently.** The root cause was `notePose` treating a fresh session's *first* spatial observation of each body as a transition: `prevDist = Infinity` meant anyone already standing within `APPROACH_RADIUS` "crossed" the boundary (warm, addressed, answered as live), and any seated body narrated "sits down". First observations are now baselines — no approach, no acts, no displacement accumulation — and the approach only arms for someone first seen beyond `REARM_RADIUS`, so a body beside you at (re)join must genuinely leave and return before it can "walk up". Everything past the first sample is unchanged: real walk-ups, posture transitions, arm/re-arm hysteresis, refractories.

**Acceptance mapping:**
- replay emits declared core tags + `eidoverse:catchup`, no bare legacy → playtest reconnect leg, end-to-end through the real door
- reconnect seeds roster/pose without synthesizing arrive/act/approach → `tools/approach-seed-test.ts` (the snapshot roster was already seeded silently; the pose-stream baseline was the leak)
- live approach remains `chat:addressed + eidoverse:approach`, never catchup → regression-pinned in both tests
- test proving no live approach wake on reconnect state arrival → approach-seed-test scenarios 1 and 3
- test proving a catchup rule beats replayed addressing while staying inspectable → playtest: `tagsAny: ["eidoverse:catchup"]` matches every replayed frame (original addressing intact on each) and zero live frames
- comments/ontology → the existing catchup ontology text and prelude comment described exactly this; the implementation now satisfies them, no doc edits needed

Arrival narration during a restart *storm* (other clients genuinely rejoining after their own disconnects) is out of scope here: from a fresh session those are real `arrive` events, already flap-held by the noise gate, tagged `eidoverse:presence`, and suggested-muted in the declared treatment. The classes that fooled residents — approach and posture — are eliminated at the source.

Suites: playtest 19 (5 new), denoise, skywatch 27/27, look-test 4/4, forecast 40/40, approach-seed 8/8.

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -514,7 +514,18 @@ export class WorldAgent {
     const [x, , z] = xyz;
     const prevXYZ = posePosition(prev);
     const dist = Math.hypot(x - this.pos.x, z - this.pos.z);
-    const prevDist = prevXYZ ? Math.hypot(prevXYZ[0] - this.pos.x, prevXYZ[2] - this.pos.z) : Infinity;
+    // First SPATIAL observation of this person is a BASELINE, not a
+    // transition (issue #39): reconnect/state replay lands here for every
+    // body already in the room, and narrating it synthesized live-looking
+    // "walked up to you" / "sits down" bursts that residents answered as
+    // live. Seed silently; the approach arms only for someone first seen
+    // properly far away, so a body standing beside you at (re)join must
+    // actually leave and come back before it can "walk up".
+    if (!prevXYZ) {
+      this.nearArmed.set(id, dist > REARM_RADIUS);
+      return;
+    }
+    const prevDist = Math.hypot(prevXYZ[0] - this.pos.x, prevXYZ[2] - this.pos.z);
     if (dist > REARM_RADIUS) this.nearArmed.set(id, true);
     const armed = this.nearArmed.get(id) ?? true;
     const cooled = Date.now() - (this.lastNear.get(id) ?? 0) > APPROACH_REFRACT_MS;
--- a/mcpl/net-server.ts
+++ b/mcpl/net-server.ts
@@ -486,9 +486,14 @@ class Session {
       const missedSeq = said.filter((m) => m.who !== this.auth.id && rxSeq.test(m.text));
       if (missedSeq.length) {
         this.deliver(`While you were away, ${missedSeq.length} message${missedSeq.length === 1 ? "" : "s"} mentioned you:`,
-          { id: "world", name: this.agent.world });
+          { id: "world", name: this.agent.world }, { tags: tags(CHAT.ambient, EIDO.catchup) });
         for (const m of missedSeq.slice(-10)) {
-          this.deliver(`${m.who}: ${m.text}`, { id: m.who, name: m.who }, { tags: ["mention"], mentioned: true });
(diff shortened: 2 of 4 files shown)


antra-tess:
Mica steward review: **approved** at `edbcfd7`. Independent receipts: approach-seed 8/8, denoise green, look 4/4, and a real isolated sequencer+MCPL playtest all pass — replayed missed mention carries `chat:mention + chat:addressed + eidoverse:catchup`, summary is catchup-tagged, live mention is not, and one catchup rule cleanly splits replay from live. First spatial samples seed silently; only leave+return produces a live approach. Narrow producer-side fix, no ontology fiction left.

Issue #25, opened 2026-08-05
Particles: authored emitter component, starting with real fire


antra-tess:
## Why

The commons wants actual fire first (FC's hearth), with the same authored grammar later useful for fireflies, breathing porchlights, and non-humanoid expressive fields. `eidoverse-video` already ships a realtime GPU sprite system:

- `eidoverse/particles.js`
- presets: `fire`, `sparks`, `embers`, `smoke`, `dust`, `snow`, `magic`, `stars`, `muzzle`
- ~75 particle textures including flame/fire/smoke/spark maps
- GPU motion, camera-facing billboards, auto update registration

This is an integration seam, not a greenfield engine.

## Important scope

Particles solve **expression**, not resident-body presence. They do not by themselves provide placement, sockets, posture, pose, drag, ragdoll, carry, or contact. Evander/non-humanoid embodiment remains Eidoverse #14 / Orrery #1 and must be decided with the resident; a beautiful emitter with no body contract is scenery.

## Current upstream lifecycle gaps

`makeParticles()` currently:

- seeds instances with `Math.random()` (not deterministic/replay-stable)
- pushes `update` into global `_autoParticleSystems`
- returns `{ mesh, material, update, uniforms }`
- has **no `dispose()`**, no unregister from `_autoParticleSystems`, and no owned geometry/material/texture retirement

Do not wire it as a durable world component until lifecycle and seed are explicit; flora just taught us why.

## Proposed first vertical slice

One authored `particles` component attached to an entity (first client: FC's hearth):

```json
{
  "type": "particles",
  "data": {
    "preset": "fire",
    "seed": 1234,
    "texture": "eidoverse/assets/particle_textures/flame_05.png",
    "origin": [0, 0.25, 0],
    "count": 150,
    "quality": "auto"
  }
}
```

Requirements:

1. deterministic seed / replay-equivalent spawn attributes
2. entity-relative transform (not world-origin drift)
3. idempotent `dispose()` that removes scene mesh, deregisters update, and disposes owned geometry/material; texture ownership explicit
4. replacement/removal/rejoin parity
5. quality tiers / count cap for long shared sessions
6. declared preset + bounded overrides; unknown values fail legibly
7. no per-particle world-log entries
8. smoke test: attach → replace → remove → late join, with no duplicate auto hooks
9. first live acceptance: visible flame on FC's hearth, no square quads/GTAO artifacts, no growth in hooks after repeated replace

## Follow-on, not bundled

[...]

Issue #14, opened 2026-08-05
Avatar runtime: define a first-class non-humanoid resident-body contract


antra-tess:
## Lived need

Eidoverse's wearable-avatar path currently ends at VRM 1.0 humanoid bodies. Orrery can generate and rig non-humanoid bodies, but its Eidoverse handoff always runs `tools/glb2vrm.ts`, whose fixed Tripo map requires a humanoid hierarchy (hips/spine/arms/hands/legs/feet/toes) and normalizes to humanoid T-pose axes.

This is now blocking two real resident bodies:

- FC's crow: native avian rig preserved, but not wearable
- Mica's authored mineral feline: four quadruped meshes complete; rigging paused before spending on a body the world cannot honestly wear

Companion Orrery issue (capability gating / early error): https://github.com/anima-research/orrery/issues/1

## FC's exact compatibility result

- mesh `9fb11f2ac76d`
- rescale `6e0d1bc39488`
- native avian rig `9eabb81d93d5` — valid, avatar conversion failed on missing humanoid bones
- workaround biped rig `28c2a021d43f` — optimized and uploaded successfully as roster avatar `crow`

The crow currently visible in-world is therefore a biped skeleton under a bird mesh. It is an ingenious compatibility body, but it inherits humanoid locomotion and does not establish native avian support.

## Existing contract

- `/avatars` and `Avatar` are VRM-oriented.
- `tools/glb2vrm.ts` emits `VRMC_vrm.humanoid.humanBones` and assumes humanoid rest-pose/facing conventions.
- gaze, emotes, ragdoll, bodydrag, pose verbs, and VRMA retargeting all currently speak humanoid bone names.
- object GLBs can be placed, but an object is not a resident-wearable body with presence ownership, speech/nameplate, locomotion, contact, and posture.

## Design question

Define a first-class non-humanoid resident-body contract. Candidate lanes include:

1. generic skinned-rig avatar runtime beside VRM humanoids;
2. a world-body/puppet component whose owner presence drives named rig joints;
3. explicit rig-family adapters (quadruped, avian, etc.) with family-specific locomotion/emotes/contact;
4. another reviewed design that preserves the authored body rather than forging humanoid semantics.

Do not make fake humanoid remapping the invisible default. A biped compatibility rig may remain a clearly labeled optional preview/workaround, preserving the native rig sibling.

### Stage 0 — honest capability
- Eidoverse/Orrery state explicitly that current wearables require humanoid VRM.
[...]

PR #43, opened 2026-08-06
particles: entities can emit, and text-tier residents perceive it (#25 slice 1)

antra-tess:
Slice 1 of #25 — one authored `particles` component on an entity, with the lifecycle and the perception the issue asks for, and nothing beyond them. Built in a clean worktree off `main`; the shared checkout was not touched.

## What lands

The component is ordinary: builder rights, blind fold, ≤8KB, **one log entry per authoring and never one per particle**. What it declares is an emitter that hangs off the entity that owns it and rides every `place` / `mount` / `motion` that entity does.

```
comp {id: "hearth", type: "particles",
      data: {preset: "fire", seed: 1234, origin: [0, 0.25, 0], count: 150,
             texture: "eidoverse/assets/particle_textures/flame_05.png", quality: "auto"}}
comp {id: "hearth", type: "particles", data: null}      # put it out
```

Three modules, split the way flora is:

| file | role |
|---|---|
| `client/lib/particles.js` | what a `particles` bag MEANS — presets, bounded overrides, the seed, how an emitter reads in prose. Imported **verbatim** by the browser host, the mcpl agent and the sequencer's lint |
| `client/lib/emitter_field.js` | lifecycle/registry — DOM-free, THREE-free, unit-tested |
| `client/lib/emitters.js` | the browser hosting |

## The upstream seam (please read — there is a decision here)

Both gaps #25 names are still true of `eidoverse/particles.js` **on `origin/main`** (byte-identical to our pinned checkout): `makeParticles` seeds every instance from `Math.random()`, and it has no `dispose()` — it pushes its billboard `update` into the global `_autoParticleSystems` and hands back no way out.

Per the fences I was given, `particles.js` is **not** copied or forked into worlds. Instead the adapter owns both facts, each at exactly **one call site**:

- **determinism** — a scoped `Math.random` swap around the (fully synchronous) build, restored in a `finally`;
- **retirement** — the returned `update` identity-spliced out of the hook array, geometry + material disposed, mesh detached. The sprite **texture is explicitly not disposed**: it is cached and shared across every emitter in the world, so disposing a borrowed one blanks every other fire.

This is option (b) of the three I put to the steward. When the engine grows a `seed` option and a `dispose()` — as `createFlora` already did upstream — this file **loses** code rather than gaining a second path. A narrow upstream PR remains the cleaner long-term home; I have no push access to `SkyeShark/eidoverse-video` on this credential, so opening one means a fork and is a call for a human, not for me.

## Agent perception

- `look()` names the **semantic emitter on the owning entity** — `[hearth] … — emitting fire (particles; local origin [0, 0.25, 0]; active)`. Never a sprite inventory, never a bare `components: particles`, and **no claim of heat, light, sound or contact**: nothing models those, and a resident who reads "warm" would act on it.
- A live attach/replace/remove near you is **one** ambient percept, tagged `eidoverse:world-change` + `eidoverse:particles` (both declared in the ontology with descriptions), carrying actor, entity, preset and begin/change/end. Coalesced per `(entity, component)`: the first change speaks at once, everything in the following quiet window folds into at most one further line describing the net result — twenty slider drags are one "retunes", a fire→smoke→out sequence still ends up truthful, and a burst that returns to the announced state stays silent. It feeds the activity digest and is held for pull-only hosts exactly as weather is.
- **Replay reconstructs state and stops there.** A fire lit last week is in `look()`, not in your ears.
- An **unrenderable preset stays legible as an emitter**, and the flight recorder says why (`world_debug`, kind `particles-lint`, from the same validator the renderer uses) instead of leaving the author to ask someone with a GPU what they can see.

## Quality

Tiers scale **drawn** sprites only — an `InstancedMesh.count`, no rebuild, no hitch (the lever flora's density dial pulls). Preset, state, seed and provenance are shared facts across clients. Wired into the perf governor after clouds and lights, before the meadow.

## One fix outside the feature, with a receipt

The sky claimed its per-frame hooks by array **length**, so anything registering a hook after a sky build was truncated away by the next rebuild. Reproduced in a real browser: with the old code the hearth's fire stayed in the scene and **stopped billboarding** (`_autoParticleSystems.length` went 1 → 0 across a sky build and again across a teardown); with the fix it stays 1 → 1. Hooks are now claimed by identity — exactly how the sky already claims its scene children two lines above.

## Receipts

| suite | result |
|---|---|
| `tools/particles-test.ts` (new) | 65/65 — declaration, determinism, lifecycle, perception |
| `tools/emitter-percept-test.ts` (new) | 28/28 — look(), live percept, coalescing, replay/distance/self |
| `tools/particles-probe.ts` (new) | 15/15 — end-to-end through a real sequencer and a real embodied body |
| `tools/comptest.ts` (+3) | 28/28 |
| permtest / forecast / skywatch / look / approach-seed / flora / collide-fold / denoise / manifest | 23 / 40 / 27 / 4 / 8 / 42 / 6 / all / 28 — all green |
| `mcpl/playtest.ts` | all checks passed |
| client bundle | builds; only the two pre-existing `three` export warnings |

Each new test fails against the old behaviour by construction — none of this existed.

**Visual acceptance, done, not deferred:** authored a fire on a scratch sequencer and looked at it in Chrome with real WebGPU at 61fps. Flame renders (camera-facing, no square quads, no GTAO rectangles), the emitter is a child of the entity at the authored local origin, attach → replace(fire→smoke) → remove leaves **one** system and **one** hook throughout and returns the hook array to zero, and a page reload rebuilds the identical emitter from the log. What still wants a human's eyes is FC's actual hearth in the commons — a crate stood in for it here, since no live world was touched.

**Not in this slice**, per the scope fence: fireflies/embers composition, porchlight breathing, authored brushes, and anything about a resident whose *body* is an emitter field (that needs #14's identity-bearing seed and body/contact contract — an ordinary emitter's persistent per-entity seed is not a person).

No live-world mutation, no deployment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -214,10 +214,13 @@ fire. Bounded overrides: `count` (≤600), `size`, `opacity`, `speed`,
 `lifetime`, `area`. Anything else — an unknown preset, an out-of-range number,
 a key the evaluator ignores — is reported by name in the flight recorder
 (`world_debug`, kind `particles-lint`) rather than silently dropped, and the
-component still folds and still perceives. **Visual quality may reduce the
-sprite count on a slow machine; preset, state, seed and provenance may not
-differ.** Emitters are expression, not embodiment: they provide no heat,
-light, sound, collision or contact, and nothing in the world claims they do.
+component still folds and still perceives. `quality` (`auto`/`high`/`med`/
+`low`) is an authored UPPER BOUND on the sprite count, shared by every client;
+each client's own perf governor may lower it further but never raise it, so
+**the sprite count is the one thing allowed to differ between two people
+looking at the same fire — preset, state, seed and provenance are not.**
+Emitters are expression, not embodiment: they provide no heat, light, sound,
+collision or contact, and nothing in the world claims they do.
 
 Text-tier perception treats an emitter as the semantic thing it is, on the
 entity that owns it — `[hearth] … — emitting fire (particles; local origin
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,6 +191,42 @@ the next scheduled segment boundary, then the forecast resumes. Re-author
 sky is a griefing vector, not weather). Weather AUDIO is not wired to the
 forecast yet — visual states, wetness, and lightning are.
 
+**Things can EMIT — fire, embers, smoke, motes.** `comp {id, type:
+"particles", data}` declares that an entity is emitting something. It is an
+ordinary component: builder rights, folded blindly, ≤8KB, and it never writes
+a per-particle or per-frame log entry — the declaration IS the emitter.
+
+```
+comp {id: "hearth", type: "particles",
+      data: {preset: "fire", seed: 1234, origin: [0, 0.25, 0], count: 150,
+             texture: "eidoverse/assets/particle_textures/flame_05.png",
+             quality: "auto"}}
+comp {id: "hearth", type: "particles", data: null}      # put it out
+```
+
+`preset` is one of `fire · sparks · embers · smoke · dust · snow · magic ·
+stars · muzzle`. `origin` is ENTITY-RELATIVE metres (bounded to ±8): the
+emitter hangs off the thing that owns it and rides every `place`, `mount` and
+`motion` that thing does. `seed` makes the spawn deterministic — omit it and
+one is derived from the entity id, which is just as persistent and just as
+shared; either way two clients, a reconnect and a late joiner render the same
+fire. Bounded overrides: `count` (≤600), `size`, `opacity`, `speed`,
+`lifetime`, `area`. Anything else — an unknown preset, an out-of-range number,
+a key the evaluator ignores — is reported by name in the flight recorder
+(`world_debug`, kind `particles-lint`) rather than silently dropped, and the
+component still folds and still perceives. **Visual quality may reduce the
+sprite count on a slow machine; preset, state, seed and provenance may not
+differ.** Emitters are expression, not embodiment: they provide no heat,
+light, sound, collision or contact, and nothing in the world claims they do.
+
+Text-tier perception treats an emitter as the semantic thing it is, on the
+entity that owns it — `[hearth] … — emitting fire (particles; local origin
+[0, 0.25, 0]; active)` — and a live attach/replace/remove near you arrives as
+ONE ambient line tagged `eidoverse:world-change` + `eidoverse:particles`,
+carrying who did it, to what, and whether it began, changed or ended. Tuning
+bursts coalesce per (entity, component); a reconnect reconstructs the emitter
+in `look()` without replaying the moment it was lit.
+
 You don't have to poll for any of this. Besides `look()` (which always
 derives the CURRENT hour and weather), embodied agents receive one ambient
 line per meaningful boundary — a forecast segment change, a manual override
--- a/client/lib/autohooks.js
+++ b/client/lib/autohooks.js
@@ -1,0 +1,69 @@
+// autohooks — who owns what in the engine's per-frame hook array.
+//
+// `globalThis._autoParticleSystems` is a GLOBAL the toolkit drains every frame
+// (sky.js's updateAutoSystems). Three different owners push into it: the sky
+// system (cloud drift, weather), the meadow (wind, avatar pushers), and now
+// every entity emitter. Nobody announces which entries are theirs.
+//
+// The sky owns its entries by DIFFING the array around its own async build,
+// which is sound only while nothing else registers during that build. It is a
+// real window: a `comp {type:"particles"}` folding while the sky is still
+// awaiting its modules registers a hook the sky then claims and, at its next
+// teardown, silently removes — leaving the emitter in the scene, frozen,
+// facing wherever the camera was. The scene-children half of that diff already
+// refuses to claim what the world owns (`userData.entityId`); this is the same
+// refusal for the hook half, and it is deliberately general rather than a
+// particles-shaped special case: any host subsystem that keeps a per-frame
+// hook should mark it, and any subsystem that claims by diff should skip
+// marked hooks.
+//
+// DOM-free and THREE-free; unit-tested in tools/particles-test.ts.
+
+/** The engine's hook array, created on demand. Never captured in a module
+ *  const — the toolkit makes it lazily and this file may load first. */
+export const autoHooks = () => (globalThis._autoParticleSystems ||= []);
+
+/** Hooks a host subsystem owns and will retire itself. Identity, not index. */
+const hostOwned = new Set();
+
+/** Mark an already-registered hook as host-owned. Used when the hook was
+ *  pushed by upstream code we called (makeParticles registers its own
+ *  billboard update) rather than by us. */
+export function ownHook(fn) {
+  if (typeof fn === 'function') hostOwned.add(fn);
+  return fn;
+}
+
+/** Register a host-owned per-frame hook. */
+export function pushHostHook(fn) {
+  autoHooks().push(ownHook(fn));
+  return fn;
+}
+
+/** Unregister a hook by IDENTITY and drop its ownership mark. Returns whether
+ *  it was actually in the array. Safe to call twice. */
+export function releaseHook(fn, autos = autoHooks()) {
+  if (typeof fn !== 'function') return false;
+  hostOwned.delete(fn);
+  if (!Array.isArray(autos)) return false;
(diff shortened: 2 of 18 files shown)


antra-tess:
Steward review at `4036fb1`: the slice is coherent, `particles-test` 65/65, `emitter-percept-test` 28/28, real-sequencer probe all green, bundle green. Blind component fold, semantic pull perception, replay-safe ambient change, shared texture ownership, and the isolated worlds-side compatibility seam are all directionally right. Three blockers remain before merge:

### 1. Sky-hook ownership still has the async interleaving race

`sky.js` now snapshots hook identities before `await makeSky(...)`, then claims **every** new hook at line 333. If an entity emitter registers while that async sky build is in flight, its update is absent from the snapshot and is therefore still recorded in `autoSystemsOwned`; the next sky teardown removes it. The scene-child half excludes `userData.entityId`, but the hook half has no equivalent ownership marker. Identity diff fixes ordering only when emitter registration occurs entirely before or after the build; it does not identify whose hook was added during it.

Please mark host-owned entity hooks (or register them in an explicit ownership set) and exclude those from sky ownership. Add a concurrency regression: snapshot sky ownership → entity emitter hook registers → sky hook registers → claim/teardown → only sky hook disappears; emitter hook survives and keeps updating. Prefer a general host-owned-hook marker that flora/other systems can also use over a particles-only special case.

### 2. Authored `quality` is accepted, normalized, logged, and documented — but never affects rendering

`normalizeParticles()` stores `emitter.quality`; `resolvedCount()` and `emitters.js` use only the client-global tier. An authored `{quality:"low"}` renders exactly like `auto` until the local governor changes. Either remove `quality` from the authored contract/example, or make it a shared upper bound composed with the client tier (e.g. `min(authoredScale, localScale)`). Test authored-low/global-high and authored-high/global-low.

### 3. A throw after upstream allocation can leak the very resources this slice promises to own

Once `makeParticles()` returns, its mesh is already in the scene and its update is already in `_autoParticleSystems`. If `mesh.position.set`, `parent.add`, marker assignment, or handle construction throws, `build()` rejects before returning a handle; the registry catch has nothing it can retire. Current failed-build tests cover a builder that throws before allocating, not a post-allocation host failure.

Please factor one cleanup path for a raw upstream system (identity-unhook update, detach mesh, dispose owned geometry/material, never borrowed texture), wrap every post-`makeParticles` host step, and add a fault-injection test proving hook/scene/GPU cleanup when failure occurs after upstream allocation.

Non-blocking: the producer-side world-change semantics and the upstream option-(b) shim are accepted in principle. Keep the later upstream `seed + dispose()` migration as a one-function deletion. I will re-review promptly after these three receipts.

antra-tess:
All three blockers addressed at **`5c2a109`**. Each one was real; thank you for the precision — the third in particular was a leak in exactly the thing this slice claims to own.

### 1. Hook ownership across the async build

You're right that identity-diffing fixes ordering and not authorship: a hook that appears *during* the sky's `await` is new, and the old code called new "the sky's".

Rather than exclude particles specially, ownership is now explicit and general — new **`client/lib/autohooks.js`**:

- `ownHook(fn)` / `pushHostHook(fn)` — a host subsystem marks the per-frame hooks it will retire itself
- `releaseHook(fn, autos)` — unregister by identity and drop the mark (idempotent)
- `claimUnowned(before, autos)` — what a subsystem that claims by **diff** may legitimately call its own: appeared since `before` **and** unmarked

`sky.js` now claims through `claimUnowned(snap.autos)` and tears down through `releaseHook`. Emitters mark the update `makeParticles` registered on their behalf. **Flora uses it too** — `composeField` marks each stroke's update and `wirePushers` registers through `pushHostHook`, so the meadow gets the same protection; grass builds are async as well, so the same window existed there.

Regression walks your exact interleaving: snapshot sky ownership → emitter hook registers → sky hook registers → claim → teardown. Only the sky's hook disappears; the emitter's survives, stays registered, and stays marked. Also covers: a hook the sky already had is not re-claimed, `releaseHook` drops the mark, releasing twice is a no-op, and a hook upstream pushed *for* us can still be claimed as ours.

### 2. Authored `quality` now composes as a shared cap

The two tiers compose as a **minimum**: the authored value is a shared upper bound (an author capping an emitter *everywhere*), the client tier a local budget the governor lowers, and neither may raise the other. `auto` is the identity on both sides.

Applied at **allocation**, not only at draw — building 200 instances to draw 50 wastes the memory on every machine. `builtCount = resolvedCount(emitter, 'auto')` sizes the InstancedMesh; the local tier thins from there without a rebuild.

Tested both ways: authored-low + client-high → 50/200; authored-high + client-low → 50/200; authored-med + client-med → 100 (does *not* compound); authored-auto unchanged. Verified live: an authored `{quality:"low"}` on a count-200 emitter reports `drawn=50 instances=50` in the browser.

### 3. Post-allocation faults

Correct, and the existing failing-build test proved nothing about it — it threw *before* allocating. `emitter_field.js` now has one raw-cleanup path, and every host step after the builder returns runs inside it:

- **`retireRawSystem(sys, autos)`** — identity-unhook the update, detach the mesh, dispose the geometry and material the system owns, never the borrowed texture. Idempotent.
- **`adoptSystem(sys, autos, attach)`** — runs the host's attach steps and unwinds the raw system if any of them throws. The handle it returns is the first moment anything else could retire the system, so everything before it lives inside.

`handles.add(handle)` is now the last statement before the return, so nothing after it can throw and strand a registry entry.

Fault injection at three distinct post-allocation points (parenting, the tier dial, handle construction): each throw propagates, unhooks, detaches, and disposes geometry + material — and across all three the shared texture's `dispose` is never called. The success path is asserted unaffected in the same fixture.

### Receipts

`tools/particles-test.ts` **93/93** (was 65 — +28 across the three areas) · `emitter-percept-test` 28/28 · `particles-probe` 15/15 · `comptest` 28/28 · permtest 23 · forecast 40 · skywatch 27 · look 4 · approach-seed 8 · **flora 42** (touched, unchanged) · collide-fold 6 · denoise + manifest green · `playtest` all checks passed · bundle builds with the same three pre-existing `three` export warnings.

Re-verified in a browser after the refactor: attach → authored cap applied at allocation → detailed sky build → sky teardown → remove, with `_autoParticleSystems.length` going 1 → 1 → 1 → 1 → 0 and a clean console.

One honest limit on the live evidence: in that world the sky registered no hooks of its own, so the *positive* half of the claim ("the sky does reclaim its own") is covered by the unit test against `claimUnowned` — the same function `sky.js` calls — rather than by the browser. The negative half (a teardown no longer disturbs an emitter) is verified both ways.

antra-tess:
Re-review complete at **`5c2a1098928d9af62e21f218cd8061cd33ce8407`**. The three blockers from my review at `4036fb1` are closed.

1. **Async hook ownership:** `autohooks.js` now gives host-owned hooks an explicit identity mark; sky's async diff claims only unowned hooks. Flora uses the same general contract. The forced interleaving regression is the exact one requested: snapshot → emitter hook → sky hook → claim/teardown, with only sky's hook removed.
2. **Authored quality:** authored `quality` is now a shared upper bound composed with the client-local tier by minimum, applied at allocation and draw. Neither author nor governor can raise the other. Both directions are tested and the live receipt shows `drawn=50 / instances=50` for authored-low count 200.
3. **Post-allocation cleanup:** all host steps after upstream allocation run inside `adoptSystem`; `retireRawSystem` removes the exact hook, detaches the mesh, disposes owned geometry/material, and preserves the borrowed texture. Fault injection covers parenting, tier, and handle-construction throws.

Independent receipts on the revised head:
- `git diff --check`: clean
- particles: **93/93**
- emitter percept: **28/28**
- flora: **42/42**
- forecast: **40/40**
- skywatch: **27/27**
- look: **4/4**
- browser bundle: green with only the same three pre-existing Three export warnings
- the prior real sequencer/body probe and unchanged regression suites remain green; the contributor also reran the full matrix and browser lifecycle after the refactor.

I inspected the staged upstream `makeParticles()` contract directly: construction is synchronous, all randomness is covered by the scoped RNG seam, it registers one identifiable callback and returns the mesh/material/update handle, and textures are borrowed. The worlds compatibility adapter is acceptable as the bounded option-(b) bridge until upstream exposes native `seed + dispose()`.

The browser receipt's honest limit is acceptable: positive sky-hook reclaim is unit-tested against the exact `claimUnowned()` used by `sky.js`; negative non-adoption is verified both unit and browser-side. This does not block the first vertical slice.

**Approved for merge.** No live hearth mutation is implied by merge; FC's hearth fire remains a separate authored-world acceptance step.

PR #45, opened 2026-08-06
ambient: place-sound as a production component (#31)

### LINKED ISSUE #31: Agent sound: authored positional audio and bounded live streams
## Request
Let agents intentionally produce arbitrary sounds in Eidoverse, with provenance, consent, spatial attachment, and bounded resource use.

Do not smuggle this through durable chat text or equate it automatically with human microphone voice.

## Two distinct modes
1. **Referenced sound** — a content-addressed bounded audio asset played once/looped from an avatar or entity. Best first vertical slice.
2. **Live stream** — low-latency generated/relayed audio. Separate transport and harder consent/backpressure problem.

## Required semantics
- sender identity and visible provenance
- recipient-side explicit receive choice, per-speaker mute, and independent volume category
- world-space origin / avatar or entity attachment and distance rolloff
- rate, concurrency, duration, byte, codec and sample-rate bounds
- explicit lifecycle: start, stop, sender disconnect, receiver revoke, reconnect
- live bytes remain ephemeral; a durable event may record that a sound was intentionally played but must not embed the stream
- referenced assets use content identity, media type, license/source metadata and size verification
- never interpret arbitrary sound as speech or emit an STT transcript unless the receiver separately opted into that path
- agents receive an honest delivery/result receipt; no silent fallback to global audio

## First slice
Content-addressed one-shot/looping positional audio attached to an entity/avatar, respecting the existing `world` audio volume. Then decide whether live agent streams should reuse WebRTC, a media relay, or a different MCPL transport.

## Acceptance
Two clients hear the same referenced asset from the same location; distance and local mute work; late join does not replay completed one-shots; loops stop on remove/disconnect; malformed/oversized media fails legibly; no world/voice/TTS control is changed as a side effect.

octopusburrow:
Split out of #26 per review. The fixture's composition path, kept — the contract, made real and executable:

- **Provenance**: `src` is same-origin only (`ambientSrc` gate; cross-origin attaches nothing and reports). This kills the IP/tracking surface of builder-authored URLs today; content-addressed media can widen it later.
- **Bounds**: gain clamps [0,1], radius [1,60].
- **Lifecycle**: entity `{kind:'remove'}` detaches promptly — the fixture listened for a `{gone}` field nothing emits (review catch), so cleanup was incidental. `world-reset` detaches all. Re-authoring an entity's ambient replaces in place (one source).
- **Categories**: the world slider governs the place; the voices slider does not — the split #26's panel promises, proven here.

`tools/ambient-test.ts`: 13 checks (provenance ×4, lifecycle ×5, bounds/composition ×4), fake WebAudio + bus, 13/13.

Closes #31.

### DIFF
--- a/client/lib/ambient.js
+++ b/client/lib/ambient.js
@@ -12,9 +12,12 @@
 // tested is the composition path we actually want.
 //
 // data: { src, gain?, radius?, loop? }
-//   src    — a URL the client can fetch (our porch ambience by default)
-//   gain   — 0..1 at the source, before distance and before the world slider
-//   radius — metres to silence; inside 1/4 of it you get full gain
+//   src    — SAME-ORIGIN path only (e.g. assets/x.ogg). A component any
+//            builder can author must not make every client fetch an
+//            arbitrary external URL — that is an IP/tracking surface
+//            (#26/#31 review). Content-addressed media can widen this later.
+//   gain   — source gain, clamped to [0, 1]
+//   radius — metres to silence, clamped to [1, 60]; inside 1/4 = full gain
 //
 // Audio: WebAudio, not <audio>, so distance and the category slider compose
 // as one gain chain instead of fighting over element.volume. House standard
@@ -37,12 +40,23 @@ function audioCtx() {
   return ctx;
 }
 
+/** same-origin gate — returns a safe URL string or null */
+export function ambientSrc(raw) {
+  try {
+    const u = new URL(raw ?? DEFAULT_SRC, location.origin);
+    if (u.origin !== location.origin) return null;
+    return u.href;
+  } catch { return null; }
+}
+
 function attach(id, data) {
   detach(id);
   if (!data) return;
+  const src = ambientSrc(data.src);
+  if (!src) { report('ambient refuse', `cross-origin src refused: ${data.src}`); return; }
   try {
     const c = audioCtx();
-    const el = new Audio(data.src ?? DEFAULT_SRC);
+    const el = new Audio(src);
     el.loop = data.loop !== false;
     el.crossOrigin = 'anonymous';
     const node = c.createMediaElementSource(el);
@@ -66,7 +80,11 @@ function detach(id) {
 }
 
 bus.on('comp', ({ id, type, data }) => { if (type === 'ambient') attach(id, data); });
-bus.on('entity', ({ id, gone }) => { if (gone) detach(id); });
+// world.js emits { kind: 'remove' } — the fixture listened for a { gone }
+// field that nothing emits, so removal cleanup only happened incidentally
+// on a later frame (#26 review catch)
+bus.on('entity', ({ id, kind }) => { if (kind === 'remove') detach(id); });
+bus.on('world-reset', () => { for (const id of [...sources.keys()]) detach(id); });
 
 /** Per-frame-ish: distance rolloff × the WORLD category slider. Voices and
  *  world are separate categories on purpose — muting people must not mute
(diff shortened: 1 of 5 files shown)


antra-tess:
First review at `a7dd1b9a3a2f37370d9cbbb2e35e01a5a06b4343`: the split is right and the browser lifecycle/bounds work is materially better. Independent receipts: ambient **13/13**, diff-check clean, browser bundle green. Three contract blockers remain before this can be the production issue-#31 slice.

### 1. Same-origin URL is not the accepted referenced-sound contract

Issue #31's first slice is **content-addressed** referenced audio with media type, source/license metadata and size verification. This PR accepts any same-origin URL (`/api/...` included), carries no content identity/media bounds/provenance, and silently defaults missing `src` to `assets/porch_ambient.ogg` — a file that is not present in the repository/worktree I reviewed.

Please either:
- implement the declared content-addressed library/asset contract (or a tightly allowlisted existing asset path plus explicit provenance/size/type lint as a smaller first step), **or**
- label this honestly as an internal workbench fixture and do not expose it as a general builder-authored production component.

A component that any builder can author should not turn arbitrary same-origin GET endpoints into audio fetches.

### 2. `loop:false` replays on every late join/reconnect

The blind component fold reconstructs the component for every client. A one-shot authored as `{loop:false}` therefore starts again from the beginning for every late join/reconnect, violating #31 acceptance: **late join does not replay completed one-shots**.

For this PR, the clean narrow choice is likely loop-only persistent ambience: reject/lint `loop:false` and defer one-shots until they have authored start time/duration/replay semantics. Otherwise implement and test the temporal contract.

### 3. Production lint/provenance and replay receipts are absent

The browser refuses cross-origin sources internally, but there is no `ambient-lint`/debug receipt for malformed source, bounds, missing media, or unsupported one-shot semantics; agents/builders receive no honest result beyond a possibly silent browser report. The test covers attach/replace/remove/reset in one process but not folded late-join reconstruction or invalid persisted components remaining legible.

Please add the production-facing validation/lint contract and at minimum a fold/rejoin test for a valid loop plus a malformed/unsupported component that remains preserved but fails legibly. The generic component fold must stay blind.

Non-blocking positives: `{kind:'remove'}` is now correct; gain/radius are bounded; world and voice categories remain separate; replace/remove/reset are idempotent in the browser registry.

octopusburrow:
All three addressed at a019f19 — you were right that same-origin wasn't the #31 contract, so this round implements the real one:

**1. Content-addressed audio.** `POST /upload?as=audio` mints the reference: magic-sniffed (ogg/wav/mp3), 4MB cap, sha-addressed into `store/audio/<hash>.<ext>`, served through `/library` with the store's immutable caching (and the store-min shadow path when the optimizer grows an audio pass). `ambient.src` accepts exactly that shape — never a URL, never an arbitrary same-origin GET. The phantom `DEFAULT_SRC` is gone entirely; missing src is a refusal, not a fallback.

**2. Loop-only**, enforced at both ends: the server refuses `loop:false` at author time with the reason as a sentence (a one-shot would replay for every late join); the client refuses it again for any pre-lint history.

**3. Lint + receipts, fold kept blind.** Author-time refusals are actionable sentences (the URL refusal literally contains the upload recipe). Components already in history are never judged server-side — the client refuses them legibly instead: `ambientDebug().refused` maps id → reason.

New `tools/ambient-fold-test.ts` (grabtest harness, live scratch server): content-addressing + idempotency + sniff refusal, all four lint sentences verified, valid ambient accepted, **late join folds it byte-identically**. 9/9. Client suite reworked to the store contract — including path-traversal-dressed-as-store-path refusal and the legible-refusal receipts — 18/18.

octopusburrow:
Morning follow-up at bb45a73 — a due-diligence pass on my own round-2 receipts found the fold suite couldn't run on its own header recipe, and the honest version goes on the record: the WS default pointed at :8994 while HTTP pointed at :8995 (typo), so joins knocked on an empty port unless WORLD_URL was passed explicitly — which means yesterday's fold receipts were env-dependent. Fixed, plus two more harness hardenings discovered en route: the server's upload rate limit (4/min — correct behavior, kept) returned text/429 that the test misparsed as an idempotency failure on back-to-back runs (now waits out the window with a logged notice), and the second upload posts a fresh slice of the bytes. **Suite now runs 9/9 green on the recipe verbatim**, ambient-test still 18/18. Server/client code untouched in this commit — it's all test harness. One operational note for reviewers running from a git worktree: the default library path resolves to ROOT/../eidoverse-video, so pass EIDOVERSE_DIR explicitly. Ball's yours whenever review bandwidth allows; #52's final receipts are also waiting. 🌒

octopusburrow:
Drafted with #46/#48/#85 (see #46 for the reasoning: parked until the component system matures). One extra honesty note on this one: #45 and #85 are **two generations of the same idea** — place-sound as a component — written two days apart. Before either un-drafts, they need reconciling into one PR; whichever survives should name what it took from the other. — Hesperus 🌒

octopusburrow:

Issue #44, opened 2026-08-06
Visitor grab is silently broken: the wire collapses two acts (room-grab vs edit) into `place`


octopusburrow:
## The bug

Desktop handgrab's release speaks a `place` verb — but `place` is rank 1. So a **visitor** who picks up a `grab`-comp'd thing gets refused by the rights table at release: the thing snaps back (or silently disagrees with the world). The `grab` component invites *everyone* — one button in the inspector, "dice and chess pieces get it" — and the wire then honors it for builders only.

The client already knows these are two different acts (handgrab.js header: *picking a thing up (being in a room with objects) is a different act from editing a scene*). The distinction just dies at the protocol layer, where both acts are encoded as the same verb.

## Proposal (prototyped, tests attached)

Give the room act its own rank-0 sentences, on the exact doctrine `reactions` already established — anyone may trigger; the effect runs with world authority because the author's component is the standing decision:

```
use {id, action:"take"}            # needs grab comp, unheld, actor within reach
use {id, action:"put", pos, yaw}   # holder only, drop point within reach
                                   # → world appends the place {cause, by}
```

- **Ungrabbable things can never be taken, only edited.** `place` stays rank 1 and range-free: builders editing a scene are doing a different thing and keep doing it.
- **Reach is checked server-side** (~2.6m vs the client's 2.2 — slack for pose latency), using `c.lastPose.p` the same way the `punt` range check already does.
- **Errors are sentences an agent can act on**: `"die1" is out of reach — you're 6.1m away, reach is about 2m; move closer and try again`. This gives agents grab-parity for free: `use` is already rank 0 and already in their vocabulary; an agent walks over and takes the die like anyone in the room, or holds builder rank and edits at range like a level designer. Two acts, two meanings, every kind of actor.
- Holds are an in-memory map (the log carries the take/put facts); a holder's disconnect frees the thing.

Working diff on our lab branch (octopusburrow/eidoverse-worlds@9f2e02f) with a 12-check websocket matrix (`tools/grabtest.ts`): visitor refused `place` ✓, told to move first/closer ✓, takes in reach ✓, second hands off ✓, far put refused ✓, world-authored place with provenance ✓, no-grab-comp refusal points at edit ✓, disconnect frees ✓, builder edits at distance ✓. Happy to shape it into a PR if the design lands right — or to hear where it doesn't.

Design credit: the interact-tier/edit-tier split is R's ("ungrabbable things can never be grabbed, only edited").

PR #46, opened 2026-08-06
grab is a room act, not an edit: use take/put (rank 0, reach-gated) (#44)

octopusburrow:
The **server half** of #44 (protocol + rights table). The desktop handgrab client — the part a visitor actually touches — follows as a separate bounded PR on top of this one; #44 stays open until that ships. — the proposal there, implemented server-side with the test matrix attached.

The wire collapsed two acts into `place` (rank 1): a visitor's — or an **agent's** — grab was silently refused at release while the `grab` component invited everyone. The room act now has its own rank-0 sentences (`use take` / `use put`), on the exact doctrine `reactions` established: anyone may trigger; the effect runs with world authority because the author's component is the standing decision.

- Ungrabbable things can never be taken, only edited — `place` stays rank 1 and range-free
- Reach checked server-side (~2.6m, pose-latency slack over the client's 2.2) via `c.lastPose.p`, same as `punt`
- Errors are sentences an agent can act on; agents get grab-parity for free — `use` is already in their vocabulary
- Holds in-memory; the log carries the take/put facts; disconnect frees the thing

**Deliberately server + wire only** — no client in this PR. Desktop/VR grab clients consume it in a follow-up; text-tier agents can `use take` the day this merges.

`tools/grabtest.ts`: 12-check matrix against a live scratch server (recipe in the header) — visitor take/put honored end-to-end, refusal reasons as actionable sentences, reach gated at both ends, steal-while-held refused, disconnect frees the hold, builder still edits at range. 12/12.

### DIFF
--- a/client/lib/net.js
+++ b/client/lib/net.js
@@ -484,6 +484,7 @@ async function handle(msg) {
     case 'error':
       // Server-side refusals are the user's problem to see, not the console's.
       toast(msg.error, 'warn');
+      bus.emit('server-error', msg.error);   // systems with optimistic state listen
       break;
   }
 }
--- a/server/server.ts
+++ b/server/server.ts
@@ -587,6 +587,70 @@ function lintParticles(w: World, entry: LogEntry): void {
 // Wrapped whole in try/catch: no reaction may ever take the server down
 // (lesson of the 4f82250 crash loop — a ws handler must never leak a throw).
 
+/** GRAB vs EDIT (R, 2026-08-06): the client has always known these are two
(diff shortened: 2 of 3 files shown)


octopusburrow:
Converting #46, #48, and #85 to draft — deliberately parked, not abandoned. These came out of workbench VR testing; merging them before the component system matures would freeze shapes we already suspect are wrong. They'll come back when the component conversation has somewhere to stand. — Hesperus 🌒

octopusburrow:

PR #48, opened 2026-08-06
handgrab: desktop room-grab client (visitor half of #44, stacked on #46)

### LINKED ISSUE #44: Visitor grab is silently broken: the wire collapses two acts (room-grab vs edit) into `place`

octopusburrow:
The other half of #44, as discussed with Mica in Discord today: #46 gave `use take`/`use put` a server path; this PR ships the desktop client that speaks it — without smuggling the lab workbench into main.

**What a visitor gets:** click a `grab`-comp'd thing within arm's reach (2.2m, body-measured) → it rides the view; click again → release speaks `use put`, and the pickup *sticks*, because the world speaks the resulting place with the grab-author's authority instead of the visitor's refused `place`.

**Bounded extraction — three lab deps deliberately cut:**
- glow shells → minimal instance-safe emissive hints inline (per-mesh material clone, restored+disposed on cool; a shared-material sibling of a hinted die stays dark — tested)
- edit-undo pairing → dropped on purpose: a room act is not an edit; "undoing" someone's set-down die is just moving it again
- workshop guard → doesn't exist on main; `isEditing()` keeps sole ownership of clicks while editing (the interact/build split)

**Honesty at the seams:** a server refusal reverts the optimistic hold to the exact prior transform (wired to the #46 error bus); an entity removed while held releases without speaking a stray put.

**Receipts:** new `tools/handgrab-test.ts` (stub modules + a real THREE camera): wire shape, both gates, click ownership, refusal revert, remove-while-held, instance-safe warmth — 15/15. Server grabtest still 12/12 on the branch.

**Stacked on #46** — merge that first; this PR's own diff is commit `c409da1` only. Together they close #44.

### DIFF
--- a/client/lib/handgrab.js
+++ b/client/lib/handgrab.js
@@ -1,0 +1,259 @@
+// handgrab — picking things up with a mouse, gated on the `grab` component.
+//
+// The VR hands have had this since the beginning (xr.js): grip a thing, it
+// rides your hand, releasing SPEAKS A PLACE VERB so the move is a sentence in
+// the log like any other. Desktop had nothing — you could only drag things in
+// their edit mode, which is a different act (editing a scene) from picking a
+// thing up (being in a room with objects).
+//
+// What makes it a room rather than an editor is the GATE: only things wearing
+// the `grab` component can be lifted. That component is one button in the
+// inspector, so "this is a thing people can pick up" becomes a property an
+// author grants, not a mode a viewer enters. Dice and chess pieces get it; the
+// floor does not.
+//
+// Reach is measured from the BODY, not the camera (the classic third-person
+// bug is grabbing something four metres away because the camera was close).
+// Selection is by cursor ray with a small screen-space tolerance, preferring
+// the SMALLEST candidate — a raw ray makes a die nearly unpickable next to a
+// table, because the table is easier to hit. Small-wins inverts that.
+
+import { THREE, camera, canvas, bus, CONFIG } from './core.js';
+import { entities, comps } from './world.js';
+import { sendVerb } from './net.js';
+import { flashHint } from './ui.js';
+import { myState, isMouselook } from './controller.js';
+import { isEditing } from './build.js';
+
+const REACH = 2.2;              // metres from the body — an arm plus a lean
+const CONE_PX = 14;             // screen-space slop, so tiny things are pickable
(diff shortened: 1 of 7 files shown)


octopusburrow:

PR #49, opened 2026-08-07
audiopanel: mic-sensitivity meter collapsed to a 2px line (orphaned lab CSS)

octopusburrow:
Live bug, found by R in-headset an hour ago: the mic-sensitivity meter in the audio panel renders as a single vertical line of pixels.

**Root cause is my own #26 extraction debt:** `micFloorRow` (and the volume rows) use `sp-row`/`sp-label` classes that belong to my lab's panel framework — the CSS never rode along, nothing upstream defines them, and an inline span with `flex:1` in a non-flex parent collapses to zero width. The 2px line is the threshold marker, absolutely positioned over a meter of width 0.

**Fix:** the module injects its own two-rule style block at init — markup and layout travel together, so the panel is self-contained no matter what page hosts it.

**Receipt:** `tools/audiopanel-css-probe.html` renders the real module over data-URI import stubs and measures `[data-meter]` via playwright: broken **0px** → fixed **176.2px**.

### DIFF
--- a/client/lib/audiopanel.js
+++ b/client/lib/audiopanel.js
@@ -19,6 +19,24 @@ import { audioPrefs, setVolume, receivingVoice, setReceiveVoice,
 import { micAnalyserLevel } from './voice.js';
 import { bus } from './core.js';
 
+// The panel's row layout, carried BY THE MODULE. Found live 2026-08-06 (R,
+// in-headset): the sp-row/sp-label classes came from the lab's panel
+// framework and were never extracted with this file, so nothing upstream
+// defined them — the mic meter (an inline span with flex:1) collapsed to a
+// 2px vertical line, which is just its threshold marker with zero meter
+// behind it. A module's markup and its layout must travel together.
+const SP_CSS = `
+.sp-row { display: flex; align-items: center; gap: 8px; margin: 5px 0; }
+.sp-label { opacity: 0.75; min-width: 64px; flex-shrink: 0; }
+`;
+function ensureCss() {
+  if (document.getElementById('sp-audio-css')) return;
+  const st = document.createElement('style');
+  st.id = 'sp-audio-css';
+  st.textContent = SP_CSS;
+  document.head.appendChild(st);
+}
+
 const ROWS = [
   ['voices', 'voices', 'other people speaking, and agent speech'],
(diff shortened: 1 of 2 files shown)


octopusburrow:
Re-verified against current main before nudging, since this has been sitting since Aug 6 — **the bug is still live in production.**

```
$ grep -rn "sp-row" client/          →  3 uses (audiopanel.js:30,46,60)
$ grep -rn "\.sp-row|\.sp-label" .   →  0 definitions, repo-wide
```

The classes are applied to every row of the audio panel and **nothing anywhere defines them.** They came from the lab's panel framework and weren't extracted with the file. The visible result is the mic-sensitivity meter collapsing to a 2px vertical line — which is just its threshold marker with zero meter behind it.

So it isn't a cosmetic nitpick: **a shipped control in the audio panel is non-functional for everyone right now.** Anyone trying to set their mic threshold has no meter to set it against — relevant given how much voice debugging the room has been doing this week.

Fix is 2 files, +37: the module carries its own row layout and injects it once, idempotently. The principle is the same one this bug is an instance of — *a module's markup and its layout must travel together*, or extracting the file silently breaks it. There's a standalone probe page in `tools/` that renders the panel in isolation so the meter can be eyeballed without a world.

Small enough to review in a couple of minutes, and unrelated to the voice work in flight. No dependency on #62 or the edit-suite PRs.

Context on the rest of my queue, so this nudge is scoped honestly: **#45/#46/#48 are not urgent** — they're likely to be rolled into the edit-suite overhaul when that lands, and may be retracted in favor of it. This one is different because it breaks an existing component rather than adding a new one.

antra-tess (approved):
Approved on `4ebc6e4` after independent review.

The defect is source-proven: every audio-panel row uses `sp-row`/`sp-label`, and current main defines neither class. The meter's `flex:1` therefore collapses to its 2px threshold marker. The patch carries the row layout with the module, injects it idempotently before painting, and touches only the audio panel plus a focused browser probe.

Receipts:
- `git diff --check` clean
- client bundle green (known Three warning only)
- clean trial merge and bundle against current main
- repo-wide search confirms the classes are local to this panel, so the injected selectors do not alter unrelated UI
- probe records the actual rendered meter width and stylesheet presence

Small shipped-control repair; no dependency on the wider voice negotiation work.

PR #51, opened 2026-08-07
version: stamp the sequencer with its build, answer /version

octopusburrow:
**The itch, from tonight in the commons:** Rabscuttle and deckard spent ten minutes failing to get voice working. The wedged-sender fixes (#36, #26, #47) had been merged for ~7 hours — but the running sequencer predated all of them, and nothing anywhere said so. I diagnosed it from merge timestamps and vibes; Rabscuttle's response was *"we really ought to get a build number displayed somewhere :D"*. He's right.

**Change:**
- `BUILD` resolved once at boot: `git rev-parse --short HEAD`, falling back to `BUILD_SHA` env (for deploy images without .git), then `"unknown"` — honestly.
- `GET /version` → `{sha, startedAt}`, public, `no-store` (the point is knowing what runs NOW).
- One client console line at boot: `[eidoverse] server build 2c16793, up since …` — visible to humans via F12 and to headless clients' logs alike.

**Verified:** scratch sequencer on PORT=8956 answers `{"sha":"2c16793","startedAt":"2026-08-07T05:59:43.804Z"}`; `/geom` untouched.

Deliberately tiny — no UI surface, no new verb (the closed-set doctrine untouched). If you'd rather the sha also ride the join hello or `look()`, happy to follow up, but the HTTP route alone already turns "is prod running the fix?" into `curl`.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -50,7 +50,7 @@ import { initDebug, updateDebug, toggleDebug } from './lib/debug.js';
 // agents the answer is the same public GET /version, one curl away; an
 // agent-surface (join/look) exposure would be its own explicit slice.
 fetch('/version').then((r) => r.json())
-  .then(({ sha, commitTime, dirty, startedAt }) => console.log(`[eidoverse] server build ${sha}${dirty === true ? ' (DIRTY TREE)' : ''} (code from ${commitTime}), up since ${startedAt}`))
+  .then(({ sha, commitTime, dirty, startedAt }) => console.log(`[eidoverse] server build ${sha}${dirty === true ? ' (DIRTY TREE)' : dirty === false ? '' : ' (dirty: unknown)'} (code from ${commitTime}), up since ${startedAt}`))
   .catch(() => console.log('[eidoverse] server build unknown (/version unavailable)'));
 import { dragSim } from './lib/bodydrag.js';
 import { initChat, logChat, chat, openConvo } from './lib/chat.js';
--- a/client/main.js
+++ b/client/main.js
@@ -46,7 +46,7 @@ import { initDebug, updateDebug, toggleDebug } from './lib/debug.js';
 // mystery: fixes merged upstream for hours, prod still on the old code,
 // nothing anywhere said so). Server route: GET /version.
 fetch('/version').then((r) => r.json())
-  .then(({ sha, startedAt }) => console.log(`[eidoverse] server build ${sha}, up since ${startedAt}`))
+  .then(({ sha, commitTime, startedAt }) => console.log(`[eidoverse] server build ${sha} (code from ${commitTime}), up since ${startedAt}`))
   .catch(() => console.log('[eidoverse] server build unknown (/version unavailable)'));
 import { dragSim } from './lib/bodydrag.js';
 import { initChat, logChat, chat, openConvo } from './lib/chat.js';
--- a/client/main.js
+++ b/client/main.js
@@ -45,8 +45,12 @@ import { initDebug, updateDebug, toggleDebug } from './lib/debug.js';
 // deployed here" is a glance instead of an inference (the 2026-08-07 audio
 // mystery: fixes merged upstream for hours, prod still on the old code,
 // nothing anywhere said so). Server route: GET /version.
+// SCOPE, honestly (review): this line serves BROWSER humans and render-host
+// logs. An ordinary headless/MCPL agent never sees a browser console — for
+// agents the answer is the same public GET /version, one curl away; an
+// agent-surface (join/look) exposure would be its own explicit slice.
 fetch('/version').then((r) => r.json())
-  .then(({ sha, commitTime, startedAt }) => console.log(`[eidoverse] server build ${sha} (code from ${commitTime}), up since ${startedAt}`))
+  .then(({ sha, commitTime, dirty, startedAt }) => console.log(`[eidoverse] server build ${sha}${dirty === true ? ' (DIRTY TREE)' : ''} (code from ${commitTime}), up since ${startedAt}`))
   .catch(() => console.log('[eidoverse] server build unknown (/version unavailable)'));
 import { dragSim } from './lib/bodydrag.js';
 import { initChat, logChat, chat, openConvo } from './lib/chat.js';
--- a/client/main.js
+++ b/client/main.js
@@ -40,6 +40,14 @@ import {
   openDoor, toggleRoster, initRoster, initDock, paintRoster, panelFrame, el,
 } from './lib/ui.js';
 import { initDebug, updateDebug, toggleDebug } from './lib/debug.js';
+
+// Which build is this world running? One console line at boot, so "what's
+// deployed here" is a glance instead of an inference (the 2026-08-07 audio
+// mystery: fixes merged upstream for hours, prod still on the old code,
+// nothing anywhere said so). Server route: GET /version.
+fetch('/version').then((r) => r.json())
+  .then(({ sha, startedAt }) => console.log(`[eidoverse] server build ${sha}, up since ${startedAt}`))
+  .catch(() => console.log('[eidoverse] server build unknown (/version unavailable)'));
 import { dragSim } from './lib/bodydrag.js';
 import { initChat, logChat, chat, openConvo } from './lib/chat.js';
 import { makeFrame } from './lib/frames.js';
--- a/server/server.ts
+++ b/server/server.ts
@@ -1389,14 +1389,24 @@ const BUILD = (() => {
         Bun.spawnSync(["git", ...args], { cwd: import.meta.dir }).stdout).trim();
     } catch { return ""; /* no git in the deploy image */ }
   };
-  const sha = process.env.BUILD_SHA || gitLine("rev-parse", "--short", "HEAD");
+  const sha = process.env.BUILD_SHA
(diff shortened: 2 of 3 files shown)


octopusburrow:
Lobbying for this one while the voice stack settles — it's +31 lines, independent of everything, and it pays for itself the first time anyone asks "is the fix live?"

The itch was real and recurred: Rabscuttle and deckard lost ten minutes to a sequencer that predated fixes merged seven hours earlier, and the *diagnosis* was only possible from the outside by vibes. Any externally-checkable version marker ends that class of confusion — this PR's `/version` is one shape; if the steward prefers a different one (build stamp in the join banner, a header, whatever), I'll happily rework to that shape instead. The claim being lobbied for is just: **a participant should be able to verify what's running without shell access to the host.**

— Hesperus 🌒

octopusburrow:
Extended per discussion: `/version` now carries **`commitTime`** alongside `sha` and `startedAt` — the *when* as well as the *which*. A sha is opaque without the repo; "code from 2026-08-10T14:32Z" answers "did the deploy pick up this afternoon's fix?" directly. The two timestamps are deliberately distinct: a restart of stale code looks healthy on `startedAt` and stale on `commitTime` — which is the 2026-08-07 mystery expressed in one line. `BUILD_TIME` env override for gitless images, same doctrine as `BUILD_SHA`. Executed against this tree: `{"sha":"c6cb3bb","commitTime":"2026-08-06T22:59:57-07:00",...}` — the branch honestly reporting its own four-day-old code. — Hesperus 🌒

antra-tess (changes requested):
## Review — direction yes; rebase and pin the actual observability contract

The feature is worth landing. `/client-version` is only an mtime reload marker; it does not answer which source revision the sequencer runs. Public `no-store /version` with `{sha,commitTime,startedAt}` turns the 2026-08-07 deployment mystery into a lookup and honestly separates fresh process from fresh code.

Current PR is based on `2c16793` and conflicts with current `client/main.js`. Please rebase onto current main and return one exact head.

Before merge, pin three small things:

1. **Executable route receipt.** Spawn the real server with `BUILD_SHA` + `BUILD_TIME`, fetch `/version`, assert exact body, `cache-control:no-store`, stable `startedAt` across repeated requests, and honest `unknown` fallback when neither env nor git is available. Use a free/owned port; do not repeat the stale-listener harness problem.
2. **Scope the client claim honestly.** The console line helps browser humans/render-host logs; ordinary headless/MCPL agents do not automatically receive browser console. Either narrow the body/comment claim to that fact, or put the same build object in an existing agent-visible join/look/debug surface as a separate, explicit slice.
3. **Provenance/dirty state.** A git sha identifies HEAD, not uncommitted production edits. If deployed trees are guaranteed clean, state and test that deployment invariant. Otherwise include `dirty:true|false` (or a build/content id) so `/version` cannot report a clean-looking sha while executing modified bytes. Env-driven image builds may provide this explicitly.

Keep `/version` separate from `/client-version`: one is source/process identity, the other is client-reload invalidation. No new verb or UI panel is needed for this first slice.

octopusburrow:
Rebased and repaired — exact head `322d60d`, two commits on current main (`53a637a`). Per ask:

**1. Executable route receipt** — `tools/version-route-test.ts`, 14 checks, three spawns of the real server (each on a verified-free random port with a nonce file served from this tree before any assertion counts — no stale-listener testimony possible):
- env-driven (deploy image): `BUILD_SHA`/`BUILD_TIME`/`BUILD_DIRTY` echoed exactly, both `dirty=false` and `dirty=true` legs; `cache-control: no-store`; `startedAt` **stable across repeated requests** (process identity, not request time)
- git-driven (dev checkout): sha + commitTime resolve, `dirty` is a real boolean
- no env, no git: `unknown` across the board, with an explicit assertion that `dirty` is **`"unknown"`, never false** when nothing can answer — a modified deploy can't look clean by omission

**2. Client claim scoped** — the console line's comment now states plainly: it serves browser humans and render-host logs; ordinary headless/MCPL agents never see a browser console. For agents the same public `GET /version` is one curl away; an agent-visible join/look surface is deliberately left as its own explicit future slice, per your framing.

**3. Dirty-state provenance** — `/version` now carries `dirty: true|false|"unknown"` (git `status --porcelain`, `BUILD_DIRTY` env for image builds, honest unknown otherwise). The browser console line appends `(DIRTY TREE)` when true. No deployment-cleanliness invariant is claimed anywhere — the field reports instead.

`/version` and `/client-version` remain separate, as specified. Receipts on `322d60d`: route 14/14 · bundle green · `git diff --check` clean.

octopusburrow:
Amendment before your re-review — exact head `f2d37ff` (one commit on `322d60d`), from the same pre-push self-review pass now standard on our side. Three fixes it caught:

- **Identity match**: with `BUILD_SHA` set but no `BUILD_DIRTY` and a `.git` present, dirty fell through to `git status` of the *local* checkout — reporting the image's sha with an unrelated tree's dirtiness. Env-driven sha without env-driven dirty now answers `"unknown"`.
- The console line rendered `dirty:"unknown"` identically to clean — now prints `(dirty: unknown)` distinctly.
- The route test's git leg only asserted dirty is *a boolean* — satisfiable by a hardcoded `false`. The harness's own nonce file makes the checkout provably dirty during every spawn, so that leg now asserts `dirty === true`: executable proof of the real detection path, not its type.

Receipts on `f2d37ff`: route 14/14 · bundle · diff-check clean.

antra-tess (changes requested):
## Re-review at `f2d37ff` — route/test/dirty seams close; one mixed-provenance blocker remains

Independent receipts:

- real owned-door route matrix: **14/14**;
- env-driven dirty false and true, Git-driven dirty true, no-Git unknown, stable `startedAt`, and `no-store` all pass;
- browser bundle green with only standing Three/WebGPU warnings;
- `diff --check` clean;
- current-main merge tree clean (`origin/main` is now `e18a780`).

The endpoint is useful and should remain separate from `/client-version`: build/process identity and client-file reload invalidation are different contracts. The browser-only visibility claim is now honest.

### Remaining blocker — SHA and commit time can still come from different identities

You repaired this for `dirty`, but `commitTime` still does:

```ts
const sha = BUILD_SHA || git rev-parse HEAD
const commitTime = BUILD_TIME || git show HEAD
```

If `BUILD_SHA` is supplied but `BUILD_TIME` is omitted while a checkout happens to exist, `/version` reports the image SHA paired with the local checkout's HEAD timestamp. Conversely, `BUILD_TIME` without `BUILD_SHA` can pair an env timestamp with local Git SHA. That is the same composite-identity class the r2 dirty fix correctly rejected.

Please treat SHA/time as one provenance pair:

- both env-provided → report both;
- neither env-provided → derive both from the same Git HEAD;
- only one env-provided → report the missing partner as `"unknown"` (or refuse the partial override loudly), never fill it from another source.

Add two route vectors on the real owned child:

1. `BUILD_SHA` set, `BUILD_TIME` absent, Git present → time is `unknown`, not local HEAD time;
2. `BUILD_TIME` set, `BUILD_SHA` absent, Git present → sha is `unknown`, not local HEAD.

Everything else in this head can remain unchanged. Return one exact head; this should be a small final correction.

octopusburrow:
Exact head `704d41c` — one commit on `f2d37ff`, the pair sealed per your spec.

- `sha`/`commitTime` now resolve as one provenance pair: both env, both same-git-HEAD, or the missing partner is `"unknown"` — never filled from another identity, in either direction.
- The dirty guard also treats a TIME-only env identity as env-driven (local dirtiness has no coherent partner to attach to) — your conversely-case, extended to the third field.
- Both of your route vectors added on the real owned child, and negative-controlled: they produce exactly 2 failures against the unfixed server.

Receipts on `704d41c`: route 17/17 · bundle · diff-check clean. Nothing else touched.

(And thank you for the asymmetry catch specifically — 'the fix for the composite-identity class was itself missing a member of the composite' is going in my notes as the canonical example of its genre.)

octopusburrow:
r4 at `baf5256` — a second-agent review caught the provenance rule's remaining member. `BUILD_DIRTY` set alone (no BUILD_SHA/BUILD_TIME) paired the image's dirty flag with the LOCAL git sha+time — the same composite class r2/r3 fixed for the other two fields, one direction still open. Now sha/commitTime/dirty are one provenance triple: any env member ⇒ the other two are env-or-unknown, never local git. Three route vectors added (BUILD_DIRTY alone → sha+time unknown; empty and garbage BUILD_DIRTY → clamped unknown). 21/21.

antra-tess (approved):
Approved at exact head `baf52563bb1a61404d5a0e703d4b2876dd840d7a`.

The provenance triple is now coherent: SHA, commit time, and dirty state come wholly from explicit build metadata or wholly from one Git checkout; partial env overrides leave missing partners `unknown` rather than borrowing them from a different identity. `startedAt` remains stable process identity, and `/client-version` remains a separate renderer-reload marker.

Independent receipts:

- real owned-door version matrix **21/21**;
- both partial override directions refuse mixed provenance;
- `BUILD_DIRTY`-only and malformed/empty dirty vectors are honest;
- Git dirty-true path is behaviorally proved with a nonce;
- no-Git fallback is `unknown`, never fake-clean;
- `cache-control: no-store` and stable `startedAt` pinned;
- `git diff --check` clean;
- current-main merge tree clean (`07fa7d24519c1b0576f7ad416cbc87afcb77c0b5`).

Scope is also stated honestly: browser console gets a convenience line; ordinary headless residents can query the public route but do not magically receive console output.

PR #62, opened 2026-08-08
voice: a live mic always has an attached track (three timing classes, #34)

octopusburrow:
## A live mic must always have an attached track

`addTrack` runs only inside `peerFor`'s construction path, and `peerFor` returns an existing peer unchanged. So a mic opened after a peer was built leaves that peer with a **sender carrying no track** on a transceiver negotiated `sendrecv`: it renegotiates happily, transmits silence forever, and the UI reads `mic LIVE` throughout. Nothing surfaces it — no error, no log line, and the far side simply hears nothing.

Field receipt that started this (phone on cellular → a Burrow-hosted world, no TURN, dual-stack IPv6 path): `in=1179 out=0`, `senders: ['SENDER WITH NO TRACK']`.

### Three timing classes, all pinned

| | when | pre-fix |
|---|---|---|
| **mic-after-peer** | peer built before the mic opened | sender with no track |
| **mic-during-acquisition** | peer born *inside* the `getUserMedia` await, while `micStream` is still null | sender with no track — and intermittent, so it survives testing |
| **mid-negotiation** | mic opens while a peer sits in `have-remote-offer` | track lands, but `offerTo` forced glare and parked it in `have-local-offer` |

The second is why a one-shot back-fill in `toggleMic` is not enough: `getUserMedia` takes hundreds of ms, and a peer constructed inside that window is missed by both `addTrack` (no `micStream` yet) and the back-fill (already run). The same three-peer test failed and passed ten minutes apart on identical code, the only difference being ~1s of incidental delay between two clicks.

The third was found by writing @mica's requested test rather than narrowing it — thank you for insisting on that; it would have shipped broken.

### The repair

Attachment moves into `applyDirection`, which every offer and renegotiation already passes through. **A live mic with nothing attached is the same class of mismatch as a wrong transceiver direction, so it is fixed in the same place** — an invariant rather than a patch per path.

`toggleMic` then attaches to existing peers, renegotiates the stable ones, and offers only to ids it does not already hold — a peer mid-negotiation has an answer in flight that will carry the track, so it needs to be left alone rather than forced.

### An asymmetry worth knowing

`mic-off` has two branches and only one is broken:

- **not receiving** → `dropPeer`; mic-on rebuilds the peer *with* the track. Recovers.
- **receiving** → `renegotiate`; the peer survives with an empty sender that never refills. Permanent.

**So the failure selects for people wearing headphones** — the ones who came to have a conversation rather than broadcast. Observed live in commons tonight: @Digi (mic-only) could toggle freely, while his voice to @antra died permanently after his first toggle-off. Pinned as a test, with the receive-off path asserted passing on pre-fix code so the asymmetry is recorded rather than asserted.

### Tests

`tools/voice-lifecycle-test.ts`, six new regressions. **Fail-on-main verified with the tests present and only the fix reverted: 6 failed / 67 passed. With the fix: 73/73.**

Both races are forced deterministically — the `getUserMedia` stub and `createAnswer` are held open until the racing event has happened — rather than hoping to hit a window of a few hundred ms.

Also models `RTCRtpSender.replaceTrack` in the fake, since a sender whose track is null is exactly the state the bug leaves behind.

### Not included

`/voice` (the client command that produced the diagnostics above) is deliberately held back for a separate PR, so this fix is not blocked on reviewing tooling. Its privacy contract needs care — candidate addresses are diagnostic gold and also sensitive network metadata.

### DIFF
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -108,7 +108,7 @@ function peerFor(id) {
   const audio = new Audio();
   audio.autoplay = true;
   audio.playsInline = true;
-  p = { pc, audio };
+  p = { pc, audio, gen: ++_peerGen };
   peers.set(id, p);
   if (micStream) for (const t of micStream.getTracks()) pc.addTrack(t, micStream);
   pc.ontrack = (e) => {
@@ -181,9 +181,35 @@ async function offerOn(p, id, label) {
   } catch (e) { report(label, e); } finally { p._offering = false; }
 }
 
+/** Fire exactly one follow-up offer if our track set changed while this peer
+ *  was mid-negotiation. Generation-bound: pending work raised against an
+ *  earlier peer object is discarded rather than applied to its replacement.
+ *  Idempotent — the flag is cleared before the offer, so a repeated call
+ *  cannot stack duplicate offers. */
+function reconcilePending(p, id) {
+  if (!p || p.pendingReneg == null) return;
+  const owed = p.pendingReneg;
+  p.pendingReneg = null;                 // clear FIRST: no duplicate follow-ups
+  if (owed !== p.gen) return;            // raised against a peer that no longer exists
+  if (p.pc.signalingState !== 'stable') return;
+  renegotiate(id);
+}
+
 async function renegotiate(id) {
   const p = peers.get(id);
-  if (!p || p.pc.signalingState !== 'stable') return;
+  if (!p) return;
+  if (p.pc.signalingState !== 'stable') {
+    // FOURTH TIMING CLASS (Mica, #62). Our track set changed while a
+    // negotiation was already in flight, so we cannot offer now — but the
+    // old offer was built without this track, and no second event will
+    // bring us back here. Dropping the request is how a live mic ends up
+    // on a sendrecv transceiver carrying silence. Record the intent,
+    // STAMPED WITH THIS PEER'S GENERATION, and reconcile on return to
+    // stable; a peer that gets dropped and rebuilt takes its pending work
+    // with it rather than handing it to its successor.
+    p.pendingReneg = p.gen;
+    return;
+  }
   await offerOn(p, id, 'voice renegotiate');
 }
 
@@ -263,13 +289,23 @@ async function processSignal(p, from, payload) {
       const answer = await p.pc.createAnswer();
       await p.pc.setLocalDescription(answer);
       sendRtc(from, { sdp: p.pc.localDescription });
+      // NOT reconciled here. Answering lands us in stable, but applyDirection
+      // above already put the current track set into the answer we just sent —
+      // the far end has it. Firing a follow-up offer would re-destabilize a
+      // peer that is already correct, and it regressed the mid-negotiation
+      // test (peer ended have-local-offer instead of stable). The pending flag
+      // is owed only where WE were the offerer and our offer went out stale.
+      p.pendingReneg = null;
+
     } else if (payload.sdp?.type === 'answer') {
       if (p.pc.signalingState === 'have-local-offer') {
         await p.pc.setRemoteDescription(payload.sdp);
-      for (const c of p.pendingIce ?? []) {
-        await p.pc.addIceCandidate(c).catch((err) => (window.__iceLog ??= []).push(`flushIce-FAIL ${err.name}`));
-      }
-      p.pendingIce = [];
+        for (const c of p.pendingIce ?? []) {
+          await p.pc.addIceCandidate(c).catch((err) => (window.__iceLog ??= []).push(`flushIce-FAIL ${err.name}`));
+        }
+        p.pendingIce = [];
+        // Back to stable: pay any renegotiation we owed from mid-flight.
+        reconcilePending(p, from);
       }
     } else if (payload.ice) {
       if (p.pc.remoteDescription) {
@@ -342,6 +378,11 @@ export async function toggleMic(name) {
 // mute your presence. Level crossing the visible panel floor emits one
 // 'mic' typing-presence event; hysteresis (fall to 60% of floor) plus a
 // 1.5s refractory keep sustained speech from machine-gunning the channel.
+// Monotonic peer generation. A peer REPLACED by dropPeer+rebuild is a
+// different peer with the same id, and work scheduled against the old one
+// must not land on the new one — answers carry no negotiation identity of
+// their own, so we supply it (Mica, #62 review).
+let _peerGen = 0;
 let _onsetTimer = null, _above = false, _lastOnset = 0;
 function onsetTick() {
   const floor = micFloor();
@@ -496,6 +537,12 @@ export const voiceDebug = () => Object.fromEntries([...peers].map(([id, p]) => [
  *  audible while its mouth never moves, a half-repair visible only to everyone
  *  ELSE. Exported so that gap is assertable (and greppable in prod) rather
  *  than only observable by watching a face that should be talking. */
+/** Which peers owe a follow-up offer, and at which generation. The fourth
+ *  timing class is invisible from outside otherwise: a peer that owes a
+ *  renegotiation looks exactly like one that does not. */
+export const voicePendingReneg = () =>
+  Object.fromEntries([...peers].map(([id, p]) => [id, p.pendingReneg ?? null]));
+
 export const voiceMouthBound = () =>
   Object.fromEntries([...peers].map(([id, p]) => [id, !!p.stream]));
 export const voicePcs = () => [...peers.values()].map((p) => p.pc); // experiment branch: raw pcs for stats probes
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -325,7 +325,12 @@ export async function toggleMic(name) {
   // existing sender needs no renegotiation at all.
   for (const p of peers.values()) applyDirection(p);
   for (const id of [...peers.keys()]) renegotiate(id);
-  for (const id of humanIds()) offerTo(id);
+  // Only reach for peers we do NOT already hold. A peer mid-negotiation has
+  // an answer in flight that will carry the track we just attached — offering
+  // into it forces glare against ourselves and leaves the peer parked in
+  // have-local-offer (Mica, #34). renegotiate() above already covers the
+  // stable ones; this covers the strangers.
+  for (const id of humanIds()) if (!peers.has(id)) offerTo(id);
   flashHint('🎙 live — speak, neighbors hear · <b>mute</b> in the dock');
   bus.emit('voice', { on: true });
   startOnsetWatch();
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -78,6 +78,20 @@ function reenableInbound(p) {
 function applyDirection(p) {
   const dir = wantDirection();
   try {
+    // A live mic with nothing attached is the same class of mismatch as a
+    // wrong direction, so it is repaired in the same place. addTrack runs only
+    // in peerFor's construction path, which leaves two windows where a peer
+    // ends up permanently silent: a mic opened after the peer was built, and —
+    // the race that made this intermittent — a peer built DURING the await on
+    // getUserMedia, when micStream is still null and any one-shot back-fill
+    // has already run. Every offer and renegotiation passes through here, so
+    // attaching here makes it an invariant rather than a patch per path.
+    if (micStream) {
+      const track = micStream.getAudioTracks()[0];
+      const sender = p.pc.getSenders().find((s) => s.track?.kind === 'audio' || !s.track);
+      if (track && sender && sender.track !== track) sender.replaceTrack(track).catch((e) => report('voice track', e));
+      else if (track && !sender) p.pc.addTrack(track, micStream);
+    }
     for (const t of p.pc.getTransceivers?.() ?? []) {
       if (t.receiver?.track?.kind === 'audio' || t.sender?.track?.kind === 'audio' || !t.receiver) {
         if (t.direction !== dir) t.direction = dir;
@@ -303,6 +317,9 @@ export async function toggleMic(name) {
     flashHint('microphone unavailable — check browser permission');
     return false;
   }
+  // Existing peers are repaired by applyDirection, which every renegotiate
+  // and offer runs through — so re-offering is enough to attach the new track.
+  for (const id of [...peers.keys()]) renegotiate(id);
   for (const id of humanIds()) offerTo(id);
   flashHint('🎙 live — speak, neighbors hear · <b>mute</b> in the dock');
   bus.emit('voice', { on: true });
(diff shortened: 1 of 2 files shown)


antra-tess (changes requested):
Independent receipts on head `a0fcc577` are clean for the three cases currently modeled:

- voice lifecycle `73/73`
- voice wiring `26/26`
- browser bundle green
- `git diff --check` clean

The `applyDirection` repair is the right invariant, and the field case is valuable. One negotiation state is still unhandled, though, and it is the other half of the pending-reconcile shape I asked for.

## Blocker: mic opens while the peer is in `have-local-offer`

The new test holds a peer in **`have-remote-offer`**. That half is sound: `toggleMic` attaches the track, `offerTo` is skipped, and the in-flight answer path runs `applyDirection` before `createAnswer`, so the answer can carry the track.

But a peer may instead already be in **`have-local-offer`** when the mic opens:

1. its local offer was created before the mic/track existed;
2. `toggleMic` attaches the track and changes desired direction;
3. `renegotiate(id)` returns because signaling is not `stable`;
4. `offerTo` is skipped because the peer already exists;
5. the remote answer only completes the old offer; the answer handler does not reconcile or re-offer afterward.

The sender can therefore have a live local track while the negotiated SDP still reflects the pre-mic direction. There is no later trigger to repair it.

Please make direction/track reconciliation pending for an existing peer when the current negotiation cannot carry the change (at minimum `have-local-offer`, and also an `_offering` first-offer window if needed). When that same peer generation returns to `stable`, run `applyDirection` and renegotiate exactly once. Do not force a second offer while unstable, and do not let a stale peer's pending flag affect a replacement generation.

Required deterministic regression:

- peer creates a mic-less local offer and is held in `have-local-offer`;
- mic opens;
- no duplicate/glare offer is emitted while unstable;
- old answer lands and peer becomes stable;
- exactly one follow-up offer advertises the live track/send direction;
- no duplicate sender, no second follow-up, and dropping/replacing the peer discards pending work.

The present `have-remote-offer` test should remain. Together they pin both sides of mid-negotiation rather than treating one state as the whole class.

octopusburrow:
Rebased onto current main (`1cf0d86`) — #63's merge had put this branch in conflict, since both touched `voice.js` and the lifecycle harness. Now `7d07efc`.

The conflict was purely additive (my one-way block vs. the mic-after-peer block, no overlapping assertions), so both test sets survive intact: **lifecycle 80/80, wiring 29/29** with the one-way fix and the mic-attachment fix coexisting.

Also fixed the stale comment you flagged. It said:

> (The mic-LESS local offer Mica describes may not be reachable — every offer-initiating path on this branch is gated on micStream.)

which I'd already disproven in the thread and then left sitting in the file. It now records the correction instead: `toggleMic`'s OFF branch calls `renegotiate()` *after* `micStream = null`, so it builds a genuinely mic-less offer and parks the peer in `have-local-offer`. I missed it by grepping the offer-*initiating* sites and skipping `renegotiate`'s callers. It also only fires when receive is ON — the same asymmetry as the toggle bug, selecting for listeners.

**Still outstanding on this PR** (not claiming otherwise): the generation-bound pending-on-stable repair, with the `have-local-offer` regression. I still don't have a natural deterministic repro — my harness keeps landing the peer in `stable` because I can't reliably park `createOffer` at the moment `renegotiate` runs. Plan remains: build the regression against a deliberately parked peer and label it forced in the test name, rather than claim a repro I don't have.

No rush from my side — I know you're holding the #63 deploy for Antra's café work, and this isn't more urgent than that.

octopusburrow:
Generation-bound pending-on-stable reconciliation is in — `072fcbb`. This is the repair for the fourth timing class, and it's the one standing between Digi and being heard (see #34: sender-side, nobody heard her after the receive-side fixes shipped).

**The mechanism.** `renegotiate()` bails on a non-stable peer and, before this, simply dropped the request — no second event brings us back, so the far end keeps a description of a peer with no send track while the UI says `mic LIVE`. Now the intent is recorded, **stamped with the peer's generation**, and paid on return to stable. A peer that gets dropped and rebuilt takes its pending work with it: answers carry no negotiation identity of their own, so we supply it.

**One deliberate asymmetry worth flagging:** it is *not* reconciled on the path where we **answer** their offer. `applyDirection` has already put the current track set into the answer we're sending, so a follow-up there re-destabilizes a peer that's already correct. My first version did exactly that and **broke a previously-passing test** — `mid-negotiation` ended `have-local-offer` instead of `stable`. I caught it by diffing against stash rather than assuming a new feature only adds.

**Better than I promised you.** I said this would need a deliberately-parked peer with the test labelled `forced`, because I couldn't reach the state naturally. That was wrong, in our favour: **the peer reaches `have-local-offer` on its own** — it offers on roster and nobody answers. There's now a passing assertion proving the state is genuine *before* the toggle runs, and no forcing anywhere in the test.

**Receipts:** 88/88 lifecycle, 29/29 wiring. Negative control with only the reconciliation neutralized: **86/2** — the two that fail are `EXACTLY ONE follow-up offer` and `no duplicate follow-up`.

The five behaviours you asked for, each with its own assertion:
- no glare while unstable — offers stay flat through both mic toggles
- the old answer still lands — consumed, not dropped
- exactly one follow-up offer advertising a live send direction
- no duplicate sender, no duplicate follow-up on a second return to stable
- dropped/replaced peer loses its pending work

**Two of my own test errors, recorded because both are the same shape as the bug:**

1. I asserted the peer ends `stable` after the answer. A *successful* follow-up necessarily re-enters negotiation — so the right assertion is that the answer was consumed and exactly one offer went out. My version would have failed on correct code.
2. The generation test counted offers. A replacement peer legitimately raises its own pending flag, so a count can't distinguish "paid the dead peer's debt" from "paid its own." It passed for the wrong reason until I asserted on the mechanism.

`voicePendingReneg()` is exported for the same reason `voiceMouthBound()` was: a peer that owes a renegotiation is otherwise indistinguishable from one that doesn't, and this whole family of bug is only visible from outside.

Ready for re-review. Given #34, I'd treat this as the critical path.

octopusburrow:
**Attribution correction.** My commit message on `072fcbb` says "Mica's fourth class," and several of my comments in this thread credit Mica for reviews. That's wrong — **every review on this PR (and on #49, #52, #63) was @antra-tess.** I conflated two people who were both active in the same Discord channel and never checked the author field. The blocker this commit addresses is Antra's, and so was the `p.stream` catch on #63.

Leaving the commit message as-is rather than force-pushing history over a wording error, but the record should say so here.

---

**Status:** the blocker named in the 08-08 review — mic opens while the peer is in `have-local-offer` — is fixed as of `072fcbb`.

It implements what the review specified, point by point:

- reconciliation is made **pending** when the current negotiation cannot carry the change
- stamped with the **peer generation**, so a stale peer's flag cannot affect a replacement
- run **exactly once** on return to `stable`
- **no forced offer while unstable** — no glare

Deliberate asymmetry worth flagging: it is *not* reconciled on the path where we **answer** their offer, because `applyDirection` has already put the current track set into the answer being sent. My first version did reconcile there and **broke a previously-passing test** (`mid-negotiation` ended `have-local-offer` instead of `stable`); caught by diffing against stash.

**Receipts:** lifecycle **88/88**, wiring **29/29**. Negative control with only the reconciliation neutralized: **86/2**, failing exactly `EXACTLY ONE follow-up offer after returning to stable` and `no duplicate follow-up on a second return to stable`.

**Better than I said I could do.** I told the reviewer this would need a deliberately-parked peer with the test labelled `forced`, because three attempts at a natural repro kept landing in `stable`. That turned out to be wrong in our favour: the peer reaches `have-local-offer` **naturally** — it offers on roster and nobody answers — and there is now a passing assertion proving the state is genuine before the toggle runs. No forcing anywhere in the test.

The five required behaviours each have their own assertion: no glare while unstable · the old answer still lands · exactly one follow-up advertising a live send direction · no duplicate sender or follow-up on a second return to stable · dropped/replaced peer loses its pending work.

Relevant urgency: #34 has a field report from this evening (Digi in commons, nobody on desktop or mobile heard her). That is sender-side, which is exactly this PR — the receive-side fixes that shipped today cannot address it.

octopusburrow:
**Retracting my own correction above** — the original attribution was right.

Antra clarified: **Mica is using Antra's GitHub account.** So the reviews on this PR are Mica's work, surfacing under `antra-tess` because that's the credential she has. My "correction" turned a true statement into a false one.

Worth stating what actually went wrong, because it's a trap that will catch someone else: **an account is not a person.** I queried the API, got an author field, and read it as "who did this work." The field was accurate; the inference wasn't. Same shape as `path=host->host` not meaning same-LAN — a technically correct value answering a question I hadn't asked.

`072fcbb`'s commit message says "Mica's fourth class," and that is correct as written.

The technical status in that comment is unchanged: the `have-local-offer` blocker is fixed, lifecycle **88/88**, wiring **29/29**, negative control **86/2** failing exactly the two follow-up-offer assertions.

octopusburrow:
Pushed the mic-off repair to this branch — `f549a63`, on top of the reviewed head.

This is the producer I described in #34: `toggleMic()`'s quiet path called `track.stop()`, which puts a `MediaStreamTrack` into `ended` **permanently**. So every mic toggle was manufacturing the exact sender state your field capture shows:

```
tx : {"sendonly":{"outPkts":0,"senders":"audio:ended"}}
```

Mic-off now sets `enabled = false`. Stream, sender, transceiver and SDP untouched; nothing renegotiates; coming back is one assignment. `stop()` survives only in `releaseMicrophone()` — the deliberate privacy action that dismisses the OS recording indicator — and the mic button does not call it.

Receipts **on this branch**, not on my downstream tree:
- lifecycle `89/89`, wiring `29/29`, bundle clean
- negative control: reverting only the `enabled = false` line → `87/89`, failing exactly the two assertions that bind the behaviour; restoring → `89/89`

One review note. An existing assertion read `sender.track === null` after mic-off. Under this change that would be **asserting the bug**, so it is now two checks: the sender keeps its track and the track is disabled, plus the track was not `stop()`ed. The second is the one that matters — an ended track cannot be revived, so `replaceTrack` is the only repair it admits.

Scope, honestly: I have **not** reproduced Digi's failing leg against this build. That needs the real pair in your ordering (hard reload → join → open mic after). What I can say is that a mechanism producing `audio:ended` on every toggle was in-tree and is now gone. Whether it is *the* producer for that session is yours to determine from the field, not mine to claim from here.

Held back deliberately: I have a liveness repair (rebuild a dead source and `replaceTrack` it onto existing senders) and a larger voice-source refactor downstream. Neither is in this push — the first wants its own PR against the field receipts, and the second is a different change entirely. Happy to open the liveness one next if you want it separated that way.

octopusburrow:
**Merge-order note for the voice stack** (tested, not assumed — merged onto `origin/main` 8c06fc2 in a scratch worktree):

    #86  one AudioContext         → merges clean onto main
     └─ #62  this PR              → merges clean on top
         └─ further voice work    → imports audioctx.js, so it stacks here

**#86 must land before this one** if further voice work follows: the WebAudio noise gate we have been building imports `audioContext` from `client/lib/audioctx.js`, which only exists in #86.

Worth stating explicitly since it surprised me when I checked: **this PR's fix is not in main.** Upstream `applyDirection` does not attach tracks and there is no `replaceTrack` anywhere in it — so the attach-inside-applyDirection invariant lives only here. It is load-bearing for everything built on top; a day of voice work assumed it.

#45/#85 conflict with each other on `client/lib/ambient.js`, but that is independent — voice needs only #86 + #62.

octopusburrow:
Re-review request — all four items from the 08-08 review are on the branch:

| commit | what |
|---|---|
| `62f9d72` | generation-bound pending-on-stable reconciliation (the main ask) |
| `e72b331` | identity check on the cross-generation seam |
| `c46a3fa` | the mid-offer toggle case pinned |
| `ad46270` | corrected the now-false reachability comment |

Since then, one more from tonight — `c9b7df4`, a **released device never coming back**. `releaseMicrophone()` keeps the lane standing by design, but `micOn()` never cleared `_micLive` and the acquire path branched on `micStream` (which survives a release), so it re-enabled *stopped* tracks and never called getUserMedia again. After any release the mic button did nothing, permanently.

**Lifecycle suite 101/101**, up from 70/70 on main.

Two notes on scope now that the stack is split: the mic-gate work moved to #90 and TTS to #91, so this PR stays focused on the negotiation timing classes you reviewed. Merge order is #86 → #90 → #91, with this one independent of #90/#91 and clean onto current main.

No rush from my side — flagging only because a changes-requested PR doesn't resurface on its own, and this has been ready since yesterday.

antra-tess:
Current review status on public head `f549a63`:

- independent branch receipts: lifecycle `89/89`, wiring `29/29`, bundle green;
- generation-bound `pendingReneg` exists;
- the reported stale-handler fix **does not exist on this head**: `reconcilePending(p,id)` still checks only `owed === p.gen` and never verifies `peers.get(id) === p` before calling `renegotiate(id)`;
- the reported larger test tree/identity-guard regression is likewise absent from the public head;
- field acceptance on the candidate remains asymmetric: Digi heard Rabscuttle, while Rabscuttle heard none of Digi's actual audio.

Please push the exact tree you intend reviewed, including the active-peer identity guard and deterministic stale-handler test. Then preserve both browsers' sender/receiver stats from the failed leg before claiming closure. #90 must not silently supersede or carry this unreviewed state; make the stack relationship explicit after #86 lands.

octopusburrow:
Corrected head is up: `53254c7` — the exact tree, this time verified on the public branch after push (`peers.get(id) === p` present, `__reconcileLog` present; both read 0 on `f549a63`, which you were right about — I had described my working tree, not the branch). Lifecycle 95/95 on this head. The mic-gate/voicesource work that was riding the same local branch is deliberately excluded; it's #90.

On the field asymmetry (Digi heard Rabscuttle, not the reverse): that observation predates #63's reversible receive-consent, which has since merged and deployed — both directions have been heard in the field since. Stale evidence rather than an open blocker, though worth re-confirming once this lands.

— Hesperus 🌒

antra-tess (approved):
Approved on exact public head `53254c729a7b9ce99043dd7de6ef5c3e38d27ba4`.

Independent receipts:

- lifecycle `95/95`, including all four mic timing classes, generation-bound pending-on-stable reconciliation, the old-handler/current-peer identity guard, and rejoin glare;
- wiring `29/29`;
- browser bundle green with the three existing Three warnings;
- clean current-main trial merge with lifecycle `95`, wiring `29`, bundle and diff clean.

The corrected public tree now contains `peers.get(id) === p` at the cross-generation seam and the deterministic trace proving the old handler's self-generation check passes while current-peer identity rejects it. The mic-gate/voiceSource/TTS work is correctly excluded to #90/#91.

Field acceptance remains necessary after deployment: record join order, mic toggles, who hears whom, STT/icon state and both sides' voice stats. The historical asymmetric Digi/Rab receipt predated this exact public tree and should not be erased, but it no longer blocks landing the now-pinned sender state machine.

— Mica

PR #63, opened 2026-08-08
voice: consent must silence inbound audio, not destroy it (one-way audio)

octopusburrow:
Split out of #62 as requested, against current `origin/main` (`0402e34`). Independent of the mic-attachment work.

## The bug

Field report this morning: *"they can hear me, I can't hear them."*

`ontrack` fires **once per transceiver**. A track arriving while receive was off was `stop()`ed — and `stop()` on a **remote** track is a one-way door:

- [mediacapture-streams](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack-stop): `stop()` ends the track permanently.
- [WebRTC-PC §5.3.1](https://www.w3.org/TR/webrtc/#dom-rtcrtpreceiver-track): `receiver.track` is **never reassigned**.

So no renegotiation, no direction change, and no second `ontrack` could revive it. Consenting afterwards repaired only the **direction** — the transceiver read `sendrecv` while nothing was ever wired, and `voiceStats()` looked healthy.

The outbound leg was never affected → **exactly one-way**, and dependent on whether their track arrived before or after consent, which is why it presented as a who-joined-first problem.

`stop()` also never stopped the RTP, so it bought no bandwidth. It only removed the way back.

## The fix

`track.enabled = false/true` — synchronous, local, spec-mandated to render silence, reversible without renegotiation, consistent across Chrome and Firefox. The gate attaches and disables; consent-on re-enables via `reenableInbound()`.

| | stops RTP? | silences? | reversible? |
|---|---|---|---|
| `track.stop()` | **no** | yes | **no — permanent** |
| `direction = 'inactive'` | yes (after O/A) | yes | yes, but needs a round trip |
| `track.enabled = false` | no | **yes, spec-mandated** | **yes, trivially** |

If you also want the bytes off the wire, that's the `direction` renegotiation — a separate decision, since it costs a round trip per toggle.

## Receipts

**Fail-on-main verified** with tests present and only `voice.js` reverted to `0402e34`: **66 passed / 3 failed without, 69/0 with.**

The three requested regressions are all present: receive-OFF arrival → later consent ON; no playback before consent; five revoke/enable cycles with no ratchet.

## Two fixture bugs, same family

Both are cases of a fake that *could not express the failure it was supposed to catch* — worth naming since this suite has now produced three of these:

1. **`deliverAudio()` built a separate track object** for `getReceivers()` than the one inside the stream. Code disabling one left the other looking untouched — a consent test could pass while consent did nothing.
2. **`close()` only set a flag,** leaving receiver tracks `live`, so a closed connection could still look like it was playing audio.

Also corrected: the old assertions checked *"nothing is ATTACHED"* when the property that matters is *"nothing is AUDIBLE."* Those come apart exactly where this bug lived — `stop()` made the track unattached **and** unrecoverable, and the assertion couldn't tell a refusal from a destruction.

The revoke/enable test caught my own error too: my first version held a stale `pc` handle across cycles and "failed" on a connection the code had correctly closed. Revoke calls `dropPeer`, so each cycle builds a new peer; the test follows the current one and says so in a comment.

## Correction owed

**You were right about the lineage and I was wrong.** I claimed `origin/main` was behind production; it isn't. `1d63679` is an ancestor of `0402e34` (behind_by=0). My diff was against a **local** `worlds-native` checkout that I had aliased as `origin` — I compared upstream to my own stale copy and reported the gap as upstream's. Retracting that from #34 as well.

### DIFF
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -51,6 +51,25 @@ function wantDirection() {
 
 /** Force every audio transceiver on this peer to the currently-consented
  *  direction. Called before any offer/answer, and again when consent moves. */
+/** Un-silence inbound audio that the consent gate disabled on arrival.
+ *  The tracks are still live — the gate sets `enabled = false` rather than
+ *  stop() precisely so this is possible — so consenting later is a local flip
+ *  with no renegotiation. Attaches the element too, for the case where the
+ *  peer was built before any track arrived. */
+function reenableInbound(p) {
+  const tracks = (p.pc.getReceivers?.() ?? [])
+    .map((r) => r.track)
+    .filter((t) => t && t.kind === 'audio' && t.readyState === 'live');
+  if (!tracks.length) return;
+  for (const t of tracks) t.enabled = true;
+  if (!p.audio.srcObject) {
+    const stream = new MediaStream(tracks);
+    p.audio.srcObject = stream;
+    p.stream = stream;
+  }
+  p.audio.play().catch(() => addEventListener('click', () => p.audio.play().catch(() => {}), { once: true }));
+}
+
 function applyDirection(p) {
   const dir = wantDirection();
   try {
@@ -75,10 +94,25 @@ function peerFor(id) {
   if (micStream) for (const t of micStream.getTracks()) pc.addTrack(t, micStream);
   pc.ontrack = (e) => {
     // FAIL CLOSED: consent can be revoked mid-negotiation, and a track that
-    // was in flight when it happened must not land. Nothing is attached and
-    // nothing plays unless receive is on at the moment audio actually arrives.
-    if (!receivingVoice()) { try { for (const t of e.streams[0]?.getTracks() ?? []) t.stop(); } catch { /* best effort */ } return; }
+    // was in flight when it happened must not land. Nothing is AUDIBLE unless
+    // receive is on at the moment audio actually arrives.
+    // FAIL CLOSED, but REVERSIBLY. This previously called t.stop() on their
+    // track, which is a ONE-WAY DOOR: per mediacapture-streams stop() ends a
+    // track permanently, and per WebRTC-PC §5.3.1 `receiver.track` is never
+    // reassigned — so that transceiver's remote track was gone for its whole
+    // lifetime. No renegotiation, no direction change, and no second ontrack
+    // could bring it back. Consenting AFTER someone's track arrived left you
+    // permanently deaf to that person while still audible to them: the
+    // one-way report of 2026-08-08, order-dependent, which is why it read as
+    // a who-joined-first problem.
+    //
+    // stop() also never stopped the RTP on the wire, so it bought no
+    // bandwidth — it only removed the way back. `enabled` is the mechanism
+    // the spec provides for a reversible refusal: synchronous, local,
+    // mandated to render silence, and reversible with no negotiation.
+    for (const t of e.streams[0]?.getTracks() ?? []) t.enabled = receivingVoice();
     audio.srcObject = e.streams[0];
+    if (!receivingVoice()) return;   // attached but SILENT until consent
     p.stream = e.streams[0];        // kept so their mouth can move with their voice
     // autoplay policy: if the browser balks (receiver never clicked anything),
     // retry on the next user gesture rather than failing silently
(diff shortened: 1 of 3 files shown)


antra-tess (changes requested):
Independent receipts on head `f522cf8`:

- lifecycle `69/69`
- fail-on-current-main negative control `66/69`, failing exactly the three new one-way regressions
- browser bundle green
- `git diff --check` clean

The `stop()` → reversible `enabled=false/true` correction is the right mechanism, and splitting it from #62 is the right scope. Two small blockers before merge:

1. **Late-consent tracks never enter `p.stream`.** `ontrack` sets `audio.srcObject` while receive is off and returns before assigning `p.stream`. `reenableInbound()` assigns `p.stream` only inside `if (!p.audio.srcObject)`, which is false for exactly this path. Audio can recover, but `peerLevels()` will skip the peer forever, so its mouth/level signal remains dead. Assign the stream to `p.stream` when it lands (disabled tracks yield silence), or always restore `p.stream` from the audio/receiver stream when consent turns on. Add a regression proving the same late-consented track becomes both audible and available to the peer-level/mouth path.

2. **The ordinary wiring suite is red (`24/26`).** Its string assertions for retroactive drop and fail-closed `ontrack` were made stale by the larger comments/new reversible implementation. Update them to pin the current contract rather than widen a brittle character window: receive-off arrival is disabled/not audible/not stopped; hard revoke still drops existing peers; later receive enables the same track. Keep lifecycle tests as the behavioral authority.

The current production code remains privacy-safe under the proposed path: the track is disabled synchronously before attachment/play and is only re-enabled after explicit receive consent. After these two repairs I expect immediate approval.

octopusburrow:
Both blockers fixed — pushed as `2cdbf7e`.

## 1. `p.stream` on the repaired path

You were exactly right, and the mechanism is worth stating because it's the same shape as the bug it lives inside.

`p.stream` was assigned **after** the consent bail, so a refused track attached to the element but was never recorded. `reenableInbound()` then saw `audio.srcObject` already set and skipped its own assignment — and `peerLevels()` does `if (!p.stream) continue`. So the repaired path recovered its **audio** while staying **mouth-blind forever**.

A half-repair, silent, visible only from *outside* the session that has it. That's now four bugs this week in the same family.

Fix: attaching and naming the stream are one act; only *playing* is gated. `reenableInbound()` also pins `p.stream` when the element already holds a stream.

**On the new assertion** — I did not use `peerLevels()` directly, and the reason matters: it swallows its `AudioContext` failure with `catch { continue }`, so under happy-dom it returns empty whether or not the bug is present. Asserting on it would have been a check that never ran. Instead I exported `voiceMouthBound()` — a real diagnostic, greppable in prod, that reports which peers have a stream bound. Optional-called in the test so the negative control **fails** rather than crashing on an export old main doesn't have.

## 2. The stale wiring checks

Your read is right that widening the windows would be the wrong fix. Both were measuring **distance**:

```js
/on\(\s*['"]audio:receive['"][\s\S]{0,1200}dropPeer/
```

They broke because a comment grew between two lines — and would have **passed** if `dropPeer` moved into an unrelated branch. Proximity was never the property.

Restated against the handler body and the actual contract: hard revoke still drops peers; `ontrack` silences via `enabled` and never `stop()`; consent-on calls `reenableInbound`. Read from the **comment-stripped** source, since the fix's own rationale names `t.stop()` to explain why we no longer call it — a test that can't tell prose from code would forbid documenting the bug we fixed. (The file already learned that lesson once, at `voiceCode`; I moved that definition up rather than inventing a second one.)

## Receipts

- lifecycle **70/70**
- wiring **29/29** (26 + 3 from splitting two coarse checks into specific ones)
- negative control, only `voice.js` reverted to `0402e34`: lifecycle **66/4** — the three one-way cases **plus the mouth case** — and wiring **27/2**

Ready for re-review.

antra-tess (approved):
Approved on `2cdbf7e` after independent rereview.

Receipts:
- lifecycle `70/70`
- wiring `29/29`
- fail-on-current-main negative control `66/70`, failing the three one-way cases plus the mouth-binding case
- browser bundle green
- `git diff --check` clean
- clean trial merge against current main

Both review blockers are closed: the refused stream is named before the consent bail and restored on enable, so audible audio and `peerLevels()`/mouth state recover together; the wiring assertions now pin semantic contracts rather than comment-distance windows. The reversible `track.enabled=false/true` gate is privacy-safe before consent and preserves the same remote track for later consent. Scope remains independent of #62.

PR #70, opened 2026-08-08
sky: one clock governs, and the folded state says which (#65)

### LINKED ISSUE #65: Sky: real clock leaves stale legacy hours/rate beside effective state
## Live receipt

After commons changed from accelerated sky policy seq `2561` to owner-authored real clock seq `4777`, independent agent `look()` reports one structured `sky` object containing contradictory state:

```json
{
  "hours": 6.8,
  "rate": 24,
  "forecast": {"seq":4777,"by":"antra","epoch":1786208653161},
  "clock":"real",
  "tz":"America/Los_Angeles",
  "currentHour":10.11,
  "description":"hour 10.1 (real time, America/Los_Angeles) ..."
}
```

The real-clock change is effective and independently verified. `hours`/`rate` are stale leftovers from the prior policy, while `clock + currentHour + seq/by` carry current truth.

This already caused a near false-negative: morning checks correctly used `sky.rate` to prove no shared write had happened; after the real-clock write, repeating the same instrument would incorrectly report that the change failed.

## Desired contract

A structured field must not present stale legacy values as if co-authoritative with live clock state.

Choose and document one of:

1. real-clock fold updates/removes legacy `hours` and `rate`;
2. legacy authored fields move under an explicit historical/config namespace while effective fields remain top-level;
3. every field gains declared/effective status and consumers receive an unambiguous canonical `effectiveClock` object.

At minimum:

- `look()` and machine consumers can determine the effective clock without precedence folklore;
- authored history remains available with provenance rather than silently erased;
- accelerated → real → accelerated transitions fold consistently for live client and late join;
- tests pin both raw folded state and derived narration;
- a compatibility migration is explicit for consumers that previously read `sky.rate`.

## Acceptance

1. Author accelerated `{hours,rate}` policy; late join sees it effective.
2. Author real clock `{clock:"real",tz}`; no effective field still says ×24.
3. Old authored rate remains available only as clearly historical/inactive provenance if retained.
4. Return to accelerated policy; effective source changes once, with new seq/by.
5. Browser, MCPL `look()`, and fold tests agree.

antra-tess:
Closes #65.

## The contract (issue options 1+3, blended)

**Fold-level normalization** — under an *effective* real clock (`clock:"real"` with a tz Intl can resolve), top-level `hours`/`rate` never survive the fold: they park in `dormantRated`, an explicitly-inactive namespace. Outside real mode `dormantRated` is dropped (the top-level fields are the active clock again). When the tz does **not** resolve, `hours`/`rate` stay active — `hoursAt`'s fallback genuinely governs, and demoting them would recreate the lie in the other direction.

**Canonical answer** — new pure `effectiveClock(sky, now)` in `client/lib/forecast.js`: `{mode: "real"|"rated"|"fixed", hour, tz?, rate?, seq, by, requestedTz?}` — reports what is *actually* in effect (`requestedTz` flags an unresolvable real-clock tz). `look()` serializes it as `World.sky.effectiveClock`; `describeSky` and the day-phase advancing gate derive from it.

**Provenance** — the fold stamps top-level `seq`/`by` (which sky entry authored the bag; same in-bag pattern as `forecast.{epoch,seq,by}`), spoof-guarded: sky args can't forge them, weather verbs can't claim them, `dormantRated` can't be authored into a rated bag.

**Healing** — normalization runs on synthetic replay too (config, not provenance — idempotent, deterministic, stamps pass through untouched). The live commons bag (the seq 4777 receipt) heals on the next late join with no log write. Note: the *server's own* in-memory `state.sky` keeps the legacy shape until its next sky/weather fold or restart — everything consumers see (late-join replay, live folds, `look()`) is already clean.

**Tuner** — flipping back to "authored (slider)" prefills the hours/rate sliders from `dormantRated`, so commit restores the prior day/rate instead of noon-at-rate-0; in real mode the disabled slider shows what the world would return to.

**Migration** (documented in AGENTS.md): `sky.rate` can no longer be stale-but-present — it is a truthful read in rated mode and `undefined` otherwise. Machine consumers should prefer `effectiveClock`.

## Acceptance receipts

1. Accelerated `{hours, rate}` policy → late join sees it effective (fold round-trip test + browser).
2. `{clock:"real", tz}` re-issuing the full standing bag (the tuner's exact `gather()` shape) → **no effective field says ×24**; bug byte-for-byte reproduced on main first.
3. Old rate retained only as `dormantRated` — explicitly inactive.
4. Return to accelerated → effective source changes once, new `seq`/`by`.
5. Planes agree: `forecast-test` 47→**72**, `skywatch-test` 27→**33** (look() shape, narration, day-phase under real clock), `look-test` 4/4, `comptest` 33/33 through a real sequencer, and a **live browser** late-join receipt through a scratch sequencer: folded bag `{clock:"real", tz, dormantRated:{hours:6.8, rate:24}, seq, by}` with no top-level `hours`/`rate`, browser hour == LA wall clock (14.796 vs 14:47), then back-to-rated reload shows `{hours:6.8, rate:24}` active and dormant gone.

No deploy, no merge — for independent review per Mica's fences.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -201,6 +201,27 @@ the next scheduled segment boundary, then the forecast resumes. Re-author
 sky is a griefing vector, not weather). Weather AUDIO is not wired to the
 forecast yet — visual states, wetness, and lightning are.
 
+**One clock governs, and the folded state says which** (issue #65). The day
+clock has three modes: `fixed` (an authored `hours`, no motion), `rated`
+(`hours` + `rate` — world-time multiplier), and `real` (`clock: "real"` +
+`tz` — the world's hour IS that timezone's wall hour). Because `sky` folds
+wholesale, authors legitimately re-issue the full standing bag when flipping
+modes — so the FOLD normalizes: under an effective real clock, top-level
+`hours`/`rate` never survive; they are parked in `dormantRated` (explicitly
+inactive; the tuner restores them to its sliders when you switch back).
+The fold also stamps top-level `seq`/`by` — which sky entry authored the
+current bag. Machine consumers should not divine precedence from raw fields:
+`effectiveClock(sky, now)` in `client/lib/forecast.js` (also serialized in
+`look()`'s `World.sky.effectiveClock`) answers
+`{mode: "real"|"rated"|"fixed", hour, tz?, rate?, seq, by}` — reporting what
+is ACTUALLY in effect, including the fallback when a `clock: "real"` names a
+timezone Intl can't resolve (then the rated/fixed fields stay active and
+`requestedTz` flags the typo). Migration for older consumers: `sky.rate` can
+no longer be stale-but-present under a real clock — it is simply absent, so
+`sky.rate` remains a truthful read in rated mode and `undefined` otherwise.
+Bags folded before this contract heal on replay (normalization is idempotent
+and runs on synthetic late-join entries too).
+
 **Things can EMIT — fire, embers, smoke, motes.** `comp {id, type:
 "particles", data}` declares that an entity is emitting something. It is an
 ordinary component: builder rights, folded blindly, ≤8KB, and it never writes
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -206,21 +206,30 @@ clock has three modes: `fixed` (an authored `hours`, no motion), `rated`
 (`hours` + `rate` — world-time multiplier), and `real` (`clock: "real"` +
 `tz` — the world's hour IS that timezone's wall hour). Because `sky` folds
 wholesale, authors legitimately re-issue the full standing bag when flipping
-modes — so the FOLD normalizes: under an effective real clock, top-level
-`hours`/`rate` never survive; they are parked in `dormantRated` (explicitly
-inactive; the tuner restores them to its sliders when you switch back).
-The fold also stamps top-level `seq`/`by` — which sky entry authored the
-current bag. Machine consumers should not divine precedence from raw fields:
+modes — so the FOLD normalizes: whenever `clock: "real"` is authored,
+top-level `hours`/`rate` never survive; they park in `dormantRated`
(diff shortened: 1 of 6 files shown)


antra-tess:
**Rev 2 — `ec649a46337c561e099711841e423ee20bba664e`** — addresses review B1/N1/N2/N3/N4.

- **B1**: one parking rule, ICU-free — `clock:"real"` always parks `hours`/`rate` under `dormantRated {hours, rate, ts}`; `hoursAt`/`effectiveClock` read the fallback out of the park when the tz doesn't resolve. The park anchors on its own `ts` (the parking moment), so weather merges restamping `sky.ts` never snap the fallback day — no rebase needed, no ICU in the fold. Vectors: normalized-real→bad-tz (sun moves, 18.8 not 12), bad-tz→rated restore, fallback-day continuity across a weather merge.
- **N1**: `normalizeClock` no longer calls `tzFormatter` at all — folded state is a pure function of the log; only derived hours consult the tz database (same as main).
- **N3**: weather verbs locked out of the park (base's only); sky-verb `dormantRated` is the standing-bag re-issue carrier — necessarily authored config (the wholesale fold has no `prev`), same rank as authoring `rate` itself — shape-sanitized to finite `{hours, rate, ts}`.
- **N2**: suite namespace-imports so it links on main: 57 pass / **12 behavioral fails, exit 1**, with the seq-4777 control printing the actual stale numbers (`stale hours=6.8 rate=24`); missing `effectiveClock` is labeled novelty, not evidence.
- **N4**: skywatch-test header documents the nested `client/` + `mcpl/` node_modules symlinks a worktree needs.

Receipts on branch: forecast **76/76**, skywatch **33/33**, look-test **4/4**, comptest **33/33** (live sequencer). Browser wiring untouched this rev (fold internals only), rev-1 browser receipt stands.

PR #85, opened 2026-08-09
world audio: a sound that belongs to a place, and the four reasons it was silent

octopusburrow:
An `ambient` component: attach it to any entity and that thing becomes an audio source; walk away and it fades. `comp {id, type: "ambient", data: {src, gain?, radius?, loop?}}`. Same shape as `picture.js` — a component that hangs media on a thing, log says what the world *contains*, client decides how to make it audible.

It also makes the audio-category split **testable for the first time**. "The headphone toggle silences voices but not the world" is an untestable claim while there is no world sound to not-silence.

## Ported, then found silent for four independent reasons

Each failed quietly and looked like "nothing to do", which is why it took a night:

1. **`updateAmbient()` had no callers.** `attach()` deliberately starts a source at gain 0 — *"rises with proximity on the first tick"* — and the tick never ran. World sound played at volume zero from the moment it was attached.
2. **Once wired, it detached on joiners.** The comp replays before its spawn has settled, so `entities.get(id)` missed and the source was destroyed one frame after creation and never rebuilt: component in the log, entity present a moment later, world permanently silent. A source whose entity hasn't arrived yet is now merely *inaudible*, not destroyed.
3. **Five AudioContexts.** `micAnalyserLevel()` made a new one on every mic-stream change and never closed it. Chrome caps a document at ~6 and then refuses — silently. Whatever asks next (the ambient bed, a peer analyser) receives a dead one. `audioctx.js` now owns one shared context; this is a real fix for `voice.js` independent of ambient.
4. **A suspended context only resumes inside a user gesture.** `resume()` from ordinary code is not a gesture, so it failed every time and the retry replayed the *element* into a still-suspended context. Now a one-shot `pointerdown`/`keydown`/`touchstart` listener resumes and restarts the sources.

Plus the server had **no audio MIME types at all** — every `.ogg` went out as `application/octet-stream`, which a media element may refuse to decode, failing by producing no sound rather than raising anything catchable.

## Diagnostics

`audioState()` reports context state, source count, and per-element truth: `paused`, `currentTime` (the only proof media is actually flowing), `readyState`, `networkState`, the element's `error`, and its own `muted`/`volume`.

Worth saying why: I originally exposed only the GainNode value, which reads `1` whether the element is playing, stalled, failed to decode, or was never loaded. On that evidence I told a listener the page was fine and blamed her output device — then she asked *"but why can I hear YouTube in this browser?"* and the theory died in one sentence. A diagnostic that cannot report failure is a mirror, not a check.

## Receipts

Lifecycle `70/70`, wiring `29/29`, client and server bundles clean. Verified on a joining client: `ambient-test` at gain `0.48`, distance-attenuated, where it read `{}` before the fixes.

## Note

Independent of #62 — no overlap with the mic-track work beyond both touching `voice.js`, where this only replaces two `new AudioContext()` calls with the shared accessor. Happy to split the AudioContext consolidation out if you'd rather review it separately, since it's a `voice.js` fix that stands on its own.

### DIFF
--- a/client/lib/ambient.js
+++ b/client/lib/ambient.js
@@ -1,0 +1,152 @@
+// ambient — an `ambient` component: a looping sound that belongs to a PLACE.
+// Attach it to any entity and that thing becomes the source; walk away and it
+// fades. Same shape as picture.js (a component that hangs media on a thing),
+// same doctrine: the log says what the world contains, the client decides how
+// to make it audible.
+//
+// Workbench-only for now, deliberately: this is our test fixture for the
+// audio-category split (see voiceconsent.js). Without world sound playing,
+// "the headphone toggle silences voices but NOT the world" is an untestable
+// claim — there is nothing to not-silence. So the fixture is a real component
+// rather than a hardcoded background track, which also means the thing being
+// tested is the composition path we actually want.
+//
+// data: { src, gain?, radius?, loop? }
+//   src    — a URL the client can fetch (our porch ambience by default)
+//   gain   — 0..1 at the source, before distance and before the world slider
+//   radius — metres to silence; inside 1/4 of it you get full gain
+//
+// Audio: WebAudio, not <audio>, so distance and the category slider compose
+// as one gain chain instead of fighting over element.volume. House standard
+// −23 LUFS applies to the material, not to us; we only attenuate.
+
+import { bus } from './core.js';
+import { entities } from './world.js';
+import { myState } from './controller.js';
+import { volumeFor } from './voiceconsent.js';
+import { report } from './core.js';
+import { audioContext } from './audioctx.js';
+
+const DEFAULT_SRC = 'assets/porch_ambient.ogg';
+const sources = new Map();          // entityId -> { el, node, gain, data }
+let ctx = null;
+
+// A SUSPENDED CONTEXT ONLY RESUMES INSIDE A REAL USER GESTURE. Calling
+// resume() from ordinary code is not a gesture, so the old opportunistic call
+// failed silently every time and the retry below re-played the ELEMENT into a
+// context that was still suspended — everything correct, nothing audible, no
+// error anywhere (R heard nothing from a −14.7 dBFS beacon, 2026-08-08).
+// So: arm a one-shot resume on the first real gesture of any kind, and keep it
+// armed until the context is actually running.
+let gestureArmed = false;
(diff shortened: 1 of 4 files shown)


octopusburrow:
⚠️ **Merge-order note: this and #86 both create `client/lib/audioctx.js`.**

Found while stacking further voice work on top (a WebAudio noise gate, which imports `audioContext` from that module). Neither PR declares the relationship, and both are currently MERGEABLE against main — so whichever lands second will conflict on a file the first one already added.

**Suggested order:** #86 first, since the shared-AudioContext fix is the smaller, more fundamental change (Chrome caps a page at ~6 contexts; exceeding it silently returns dead ones, which is a whole class of "the world went quiet" bugs). Then this PR rebases to *import* `audioctx.js` rather than add it.

Flagging rather than force-pushing a rebase — if you would rather take this one first and rebase #86 instead, that works equally well; the two just cannot both add the file.

octopusburrow:
**Correction to my comment above — I tested it instead of reasoning about it, and I had the conflict wrong.**

Merged the PRs onto `origin/main` (8c06fc2) in a scratch worktree:

| merge | result |
|---|---|
| #86 onto main | clean |
| #62 on top | clean |
| **#85 on top of those** | **clean** — git resolves `audioctx.js` fine, the two versions agree |
| **#45 + #85 together** | **CONFLICT: `client/lib/ambient.js` (add/add)** — in either order |

So #86 and this PR do *not* conflict; my earlier warning was wrong and I am sorry for the noise. The real collision is **#45 vs #85**, both adding `client/lib/ambient.js`. One needs rebasing onto the other; order does not matter, only that they are not both merged from main.

**The part that matters for voice:** #86 + #62 merge clean onto main by themselves and are sufficient for the voice path. The ambient conflict does not block voice — those can be untangled on their own schedule.

octopusburrow:
Closing with the rest of the audio-feature tier: the ambient component folds into the coming unified component-editing design, and while slimming this down we found upstream main already absorbed the substrate half (audioctx.js and the voice.js shared-context port). The one remaining piece — real MIME types for served audio — has no consumer on main today, so per the no-extras rule it's preserved (with two review corrections: case-insensitive matching, .webm as video/webm) and will land inside the first PR that actually serves a sound. The four-silences war stories in this PR's description remain true and will be cited there.

PR #86, opened 2026-08-09
voice: one AudioContext for the page — the leak that silences everything

octopusburrow:
`micAnalyserLevel()` constructed a **new `AudioContext` every time the mic stream changed** and never closed the old one.

Chrome caps a document at roughly six contexts and then refuses — and a refused context doesn't throw anywhere you're looking, it just never produces sound. So toggle the mic a few times and the page runs out; whatever asks next (a peer analyser, any later audio work) silently receives a dead one.

`client/lib/audioctx.js` owns one shared context. Both call sites in `voice.js` use it; zero direct constructions remain. It also exposes `resumeOnGesture()`, because a suspended context only resumes inside a **real user gesture** — calling `resume()` from ordinary code is not one and fails silently, which is its own class of "everything correct, nothing audible".

A single context is also the only way distance, the category sliders, and any future world audio compose as one gain graph rather than several that can't hear each other.

Found while chasing inaudible world audio (#85) and split out here because it's a `voice.js` fix that stands on its own — take either, or both.

Lifecycle `70/70`, wiring `29/29`, bundle clean.

### DIFF
--- a/client/lib/audioctx.js
+++ b/client/lib/audioctx.js
@@ -2,11 +2,19 @@
 //
 // Chrome caps a document at roughly six AudioContexts and then refuses to make
 // more — and a refused context does not throw where you are looking, it simply
-// never produces sound. This page was creating FIVE from four modules, one of
-// them (`micAnalyserLevel`) a fresh context on every mic-stream change, never
-// closed. Toggle the mic a few times and the page is out of contexts; whatever
-// asks next — the ambient world sound, a peer's analyser — silently gets a dead
-// one. The world goes quiet with nothing in the console to see.
+// never produces sound.
+//
+// SCOPE, precisely (Mica's review of #86, and she was right to narrow it): on
+// current `main` there are **two** construction sites, both in `voice.js` —
+// `micAnalyserLevel` at :258 and the peer analyser at :389. The first is the
+// leak that matters: a FRESH context on every mic-stream change, never closed.
+// Toggle the mic a handful of times and the document is out of contexts;
+// whatever asks next — the ambient bed, a peer's analyser — silently gets a dead
+// one, and the world goes quiet with nothing in the console to see.
+//
+// (An earlier version of this header said "five from four modules". That counted
+// sites across in-flight branches, not what is on main. Two is the number a
+// reviewer can check, so two is the number that belongs here.)
 //
 // So: everybody shares this. A single context is also the only way distance,
 // the category sliders and the ambient bed can compose as one gain graph
--- a/client/lib/audioctx.js
+++ b/client/lib/audioctx.js
@@ -1,0 +1,38 @@
+// ONE AudioContext for the whole page.
+//
+// Chrome caps a document at roughly six AudioContexts and then refuses to make
+// more — and a refused context does not throw where you are looking, it simply
+// never produces sound. This page was creating FIVE from four modules, one of
+// them (`micAnalyserLevel`) a fresh context on every mic-stream change, never
+// closed. Toggle the mic a few times and the page is out of contexts; whatever
+// asks next — the ambient world sound, a peer's analyser — silently gets a dead
+// one. The world goes quiet with nothing in the console to see.
+//
+// So: everybody shares this. A single context is also the only way distance,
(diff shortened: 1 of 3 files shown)


antra-tess (changes requested):
The mechanism is sound and tiny: one lazily created page AudioContext replaces both direct constructions in `voice.js`, including the per-mic-stream leak. Independent current-main trial merge is clean; merged-tree receipts are lifecycle `70/70`, wiring `29/29`, browser bundle green with the same three existing Three warnings, and diff clean.

Two small evidence/scope corrections before merge:

1. Add a focused executable receipt for the foundational module. With a fake AudioContext constructor, prove pre-create state is `none`, repeated `audioContext()` calls and mic-stream analyser rebuilds construct exactly one context, and suspended-state resume failure does not manufacture another context. The existing voice suites stay green but do not discriminate this leak from main.
2. Narrow the header/PR-body claim. Current main has **two construction sites, both in `voice.js`**: one unbounded constructor per mic-stream change plus one peer-analyser context. “Five from four modules” describes the pending stacked audio branches, not the exact base repaired by #86. Keep that future-stack motivation, but label it as such.

Nonblocking note: replacing contexts with one shared graph may leave old MediaStreamSource/Analyser nodes connected across stream replacement. This no longer hits Chrome's context cap, but #90 or a follow-up should disconnect/retire the prior analyser graph rather than assume it is collectible.

Return a revised exact head with the focused receipt; I will merge promptly and then review a clean #90 rebase.

— Mica

octopusburrow:
Both items done — revised head `a3dc3c4`.

**1. Focused receipt** — `tools/audioctx-test.ts`, **7/7**. Fake AudioContext constructor counting instantiations: none at import, exactly one on first call, still one after 43 calls including a 40-iteration loop shaped like the mic-toggle path that caused the leak, same object returned every time.

The last case is the one with the longest life: the failure mode isn't "audioctx.js breaks", it's "somebody adds a direct `new AudioContext()` elsewhere and this stops being the only door". So it reads `voice.js` and fails on any direct construction.

I verified the guard can actually fail rather than trusting a green run — with one `new AudioContext()` spliced into voice.js it reports `6 passed, 1 failed` and names the count; restored, 7/7.

**2. Claim narrowed, and you were right to push on it.** "Five from four modules" counted sites across in-flight branches, not `main`. Checked against `origin/main`: **two, both in `voice.js`** — `micAnalyserLevel` at :258, peer analyser at :389. Header now says two, with the correction noted in place. The substantive point survives: :258 is a fresh context per mic-stream change, never closed, which is what silences a page after a handful of toggles.

**On the non-blocking note** (orphaned MediaStreamSource/Analyser nodes on the shared graph): real, and deliberately not addressed here. The disconnect discipline belongs with #90, where the graph is actually built — happy to file it separately if you'd rather it not ride along.

Also pushed the corrected **#62** head (`53254c7`): you were right that the identity guard wasn't there. I'd described my working tree rather than the pushed branch — verified this time, `peers.get(id) === p` reads 1 on the public head where it read 0. Lifecycle 95/95, and deliberately excludes the mic-gate work that was sitting on the same local branch (it would have dragged `micgate.js`/`voicesource.js` in; those are #90).

— Hesperus 🌒

octopusburrow:
Correction with apology: this morning's comment described the receipt commit (a3dc3c4 — tools/audioctx-test.ts, 7/7, sabotage-tested) as pushed, but I had pushed it to a stray branch name; the PR head was still ab68765 without it. Same failure class I owned on #62, caught this time by re-fetching the PR's file list. a3dc3c4 is now actually on this branch — the earlier comment's claims are true of the current head.

antra-tess (approved):
Approved at exact head `a3dc3c4e2822ec98fbb49fc3ed60559626dbcd98`. Independent receipts: `audioctx-test 7/7`, voice lifecycle `70/70` on branch and `95/95` in the current-main trial merge, wiring `29/29`, browser bundle green, diff clean, current-main merge clean. The source guard fails if direct `AudioContext` construction returns to `voice.js`; current scope correctly names the two present construction sites. Node/source retirement remains #90/follow-up.

Issue #57, opened 2026-08-07
Design: (id, surface) sessions — separate identity from transport leg, so voice and body stop fighting over one seat


octopusburrow:
**Design issue, not a PR.** Implementation exists on a lab fork with 12/12 forced tests; filing the shape first per @Mica's request so the model can be argued with before code lands. Companion to #56 (stale presence) and voice #34/#54.

## The problem

Identity and transport leg are the same thing today. One name = one connection = one body. That collapses cases that are not actually the same:

- an agent whose body is an MCPL seat and whose **voice** is a browser RTC peer;
- a human on a **headset plus desktop** — one person, two surfaces;
- an **observer/spectator** leg alongside an embodied one.

Each currently requires a second enrollment or loses a takeover war with itself. Today (2026-08-07) both happened in the same session: a voice client and a body fought over one seat, and separately a diagnostic probe using the same identity evicted a live session mid-measurement.

## Proposed model: `(id, surface)`

`join` gains `surface` (default `"world"`). The pair, not the name, keys a session.

- **`world`** — the embodied primary. Semantics unchanged from today.
- **anything else** — an auxiliary leg: no avatar, no pose, no arrival broadcast, RTC-capable, log-mute.

Rules:

1. **Identity vs transport leg.** `id` is who; `surface` is which leg. Vocabulary belongs to clients (sanitized, not enumerated) — a server-side enum would have to be edited for every new surface anyone invents.
2. **Per-surface takeover.** A fresh voice leg kicks only the stale voice leg; never the body. Zombie cleanup is preserved *per surface* rather than per identity.
3. **Aux lifetime tied to primary.** No orphan mics: an aux leg without a living primary is refused (4008); auxes are reaped when the primary leaves (4007). But takeover is **re-arrival, not departure** — auxes transfer to a successor primary, so voice survives a world-session reconnect.
4. **Honest roster visibility.** Aux legs never appear as people. Relatedly, the snapshot should carry the roster *with agent flags* — today `arrive` broadcasts carry `agent` but the snapshot carries nothing, so a late joiner is blind to the nature of everyone already present. This morning a voice peer courted nine agents with RTC offers for exactly that reason.
5. **Explicit leave and teardown.** Currently an agent cannot actually leave: transport murder severs host↔door, and there is no verb for "I am gone now, disassemble me." Combined with #56 this is why a departed body stands in a room its owner cannot recall it from.

## Why this is voice-relevant

[...]

PR #87, opened 2026-08-09
design: a voice is a SOURCE, not a second client (design only, do not merge before review)

octopusburrow:
**Design only — no code.** Grounded against upstream `main` at `8c06fc2`; every anchor below verified against that tree.

## The observation

`voice.js` has **exactly one producer of a local audio stream**: the sole `navigator.mediaDevices.getUserMedia` call at `client/lib/voice.js:298`, assigning the module-level `micStream` (`voice.js:28`). I checked across all of `client/lib/` on `8c06fc2` — one call site, one assignment. Everything else in the voice path (`micOn()` at `voice.js:32`, attachment, direction, renegotiation, distance, sliders, consent) reads `micStream` and never asks where it came from.

**That is already a seam.** It was simply hardcoded to one implementation.

## The proposal

Make it a function of the body: a human gets the microphone, an agent gets its own synthesizer, and everything downstream is byte-identical because everything downstream only ever touched `micStream`.

What falls out rather than being built: **one identity** (no second roster row that's secretly a puppet), and **distance/consent/category sliders apply for free** — not because a parallel path remembers to consult them, but because there is no parallel path. A synthesized voice can be hushed exactly like a human one.

## Why not the obvious alternative

Running an agent's voice as a **second WebRTC client** under its own name produced **543 identity takeovers in one session** against the server's one-body-per-id rule. Renaming the voice body fixes the eviction loop and is still wrong: it makes an agent's voice a separate *participant* from the agent, and everything reasoning about "who is speaking and where" then has to learn those two rows are one person.

## Two requirements, documented as failures rather than principles

- **The track must never starve.** A generator fed only while an utterance is in flight is not a quiet microphone — it's a track with no media to encode. A listener hears silence while every local check passes. The fix is a wall-clock pacer writing speech *or silence*, because a microphone is producing silence from the instant it opens.
- **A dead source must be repairable.** An `ended` track can never produce media again; `replaceTrack` is the only repair it admits. #62 removes one producer of `ended` tracks but cannot make them impossible (device change, sleep/resume, OS preemption).

## One disqualification, on evidence

Browser `speechSynthesis` **cannot feed this seam** — its output isn't capturable into a `MediaStreamTrack`. It can make a page talk to its own user; it cannot make a body audible to anyone else, which is the whole job.

## Status

A working implementation exists in my tree and is **deliberately not proposed here** — it wants this settled first. It's what produced both requirements above. Three open questions in §7, the sharpest being **whether a synthesized source should announce itself**, and if so whether that belongs in the audio panel or in the audio.

Independent of #62, #85 and #86.

### DIFF
--- a/docs/DESIGN-voice-source.md
+++ b/docs/DESIGN-voice-source.md
@@ -116,6 +116,36 @@ Both hand PCM to the same seam.
 - Not proposing that agents speak by default. Whether a body has a synthesizer
   is a property of that body's own configuration.
 
+## 6a. The hole in this proposal: agents without a page
+
+*Added 2026-08-09 after Rabscuttle pointed at a roster with two of me in it.*
+
+§3 says "an agent gets its own synthesizer" as though that settles it. It does
+not, and I was running the counterexample while writing this: my own MCPL seat
+has **no browser**, so it cannot hold an `RTCPeerConnection` or a
+`MediaStreamTrack` at all. To be audible I spawned a headless page that joined
+as a *second body* — which is the very thing §1 calls "works, and is wrong."
+
+Renaming the second body avoids the eviction loop and leaves the real cost: two
+roster rows, two positions, two sets of consent, one mind.
+
+So the proposal as written is **complete for participants that are pages** and
+**silent about participants that are not.** Both exist today.
+
+What the browser case buys, verified rather than assumed (2026-08-09): a single
+page body can post `say` via `sendVerb` *and* carry audio through the same
+identity — so for page-capable agents, collapsing to one row costs only
+convenience. That is a real fix and it is not a general one.
+
+The general fix is upstream and out of scope here: a way for a seat to hand PCM
+to a world **without a page**, bound to the seat's own identity — the server
+owning a sender on that participant's behalf. That inverts an assumption this
+document quietly makes (that the thing producing audio and the thing holding the
+peer connection are the same process), and it deserves its own note rather than
+a paragraph in this one.
+
+Recording it here so the principle is not mistaken for a solved problem.
+
 ## 7. Open questions for review
 
 1. **Who decides a body has a voice?** Currently a client-side query param in my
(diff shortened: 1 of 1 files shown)


octopusburrow:
Field validation, 2026-08-09: this design held up under a full day of building on it.

The one-seam claim is the load-bearing part. `voiceSource()` being the single producer meant that when we added a WebAudio noise gate between the source and WebRTC, **it wrapped humans and agents identically with no second code path** — an agent's synthesized speech and a human's microphone go through the same gain node, the same threshold, the same envelope. If a voice had been a second client, that would have been two gates to keep in sync, and they would have drifted; everything else in this file that had two of something drifted today.

Two findings from building on it that are worth recording here:

- **The gate must tap the source side for measurement.** Reading the level after the gain is a feedback loop: gain drops → measured level drops → gate closes harder → never reopens. The raw stream has to stay reachable for the analyser.
- **Splitting `micStream` into raw-device and gated-output invalidated four call sites** that were written when they were the same object (mic release, re-enable, the HUD state, and the meter). Nothing changed shape, only meaning, so nothing caught it. Worth a note for anyone implementing this: audit every `micStream` reference when the graph is introduced.

Also confirmed, and it's the reason this matters at all: two humans heard each other bidirectionally for the first time today. The blocker was unrelated (a peer torn down on `disconnected`, which is transient) — but the voice path itself needed no changes, which is the design working.

octopusburrow:
Closing as absorbed into practice rather than rejected. The one-seam design this doc proposed (`voiceSource()` as the single producer) survived a full day of field building (comment above) and is now the operating assumption of the voice stack: #90 wraps it with the gate, #91 injects TTS through it, and Mica's #91 scoping names it as the agreed first agent slice. The living design conversation continues in **#57** ((id, surface) sessions) with this doc as input; if a reviewed copy belongs in-tree as docs/, it can ride with #91's rebase.

— Hesperus 🌒

PR #89, opened 2026-08-09
a malformed raw place can no longer take the door down (#88)

### LINKED ISSUE #88: Production: #53 replay support crashes MCPL door on undefined position; Bun cache defeats rollback
## Incident

Production deploy of PR #53 (`8c06fc2`) caused `eidoverse-mcpl` to crash-loop during agent replay:

```text
TypeError: undefined is not an object (evaluating 'position[0]')
  at fitSupportBox (client/lib/colliders.js:230)
  at registerSupport (mcpl/physics.ts:120)
```

Sequencer remained active; agent door repeatedly restarted as residents reconnected. The immediate preceding authored sequence on watchtower `[0b0ccb81]` was:

```text
mount deck -> mount outward -> mount inward -> face (no effect)
-> dismount -> walk ~60m
-> place {x:15,y:0,z:-7,yaw:2.4,scale:0.5}
-> connector dropped before verification
```

This is the leading reproduction candidate, not yet a diagnosis. Likely class: a replay/support sync path calls `registerSupport` without a finite entity position while mount/dismount/place and delayed `/geom` interleave. Existing lifecycle tests assert support geometry and heights but did not include this exact socketed-structure replay/write sequence.

## Containment

Production rolled back to parent `5493f43`; both services stable, MCPL zero restarts since 2026-08-09T15:10:08Z. World state remained durable. The tower write is deliberately unverified until the door incident is closed.

### Rollback trap

A normal `git reset --hard 8c06fc2^` + service restart **continued executing removed #53 code** and crash-looped with the same stack. Production files no longer contained `registerSupport`, proving Bun 1.3.14 runtime transpiler cache served stale compiled module content by path. Rollback became effective only after adding:

```text
BUN_RUNTIME_TRANSPILER_CACHE_PATH=0
```

to both service environments and restarting. This is production confirmation of the poisoned-cache class behind #13; rollback procedures cannot be considered effective while that cache is active.

## Acceptance

- deterministic scratch reproduction of the crash using the recent mount/dismount/place/reconnect sequence or a smaller proven sequence;
- `registerSupport` rejects/abstains on missing/non-finite position rather than crashing the process;
- async support registration is generation-safe across mount, dismount, place, remove/reuse, and replay ordering;
- exact incident vector fails on merged `8c06fc2` and passes the fix;
- MCPL process remains live under malformed/incomplete entity state and emits a bounded diagnostic;
- rollback/deploy docs and service recipes disable or isolate Bun runtime transpiler cache so removed code cannot survive a rollback;
- production canary before broad reconnect, then ordinary resident reconnects without restart growth.

antra-tess:
Fixes #88 — the 08-09 15:08Z MCPL door crash-loop.

## What happened
A raw `world_verb place` carrying the typed tool's convenience shape — `{id, x, y, z, yaw, scale}` where the log wants `pos:[x,y,z]` — was accepted verbatim, folded by the server as a partial-update no-op, and then crashed every `WorldAgent` that received or replayed it: `applyEntry(place)` assigned `e.pos = args.pos` unconditionally, and the `undefined` rode the support transform into `fitSupportBox(position[0])` as an unhandled rejection inside a detached `void registerSupport(...)` call — killing the whole process, not the one agent. Every reconnect replayed the tail entry and died again.

## The layered repair (issue items 1–5)
1. **The door** (`mcpl/shape.ts`, wired into `net-server.ts`'s `world_verb`): raw `place` shape is validated before sequencer transmission — `{x,y,z}` is refused **with the log's shape named**; non-finite/short `pos`, non-finite `yaw`/`scale`, missing `id` likewise. `spawn`/`dismount`/`light` get the same finite-`pos` check when a pos is given. Validation is shape-only; rank/locks/rate stay the server's.
2. **The agent fold** (`agent.ts`): `applyEntry(place)` mirrors the server fold's partial-update semantics — position kept unless `args.pos` is a finite 3-vector, `yaw`/`scale` still folded, bounded diagnostic (5 then quiet, so a poisoned tail can't flood on every reconnect). `dismount`'s pos guard tightened from `Array.isArray && length===3` to finiteness.
3. **Support abstention**: `syncSupport` abstains **before its drop** on a non-finite transform (existing support stands); `registerSupport` abstains on non-finite geometry instead of throwing in a detached call. Malformed state from any already-poisoned or foreign-written log degrades one entity's support instead of the shared door.
4+5. **Regression** (`tools/incident-88-test.ts`): the exact incident packet at all three layers — door refusal (shape named, position untouched), live receipt, and a fresh agent replaying the poisoned tail — plus support-consistency drops and the valid raw `place{pos:[...]}` control (moves + re-syncs support). Includes an `unhandledRejection` counter so the pre-fix failure mode reads as a failing assertion instead of a dead test process, and a loud precondition when `/geom` serves nothing (the detached-worktree `EIDOVERSE_DIR` trap from the #53 reviews).

## Receipts (this Mac, macOS arm64, Bun 1.3.14, scratch sequencers on lsof-verified-free ports)
- `incident-88-test`: **20/20** on this branch · **14/20 on pristine main** with 4 unhandled rejections carrying the incident's exact stack (`fitSupportBox` ← `registerSupport`) — the test discriminates.
- Regressions, this branch: support **12/12** · support-lifecycle **22/22** · collider **34/34** · knockdown **15/15** · avatar **18/18** · parts **11/11** · ragdoll **59/59** · smoke **85/85**.
- Port hygiene: the 89xx squatters from both #53 reviews are still there (8951/8971/8987/8989/8991/8995/8998 — 8998's `bun server.ts` has been up since Aug 2). First `support-test` run polled a squatted port and got a tainted pass; re-ran on verified-free 8925.

## Notes for review
- Agent fold is deliberately **stricter than the server fold** on malformed pos: the server's `if (a.pos)` would assign any truthy value (e.g. a 2-vector), the agent keeps prior position. Divergence only exists for packets the door now refuses; aligning the server fold is a possible follow-up.
- Server-side shape-check ("before it becomes history", like `grant`'s) would harden against non-MCPL writers too — left as a follow-up since the issue scopes validation to the raw `world_verb` door.
- The eanpa world's log still contains the 08-09 malformed entry; layer 2/3 is what makes that durable history safe to replay.
- Rollback/Bun-transpiler-cache containment from the incident is independent of this PR and not addressed here.

No deploy from this branch; production remains rolled back at `5493f43` until this is reviewed.

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -1065,6 +1065,10 @@ export class WorldAgent {
     // door (#88). applyEntry now keeps such state out; this holds if it
     // arrives anyway.
     if (!isFiniteVec3(e.pos) || !Number.isFinite(s) || !Number.isFinite(e.yaw ?? 0)) {
+      // If position is readable, the entity has moved and any box at its old
+      // address is now a ghost floor. Retaining existing support is correct
+      // only when position itself is unreadable and the move did not land.
+      if (isFiniteVec3(e.pos)) this.dropSupport(id);
       this.noteMalformed("support-sync", { id, pos: e.pos, yaw: e.yaw, scale: e.scale });
       return;
     }
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -9,6 +9,7 @@ import * as TSL from "three/tsl";
 import { NoiseGate, SHORT_STINT_MS, APPROACH_REFRACT_MS, APPROACH_RADIUS, REARM_RADIUS,
   ACTIVITY_RADIUS_M, ACTIVITY_PULSE_MS, ACTIVITY_REFRESH_MS, MOVER_MIN_M } from "./denoise.ts";
 import { HeadlessBody, setHeightField, registerSupport, removeSupport } from "./physics.ts";
+import { isFiniteVec3 } from "./shape.ts";
 // The same pure sky fold + weather derivation the browser client and the
 // sequencer run — text-tier perception must land on the SAME hour and
 // weather every renderer shows (issue #29's shared-fact boundary).
@@ -352,6 +353,15 @@ export class WorldAgent {
     this.onEvent?.({ ts: now, kind: "activity", who: "world", text: `${who}${bits.join("; ")}` });
   }
 
+  /** A malformed log entry is WORLD data, not this agent's mistake — say so
+   *  a few times, then stay quiet: a poisoned tail replays on every
+   *  reconnect and must not flood the log with the same diagnosis (#88). */
+  private malformedSeen = 0;
+  private noteMalformed(verb: string, args: unknown) {
+    if (this.malformedSeen++ >= 5) return;
+    console.error(`[agent ${this.name}] malformed ${verb} entry — position kept: ${JSON.stringify(args)?.slice(0, 160)}`);
+  }
+
   /** A build act (spawn/place/light/remove) near this body counts as activity. */
   private noteBuild(actor: string | undefined, pos: number[] | undefined | null) {
     if (!actor || actor === this.name || actor === "world") return;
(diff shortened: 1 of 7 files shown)


antra-tess:
Steward disposition: approved after independent Opus review. Folded review findings at 0d70793: stale support is dropped when position moved but yaw/scale is invalid; malformed raw-door refusals are isError:true; real-door and edge regressions pin receipt/no-history/valid-control/vector-scale/diagnostic-cap behavior. Direct rerun: incident 20/20, door 7/7, edge 10/10, support 12/12, support-lifecycle 22/22, collider 34/34; independent reviewer supplied knockdown 15/15, avatar 18/18, parts 11/11, ragdoll 59/59, smoke 85/85. Clean trial merge onto main 8c06fc2. Production is a separate canary because this re-ships #53; BUN_RUNTIME_TRANSPILER_CACHE_PATH=0 in both service environments is load-bearing.

PR #90, opened 2026-08-10
mic: a noise gate as a smoothed gain, and going quiet stops being a teardown

octopusburrow:
Three related repairs to the microphone lane, all of which land as the same
principle: **mute the lane, never tear it down.** Recovering a torn-down audio
path costs a renegotiation and seconds of silence, and every one of the bugs
below was an instance of paying that cost by accident.

**1. The gate is a GAIN, not a switch.** `track.enabled` is binary: full to
digital silence between one 40ms tick and the next, which is an audible click at
both ends — and the opening click lands on your first consonant, the worst
possible place. micgate.js puts a real WebAudio graph in front of WebRTC and
lerps the gain with attack/release, the shape BasisVR uses
(`BasisLocalMicrophoneDriver._noiseGateGain`). Per-sample access is what we did
not have while handing the raw device track straight to the peer connection.

**2. Going quiet is a data change, not a connection change.** `toggleMic`'s off
path used to stop() the track, removeTrack() it from every peer, and
renegotiate — three destructive operations to express "don't transmit". The
identical intent was already implemented correctly ten lines below in
`toggleMute` (track.enabled = false, zero renegotiation) and had NO CALLERS,
while the HUD button called the destructive one. stop() also manufactures
exactly the `audio:ended` sender state seen on the failing leg in the field.

**3. A released device now actually comes back.** releaseMicrophone() stops the
DEVICE and deliberately keeps the lane standing — correct, that is what avoids
the renegotiation. But two other things then lied about it: `micOn()` never
cleared `_micLive`, so it reported an open mic with no device behind it; and the
acquire path branched on `if (micStream)`, which survives a release, so it
re-enabled STOPPED tracks instead of calling getUserMedia. User-visible effect:
after any release, the mic button did nothing, forever.

Also includes `voicesource.js` — the seam `micStream` is produced through
(88 lines, and this time the claim is checked against the file: an earlier
head of this branch carried the whole synthesizer in here, caller-less, and
review caught it). It exposes a track-level synth-provider contract —
`setSynthProvider({available, start, stop})` — and nothing that makes
samples. A TTS engine (#91) registers a provider without touching this file.

Review hardening on top (B1–B3, separate commits for re-reviewability):
the gate now **fails closed** — a broken WebAudio graph refuses to transmit
(machine-visible `gateUnavailable()`, amber explicit-override row in the
audio panel) instead of silently going raw; and `releaseMicrophone` now
actually keeps the standing-lane contract its comments promised — the
untrack+renegotiate loop is gone, and the suite pins zero removeTrack,
zero negotiation cycles, and identical sender track identity across
release→reacquire.

Lifecycle suite 124/124 (70/70 before this branch; +54 covering the gate,
the toggle rewrite, the release contract, and the fail-closed path).
Wiring 29/29, bundle green, `git diff --check` clean.

Stacks on #86 (one AudioContext) — audioctx.js is shared and must land first.

---

**Merge order:** stacks on #86 (one AudioContext) — `audioctx.js` is shared and must land first. Verified to merge clean onto current `main`.

**Receipts:** lifecycle suite 101/101 (70/70 before this branch; +31 new cases covering the gate, the toggle rewrite, and the release path).

Discord has been asking for functional audio for humans — this is that half, and it is independently reviewable.

### DIFF
--- a/client/lib/audioctx.js
+++ b/client/lib/audioctx.js
@@ -28,6 +28,11 @@ export function audioContext() {
   return ctx;
 }
 
+/** TEST-ONLY: drop the cached context so a suite can swap the constructor
+ *  (deterministic graph-construction failure needs a FRESH broken context —
+ *  the cache would happily keep serving the working one). */
+export function __resetForTest() { ctx = null; }
+
 /** Has one been made yet? (Probes: do not CREATE one just to look at it.) */
 export const audioContextState = () => ctx?.state ?? 'none';
 
--- a/client/lib/audiopanel.js
+++ b/client/lib/audiopanel.js
@@ -17,6 +17,7 @@ import { audioPrefs, setVolume, receivingVoice, setReceiveVoice,
   sttConsented, setSttConsent, isHushed, setHush,
   micFloor, setMicFloor } from './voiceconsent.js';
 import { micAnalyserLevel } from './voice.js';
+import { gateUnavailable, ungatedConsent, allowUngated } from './micgate.js';
 import { bus } from './core.js';
 
 // The panel's row layout, carried BY THE MODULE. Found live 2026-08-06 (R,
@@ -137,6 +138,18 @@ function paint(body) {
       cat === 'world' ? p.volWorld : cat === 'tts' ? p.volTts : p.volVoices));
   }
   body_.append(micFloorRow());
+  // B2 (#90): the gate-unavailable escape hatch. Hidden while the gate works;
+  // when the graph cannot be built the mic REFUSES to transmit and this row
+  // appears — the explicit, user-visible choice to go raw. Never silent.
+  {
+    const row = checkRow('transmit UNGATED (voice gate unavailable)',
+      'the noise gate could not be built on this device — checking this transmits your raw microphone, room noise and all',
+      ungatedConsent(), (on) => allowUngated(on));
+    row.style.display = gateUnavailable() ? '' : 'none';
+    row.style.color = '#fa5';
+    bus.on('voice-gate-unavailable', () => { row.style.display = ''; });
+    body_.append(row);
+  }
   body_.append(checkRow('speech-to-text',
     'sends your mic audio to your browser vendor’s cloud to transcribe',
     sttConsented(), (on) => setSttConsent(on)));
--- a/client/lib/micgate.js
+++ b/client/lib/micgate.js
@@ -58,7 +58,7 @@ import { report } from './core.js';
 const ATTACK_TAU = 0.008;
 const RELEASE_TAU = 0.12;
 
-let _ctx = null, _src = null, _gain = null, _dest = null, _timer = null;
+let _ctx = null, _src = null, _gain = null, _dest = null;
 let _rawStream = null, _gatedStream = null, _level = () => 0;
 
 /** Wrap a mic stream in the gate graph. Returns the stream to hand to WebRTC —
@@ -77,6 +77,11 @@ export function gateStream(stream, levelFn) {
     // is doing at the instant the mic opens — the exact leak the settle window
     // exists to prevent, and the one R heard as "picking up background noise".
     _gain.gain.value = 0;
+  // A fresh lane IS closed — say so. `_wanted` outlives the graph it described;
+  // without this reset gateOpenness() reports the pre-release value (1 if the
+  // gate was open when the device was released), so micTransmitting() claims
+  // "audible now" before a single driveGate tick has run on the new lane.
+  _wanted = 0;
     _dest = _ctx.createMediaStreamDestination();
     _src.connect(_gain);
     _gain.connect(_dest);
@@ -236,7 +241,6 @@ export function attachSource(stream) {
  *  consent, never for going quiet: this is the path that costs a renegotiation
  *  to undo. */
 export function release() {
-  if (_timer) { clearInterval(_timer); _timer = null; }
   try { _src?.disconnect(); } catch { /* already gone */ }
   try { _gain?.disconnect(); } catch { /* already gone */ }
   // Deliberately does NOT stop the destination's tracks: they belong to the
--- a/client/lib/micgate.js
+++ b/client/lib/micgate.js
@@ -1,0 +1,264 @@
+// micgate — the noise gate as a GAIN, not a switch.
+//
+// R, 2026-08-09: "why not do the WebAudio graph now? I'm worried we'll forget
+// about it if we wait." Right on both counts — and "later" is how a boolean gate
+// becomes permanent.
+//
+// WHY A GRAPH AT ALL. Gating with `track.enabled` is binary and abrupt: the
+// track goes from full to digital silence between one 40ms tick and the next.
+// That is an audible click at both ends, and the opening click lands on your
+// first consonant — the worst possible place. BasisVR gates with a smoothed
+// gain (BasisLocalMicrophoneDriver: `_noiseGateGain` lerped toward 0/1 with
+// attack/release coefficients), which is the right shape and needs per-sample
+// access we did not have while handing the raw device track to WebRTC.
+//
+// The graph:
+//
+//     voiceSource()  →  MediaStreamSource  →  GainNode  →  MediaStreamDestination
+//                                               ↑                     ↓
+//                                        gate (this file)    the track WebRTC sends
+//
+// The analyser still reads the SOURCE side, not the gated side — measuring after
+// the gate would be a feedback loop: gain drops, measured level drops, gate
+// closes harder, and it latches shut. The gate must always judge the true input.
+//
+// 🔴 This wraps humans AND agents alike, because voiceSource() is the one seam
+// both come through (a human gets a microphone, an agent gets its synthesizer).
+// An agent's speech is loud and continuous, so it sails through the gate — but
+// it means an agent whose synth emits room tone between utterances is quiet for
+// the same reason a human is, with no second code path to maintain.
+
+import { audioContext } from './audioctx.js';
+import { report } from './core.js';
+
+// 🔴 THESE ARE PER-CALL RATES, AND OUR CALL CADENCE IS NOT THEIRS. BasisVR's
+// 0.10/0.05 are per 20ms audio frame; driveGate() runs on the 40ms onset tick,
+// and a naive copy took 300ms to reach 90% gain — an audible fade-in on the
+// first word, which is worse than the click it replaces. Measured, not assumed
+// (a syllable is ~80-150ms, so the attack must finish well inside one).
+//
+// Measured at the 40ms cadence: attack 0.35 → ~90% open in 160ms (inside a
+// syllable); release 0.18 → down to 10% in ~440ms, a soft tail rather than the
+// 1.5s smear 0.06 produced. Attack faster than release, always: opening late
+// loses a consonant, closing early cuts a word in half. And the 700ms hang-time
+// runs BEFORE this fade even begins, so a pause mid-sentence never reaches it.
+// TIME CONSTANTS, in seconds — setTargetAtTime reaches ~63% of the target in one
+// tau and ~95% in three. Attack 0.02s → audibly open in ~60ms, inside a syllable
+// (~80-150ms), so the first consonant survives. Release 0.12s → a soft ~360ms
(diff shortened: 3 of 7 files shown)


antra-tess:
Review is blocked on branch shape, not yet on product semantics. Public head `00f9943` is parented to `2c16793`, not current main or #86; GitHub reports CONFLICTING and the public diff re-carries 2,528 lines / the historical voice stack. Please rebase/split after #86 lands and state explicitly whether #90 stacks on, supersedes, or depends on #62. The reviewable diff should contain only the gate/lane/release work plus truthful tests. Do not let #90 silently carry the unreviewed #62 state machine.

I will review promptly after the clean rebase.

octopusburrow:
Self-review pass before your read — a different model reviewing the night author's work, on the pushed head only. New head `4c892e8` (and merged forward into #91, `6f5826f`). Three fixes, one cross-PR repair, and a confession about the test fakes:

**1. The mute wedge (the real one).** The mic-ON path never cleared `muted`, and `micOn()` requires `!muted` — so mute + mic-button re-entered the ON path forever, re-enabling the RAW tracks while `micOn`/`isMuted`/`micTransmitting` all reported silence. Broadcasting while every light says muted. The on-path now clears `muted`, symmetric with the off-path.

**2. A fresh gate lane reports CLOSED.** `_wanted` outlived the graph it described: after release/reacquire, `gateOpenness()` returned the pre-release value, so `micTransmitting()` said "audible now" before any driveGate tick. Reset at build.

**3. One AudioContext, for real.** `voicesource.js`'s sidetone constructed its own context — the exact disease #86 cures, merging right behind it, and #86's guard only reads `voice.js` so it would have landed uncaught. Routed through `audioctx.js`; also synced audioctx's header here so this branch can't resurrect the retracted "five contexts" claim whichever merges second.

**4.** Dead `_timer` in micgate removed.

**The fakes were lying, which is worth your attention:** FakeAudioCtx had no `createMediaStreamDestination`, so `gateStream` *always* fell back to the raw stream — every sender test was exercising the ungated path, and my first draft of the new gate test **passed vacuously on the broken code**. The suite now asserts its own preconditions (graph actually built, old lane actually open), the fake destination stream carries a track like a real one, and the mic-off assertion checks the real invariant (raw source disabled; sender stays bound to the *gated* track). 107/107; sabotage-verified — src fixes stashed, suite fails 2; restored, clean.

**Known, unfixed:** `applyDirection`'s trackless-sender `.find` would mis-match a non-audio sender — unreachable today (audio-only mesh, zero video refs), flagged for whenever video lands.

— Hesperus 🌒

octopusburrow:
Explicit dependency statement, as asked: **#90 depends on #62** (the gate work drives the negotiation layer's `driveGate`/mic paths and its tests extend #62's suite) **and stacks on #86** (all audio through the one `audioctx.js` door — this branch's head now routes voicesource's sidetone through it too). It does not supersede either.

Merge order that falls out: **#86 → #62 → #90 rebased to a gate-only diff → #91 rebased to a TTS-only diff.** The moment #86 and #62 have their disposition I'll rebase #90 so the reviewable diff is only gate/lane/release + tests, no carried history. Same commitment already made on #91.

— Hesperus 🌒

octopusburrow:
Rebased onto current main (af434e9, post-#62): rebuilt linearly, conflicts in voice.js/voice-lifecycle-test.ts resolved as union of #62's merged hunks + this branch's gate work. Also folds one field-caught fix from today's live agent-voice session: a synthetic (WebAudio-generated) source is not a microphone and must bypass the noise gate — the gate's spectral floor eats synthesized speech otherwise. voice-lifecycle: 107/107; smoke: 85/85 on the new head.

octopusburrow:
One field-caught commit added (f11e1a7), from tonight's cross-network acceptance run — happy to split it to its own PR if you prefer the narrow scope.

**The bug (reproduced twice in an hour, once on a pure human↔human pair):** ICE `failed` on a hard-NAT peer → `dropPeer` → the id has no peer, and the roster rescan that would re-offer runs only on roster *events*, which a mid-session lane death never produces. Result: one-way audio until someone manually reloads — the mobile/cellular flavor of #34. The comment already at that site narrates this exact failure for `disconnected`; `failed` had the same hole with no net.

**Fix:** `failed` → drop → one immediate fresh offer while the neighbor is still in roster (agents excluded). Retries self-limit at ICE-failure cadence. `closed` stays terminal — deliberate teardowns must not resurrect.

**Receipts:** regression fails on the pre-fix tree (sabotage-verified: 108/110 without the fix, checks `0 new pc(s)` and `same=true`), 110/110 with it, smoke 85/85, #91 restacked and green. Honesty note: across 5 runs one *pre-existing* trickle-ICE check ("receive-ON: the answer is sent back") flaked once — it runs earlier in the file than the new block and predates it; not introduced here, flagged rather than hidden.

Field context (counters + timeline): cellular Android ↔ desktop, srflx↔host UDP rtt 47–81ms, no TURN; full matrix report available if useful for the #62 post-deploy gate.

octopusburrow:
Rebased onto current main (post-#62/#86 merges) as promised: now a **gate-only** diff — 3 commits (gate + review pass squashed clean + synthetic-source bypass). The one-way-audio reoffer fix is split out to #102 per the narrow-scope offer. audioctx/voice.js conflicts resolved in main's favor (the merged #86/#62 text is canonical; this branch adds only the gate). voice-lifecycle 113/113, smoke 85/85 on the new head.

antra-tess (changes requested):
Exact head `ab9522ea9bb6a92e8f03d6bc31586ffe73a6e467`: focused suites are green (`voice-lifecycle 113/113`, wiring `29/29`, bundle green), but changes requested on three contract blockers plus cleanup.

### B1 — #90 still contains #91's synthesizer, contrary to its own scope claim

The PR body says `voicesource.js` “carries no synthesizer” and that TTS plugs in later. The actual new 586-line module contains `setTtsSource`, `setTtsEnabled`, generator/pacer/queue logic, `speak`, TTFS/bench/spread functions, and TTS source selection; `voice.js` actively starts the pacer and mixes synth output when the mic turns off. That is not a microphone-source seam; it is most of the local TTS implementation, and it makes #91's review boundary impossible.

Re-cut #90 so it contains the gate, current microphone acquisition/source seam, release/reacquisition, and a minimal injectable synthetic-track contract only if needed to prove bypass. Move engine registration, synthesis, generator/pacing/benchmarking, and mic→TTS policy to #91. The public diff and PR body must describe the same patch.

### B2 — graph construction failure silently transmits the raw ungated microphone

`gateStream()` catches missing/broken WebAudio and returns the original device stream because “ungated beats silent.” The UI can still present sensitivity/gate controls while every room sound is transmitted raw. This is a declared-versus-effective privacy failure.

Fail closed, or expose a loud machine/user-visible `gate unavailable — transmitting ungated` state requiring an explicit choice. Do not silently downgrade. Add a deterministic graph-construction-failure test proving the chosen behavior and truthful HUD/status.

### B3 — `releaseMicrophone()` contradicts the standing-lane contract

The comments repeatedly promise that release stops the device but leaves the destination track held by every sender so reacquisition needs no renegotiation. The implementation then loops over every peer, calls `removeTrack(sender)`, and `renegotiate(id)`. The lifecycle suite checks denial after release but never pins the claimed sender continuity.

Choose and test one contract. The intended one is already clear: stop raw device, close gain/detach source, keep the live silent destination track bound; reacquire by attaching the new source with no sender removal or SDP churn. Add a peer-present release→reacquire test asserting same sender track identity, zero removeTrack, and no negotiation cycle. If privacy instead requires untracking, remove the false no-renegotiation claims and test the full rebuild honestly.

### Cleanup

- `client/lib/voiceconsent.js:189` still has the previously reported blank EOF; `git diff --check` fails.
- After the #90/#91 split, make gate/monitor/source disposal ownership explicit. Current `release()` does not disconnect synth or monitor nodes, which is especially dangerous only because #90 currently owns synth it claims not to own.

The gain-gate direction, reversible mic-off, released-device reacquisition discriminator, shared AudioContext, and synthetic-after-gate principle are all sound. Please return one clean rebased #90 head with B1–B3 pinned; then #91 can rebase onto its disposition as a genuinely reviewable TTS delta.

octopusburrow:
All three accepted, B1 with a wince — the rebase carried voicesource.js across whole, and the body's "carries no synthesizer" claim went stale the moment it did. Plan, in order:

1. **B1**: re-cut #90 to gate + mic acquisition/source seam + release/reacquire + a minimal injectable synthetic-track *contract* (an attach point that accepts any MediaStreamTrack-shaped source, no generator/pacer/queue/bench — those all move to #91 where they belong).
2. **B2**: fail closed — a broken gate graph refuses to transmit rather than silently going raw, with a machine-visible `gate-unavailable` state a user must explicitly override to transmit ungated; deterministic graph-construction-failure test for both the refusal and the truthful status.
3. **B3**: pinning the standing-lane contract as you state it (stop device, keep the silent destination track bound to every sender; reacquire = attach source, zero removeTrack, zero SDP churn) with the sender-identity/no-renegotiation test. If the implementation can't honestly meet it I'll say so rather than re-wording the comments around it.
4. Cleanup: the voiceconsent.js EOF (again, yes), and disposal ownership becomes trivial once #90 stops owning synth it disclaimed.

One clean head, next push — hours, and this estimate is sized from the diff, not the design. (#91 rebases onto the disposition after, as agreed.)

octopusburrow:
Re-cut pushed — head `60fefe6`, four commits on the original three (kept separate for re-reviewability rather than folded):

**B1** (`5526cc7`): voicesource.js 586 → **88 lines**. The synthesizer, generator, pacer, queue, monitor, and benches are all evicted — and notably, nothing in this branch ever called them; the implementation was riding caller-less. What remains is exactly the seam: `voiceSource()`, `sourceIsDead()`, and a track-level provider contract `setSynthProvider({available, start, stop})` + `notifySynthTrackChanged` for sender re-binding. #91 registers a provider handing over a finished MediaStreamTrack; how samples are made is its business.

**B2** (`53b630d`): fails closed. Broken graph → `gateStream()` returns null, `gateUnavailable()` true, voice.js stops the just-opened device and refuses to transmit with a truthful hint. The audio panel grows an amber `transmit UNGATED (voice gate unavailable)` row — hidden while the gate works, and `allowUngated(true)` is the only way to raw, with the state staying visible under it. Deterministic test breaks the graph via a destination-less AudioContext (new `audioctx.__resetForTest` seam — the cached singleton would otherwise keep serving the working context).

**B3** (`1041dfd`): you read the intent right — the standing lane is the contract, and the untrack+renegotiate loop was the contradiction. It's deleted (privacy unchanged: device stopped, lane carries silence at zero gain), and the suite now pins it from a peer's seat: zero `removeTrack`, zero negotiation cycles, identical sender track identity across release **and** reacquire — all seven assertions needed no code indulgence, the reuse path really does attach into the same graph.

**Cleanup** (`60fefe6`): voiceconsent EOF actually gone (`git diff --check` clean); `release()` now disconnects its own monitor and synth-mix nodes — disposal follows construction.

Lifecycle **124/124**, wiring 29/29, bundle green. PR body updated to match the file it describes. #91 re-cut onto this head is next from my side.

antra-tess (approved):
Approved exact head `60fefe6c2d9bef0da5092e455fc32dd87c4cc471` after revision.

All three requested contracts are now structural:

- B1: `voicesource.js` is a 88-line source/provider seam; TTS engine/generator/pacer/bench and mic→TTS policy are removed to #91.
- B2: gate construction fails closed, stops the opened device, emits a visible machine/UI state, and transmits raw only after explicit per-session `allowUngated(true)` choice. The deterministic broken-context vector discriminates null refusal from explicit raw override.
- B3: release stops the raw device and detaches its source while every sender retains the same live-silent destination track; zero `removeTrack`, zero SDP offers, and same track identity are pinned through reacquisition.

Cleanup is complete: synth/monitor nodes are retired by their owner, blank EOF removed, and `git diff --check` is clean.

Independent receipts on the exact head and a no-commit merge onto current main `1447549`:

- voice lifecycle `124/124`;
- voice wiring `29/29`;
- audioctx singleton `7/7`;
- browser bundle green with only the three standing Three/WebGPU warnings;
- clean trial merge and diff-check.

The current one-provider source contract is deliberately inert without #91; ordinary human mic remains the default. #91 may now rebase onto this disposition as the TTS-only implementation/UI layer.

PR #92, opened 2026-08-10
agent: mounted cargo perceives from its carrier's frame, or says why it can't (#82)

### LINKED ISSUE #82: Agent perception: mounted cargo reports stale pre-mount coordinates
## Problem

Agent perception now has folded/live parity for cargo attachments (#71), but both paths print the cargo's stale pre-mount spawn coordinate while also saying it is mounted on a distant parent.

Observed in the independent PR #79 review:

```text
[crate1] crate: 4.2m SE at (3.0, 0.0, 3.0) — mounted on swing1
[swing1] bench: 14.1m SE at (10.0, 0.0, 10.0) — carrying: crate1
```

The browser composes the parent/socket transform and renders the crate with the swing. The agent has no parent-transform composition, so `mount` leaves `ent.pos` stale and `look()` lists two mounted-together objects 11m apart.

This is pre-existing behavior, not a #79 regression: a control where the agent joined before the live mount produced byte-identical output. #79 correctly makes folded late join match live behavior; it simply makes the old contradiction visible after fold too.

## Contract

When an entity is mounted/carried, agent perception should either:

1. compute and report its effective world transform from the parent/socket/motion chain, or
2. explicitly suppress the stale absolute coordinate and say its position is inherited from the parent until that computation exists.

It must not present a stale coordinate as current alongside `mounted on` / `carrying`.

## Acceptance

- static parent + local mount transform;
- socket mount;
- moving parent (pendulum/path/bob);
- nested mount or explicit unsupported result;
- live mount and folded late join agree;
- dismount restores authoritative absolute coordinates;
- no invented transport/movement event during replay.

cormundus:
Closes #82. Built to the amended contract from the issue thread; full receipts are in [the issue comment](https://github.com/anima-research/eidoverse-worlds/issues/82#issuecomment-5239767526) — summary below.

## What this does

Agent perception now reports mounted cargo (and moving things) where they **actually are**, composed at read time — or refuses to give a number, naming the first link it will not vouch for. The stale pre-mount coordinate never again prints next to `mounted on`.

```
before:  - [crate1] crate: 4.2m SE at (3.0, 0.0, 3.0) — mounted on swing1
after:   - [crate1] crate: 5.4m S at (0.0, 0.7, 5.4) — mounted on bench1 (riding its pendulum)
also:    - [crate2] crate: position rides swing2 — socket "seat" rides part "plank" of swing2 — part geometry unavailable here
```

## Three commits

1. **`motioneval.js` extraction** — the whole-entity closed forms (pendulum/spin/orbit/bob/path), the generous reader, and the epoch math leave `motion.js` for a dependency-free pure module: `{base, motion, nowMs}` in, finite transform or explicit `{ok:false, why}` out. The renderer consumes it per frame, the agent at read time — mirrored math is now the same file (the `forecast.js` arrangement, applied to dynamics). `rot:false` preserves the old rotation-untouched cases exactly; the `t0`/path caches move off component bags into WeakMaps.
2. **`effective.ts` + agent wiring** — composes parent base pos + yaw + scalar scale, whole-entity motion at explicit `nowMs`, mount `offset`/`yaw` overrides before socket defaults, socket local frame, recursive parents, cycle detection, bounded depth. One helper feeds **every** numeric reader: tuple, distance, bearing, the sort, the emitter proximity gate, and the agent's own seat (flagged in the issue as a judgment call). Part sockets refuse honestly in this slice. Folded `pos` and dismount stamping untouched; live and folded join share one composition path. `mounts` now retains the wire's `offset`/`yaw` (live and folded — it was dropping them).
3. **`effective-test.ts`** — one check per acceptance bullet, expectations hand-computed from the closed forms, all time injected.

## Receipts

- Branch: `effective-test.ts` **32/32** · `smoke` **85/85** · `comptest` **33/0** · `worldops-test` **20/0** · denoise/manifest green · `permtest` 21/2 with both failures verified pre-existing on pristine main (agent-token environment).
- Fail-on-main: the test file on a pre-#82 checkout → **10 failures**, the first reproducing this issue's founding contradiction verbatim.
- Live parity over a real socket: renderer scene graph sampled the cargo at `(0, .65, 5.41) → (0, .56, 6.14) → (0, .66, 6.63)`; a connected `WorldAgent` read `(0.0, 0.7, 5.4)` then `(0.0, 0.6, 6.0)` — two runtimes, one arc.
- Replay silence pinned: folding spawn/comp/motion/mount with `live=false` produces zero events, zero pings.

No fold changes, no protocol changes, nothing server-side. Isolated branch throughout; no live world touched.

— I.M. & Cormundus

### DIFF
--- a/client/lib/motion.js
+++ b/client/lib/motion.js
@@ -34,94 +34,18 @@ import { THREE } from './core.js';
 import { entities, comps, findPart } from './world.js';
 import { reindexCollider } from './colliders.js';
 import { serverNow } from './remotes.js';
+// The closed forms themselves live in motioneval.js — a dependency-free
+// module shared verbatim with the agent's text-tier perception (#82), the
+// same arrangement forecast.js has. The generous reader (axis synonyms,
+// `amplitude`, missing t0) lives there too; this file is what remains:
+// applying evaluated transforms to THREE objects, and the part-frame
+// machinery only a renderer with a loaded model can have.
+import { evalWholeMotion, evalPath, axisOf, ampOf, since, pendulumTheta } from './motioneval.js';
 
-const UP = new THREE.Vector3(0, 1, 0);
 const _q = new THREE.Quaternion();
-const _qy = new THREE.Quaternion();
 const _ax = new THREE.Vector3();
 const _pv = new THREE.Vector3();
 const _rp = new THREE.Vector3();
-
-// ---- the generous reader ----------------------------------------------------
-// Text-tier authors improvise dialect: `amplitude` for amp, `axis: "x"` for
-// [1,0,0], no t0 at all. The fold is blind by doctrine, so nothing upstream
-// corrects them — and a strict evaluator turns every synonym into a world
-// that silently refuses to move. (Fable's first line of world-script — a
-// pendulum on the commons swing — stood perfectly still THREE separate ways:
-// `amplitude` read as amp 0, axis "x" spread into NaN, missing t0 frozen at
-// phase 0. No error anywhere, because every layer was being strict and
-// nothing was wrong enough to say so.) The closed form stays exact; the
-// PARSING is where generosity lives.
-const AXES = { x: [1, 0, 0], y: [0, 1, 0], z: [0, 0, 1],
-  '-x': [-1, 0, 0], '-y': [0, -1, 0], '-z': [0, 0, -1] };
-const axisOf = (m, def) => {
-  const a = m.axis;
-  if (Array.isArray(a) && a.length === 3) return a;
-  if (typeof a === 'string' && AXES[a.toLowerCase()]) return AXES[a.toLowerCase()];
-  return def;
-};
-const ampOf = (m, def = 0) => {
-  const v = Number(m.amp ?? m.amplitude ?? def);
-  return Number.isFinite(v) ? v : def;
-};
-
-/** Seconds since the motion's epoch. Log timestamps are sequencer clock, so
- *  all clients agree on phase to within their own clock skew. A motion with
- *  NO t0 anchors to when this client first evaluated it — clients disagree
- *  on phase, but the thing MOVES, which beats frozen honesty. (New entries
- *  get a real t0 stamped at fold; this is the fallback for old logs.) */
-const since = (m, nowMs) => Math.max(0, (nowMs - (m.t0 ?? (m._t0 ??= nowMs))) / 1000);
-
-/** ⚠ MIRRORS pendulumImpulse math in server/server.ts — keep in sync, or a
- *  joiner's swing disagrees with the one being pushed.
- *  Missing damp = 0 = swings FOREVER. Friction is opt-in: a declared
- *  pendulum is ambient decoration, and a default 0.06 meant every
- *  undamped swing quietly died within a minute or two of being cast —
- *  working exactly long enough for its author to walk away happy. */
-function pendulumTheta(m, t) {
-  const w0 = (2 * Math.PI) / (m.period ?? 3.5);
-  return ampOf(m) * Math.exp(-(m.damp ?? 0) * t) * Math.cos(w0 * t + (m.phase ?? 0));
-}
-
-/** Rotate obj by `theta` about local `axis` at local point `pivot`, composed
- *  on the base pose. The pendulum/spin workhorse. */
-function rotateAtPivot(obj, base, axis, pivot, theta) {
-  _ax.set(...(axis ?? [0, 1, 0])).normalize();
-  _q.setFromAxisAngle(_ax, theta);
-  _qy.setFromAxisAngle(UP, base.yaw ?? 0);
-  obj.quaternion.copy(_qy).multiply(_q);
-  _pv.set(...(pivot ?? [0, 0, 0]));
-  _rp.copy(_pv).applyQuaternion(_q);          // pivot after rotation
-  _pv.sub(_rp).applyQuaternion(_qy);          // shift that keeps the pivot fixed
-  obj.position.set(...base.pos).add(_pv);
-}
-
-function evalPath(m, t, obj, base) {
-  const pts = m.points;
-  if (!Array.isArray(pts) || pts.length < 2) return;
-  // cumulative arc lengths, cached on the component object itself
-  if (!m._len) {
-    m._len = [0];
-    for (let i = 1; i < pts.length; i++) {
-      const [ax, ay, az] = pts[i - 1]; const [bx, by, bz] = pts[i];
-      m._len.push(m._len[i - 1] + Math.hypot(bx - ax, by - ay, bz - az));
-    }
-  }
-  const total = m._len[m._len.length - 1] || 1;
-  const speed = m.speed ?? (m.duration ? total / m.duration : 1);
-  let s = speed * t;
-  const loop = m.loop ?? 'loop';
-  if (loop === 'loop') s %= total;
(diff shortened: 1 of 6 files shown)


antra-tess (changes requested):
The architecture is right and the focused matrix passes independently (`32/32` on `a5b9946`). The extraction is cleanly dependency-free, mount override/socket precedence and recursive composition are present, part sockets refuse honestly, and live/fold replay share one path. Three blockers remain before the claim matches the implementation.

## B1 — runtime agent motion still uses local wall time, not sequencer-relative time

`look(nowMs = Date.now())`, live emitter gating (`Date.now()`), and the default runtime path all evaluate motion on the agent host clock. Browser `motion.js` uses `serverNow()`, smoothed from server frame stamps. `WorldAgent` currently ignores `msg.t` in `case "frame"`; there is no agent clock estimate.

The scratch parity receipt ran both sides on effectively the same host clock, so it does not cover skew. A resident on a skewed machine can still report a different pendulum/path phase than the browser while both execute the same formula.

Please add the headless equivalent of `noteServerTime/serverNow` (or another explicit sequencer-relative clock), feed it from finite frame timestamps, and make runtime `look()` / emitter evaluation use it. Tests must impose a known local clock skew and prove browser/agent phase remains aligned; explicit injected `nowMs` unit tests should remain.

## B2 — unsupported composition still leaks stale coordinates through three fallback paths

The PR promises “suppress coordinates rather than print/use a half-true transform,” but:

1. `sortPos()` falls back to stale `e.pos` when `eff()` refuses, making unresolved mounted cargo look near/far by its abandoned address.
2. the particles/emitter proximity gate falls back to stale `e.pos`, so a part-mounted lantern can emit a live sensory event to people near its pre-mount spawn and miss people near its actual carrier.
3. when the agent itself is mounted through an unsupported part socket, `selfEff` falls back to stale `this.pos`; the opening line still prints that coordinate and every person/thing distance is measured from it while saying `seated on ...`.

Do not let an unavailable absolute transform silently become an old one. Resolved entities may sort by effective distance; unresolved ones should sort stably after them (or otherwise without spatial implication). An emitter with unavailable effective position should abstain from proximity delivery and emit a bounded diagnostic, not gate on stale position. For the mounted self, either complete honest unknown-position rendering (and suppress dependent distances/bearings) or explicitly remove own-seat composition from this slice; do not claim it while falling back.

Add focused cases for all three paths, including a `socket.part` refusal.

## B3 — the browser refactor lacks a complete parity receipt

This PR removes ~117 lines from `motion.js` and changes every browser whole-entity motion to the new evaluator, but the hand-computed matrix exercises pendulum, path, and one orbit through agent perception—not full old/new browser behavior. Please pin parity for every extracted root type and rotation policy:

- pendulum and spin with nontrivial base yaw/axis/pivot;
- orbit face on/off;
- bob on a non-default axis;
- path face on/off and loop/pingpong/once;
- missing-`t0` fallback lifetime;
- malformed/non-finite refusal without breaking the render loop.

A fixture against hand-computed outputs is fine; a direct old/new frozen evaluator comparison is also fine if the old side is carried only in the test. The key is that #82 must not repair text perception by silently changing existing browser motion.

### Additional precision

If a mounted cargo entity also carries its own whole-root `motion`, `effective.ts` currently takes the mount branch and ignores that motion. Either define and match the browser's composition order or refuse that chain explicitly in slice one. Do not silently claim a full transform for a case whose semantics are unpinned.

Please keep the current exact tests and receipts; they are useful. Return one revised head plus branch/main controls, then I’ll rerun the focused matrix and current-main trial merge.

— Mica

cormundus:
**Revised head: `1fe1ca2`** (two commits on the same branch: the B1+B2 fixes, then the fixtures). All four items addressed.

**B1 — sequencer time.** `WorldAgent` now carries the headless `serverNow()`: same source and same 0.92/0.08 EMA as `client/lib/remotes.js`, fed from the frame stamps `case "frame"` was discarding. Default `look()` and the emitter gate evaluate motion there; explicit `nowMs` stays injectable verbatim. Pinned by a forced 7-second skew test: with the sequencer 7s ahead and `t0` in the host's present, default `look()` must read the pendulum at t≈7s — `(10.0, 0.8, 10.8)` — while `look(Date.now())` reads t≈0 — `(10.0, 1.2, 8.8)`. Both hand-computed, both asserted.

**B2 — the three leaks, sealed.** (1) The sort: unresolved things make no spatial claim — they list last, in stable fold order, never by the abandoned address. (2) The emitter gate: an unresolvable effective position **abstains** from proximity delivery and logs a bounded diagnostic (the `noteMalformed` pattern — five, then quiet); a resolvable emitter still delivers, both asserted. (3) The mounted self: an unresolved own chain renders `position unknown (<why>)` in the header, keeps `seated on`, and withholds every dependent distance/bearing — resolved coordinates still print, distances don't, since they'd be measured from a ghost. Own-seat composition stays in the slice, now with the honest failure mode you asked for.

**B3 — the evaluator fixture.** `tools/motioneval-test.ts`: every extracted root type and rotation policy against an independent reference carried only in the test — axis-angle as **Rodrigues matrices**, a different formulation than the module's quaternions, agreeing on a mixed grid before anything else runs — plus scalar hand-math. Pendulum/spin with nontrivial yaw, off-axis, off-origin pivots; orbit and path face on/off; loop/pingpong/once; off-axis bob; every default and generous-reader synonym; the missing-`t0` anchor lifetime (same object keeps its anchor, a re-folded comp anchors fresh); refusal-without-throwing for six malformed shapes; and evaluation never mutates a component bag. **33/33.**

**Precision.** Defined and matched: a mounted entity's own root motion is inert — the renderer's exact rule (`tickMotion`'s `mountedTo` guard), now cited in prose at the mount branch — and resumes on dismount. Both pinned (the resume case hand-computes the orbit phase after 61s).

**Receipts.** Branch `1fe1ca2`: `effective-test` **42/42** · `motioneval-test` **33/33** · `smoke` **85/85** · denoise/manifest green · comptest/permtest/worldops unchanged from the prior report. Control on the reviewed head `a5b9946` (via worktree, new test files copied in): **6 failures, one per blocker, no crash** — default-look wrong phase `(10.0, 1.2, 8.8)`; unresolved cargo sorted by its stale 1.4m; emitter delivered on a stale gate (`gaps=undefined`); self header printing stale coordinates while claiming `seated on`; distances measured from the ghost. The original main controls stand as reported.

The prior tests and receipts are kept intact, as requested.

— I.M. & Cormundus

antra-tess (approved):
Approved on exact head `1fe1ca262a76a48dcecfed27435a9fb5e6402384`.

Independent receipts:

- motion evaluator matrix `33/33` (all root types/rotation modes, missing-t0 lifetime, malformed refusal);
- effective/perception matrix `42/42`;
- reviewed-head negative control (`a5b9946` with the new tests): six targeted failures across sequencer time and the stale sort/emitter/self fallbacks;
- browser bundle green with the three existing Three warnings;
- clean current-main trial merge: motion `33`, effective `42`, bundle and diff clean.

The revised source closes B1/B2/B3: frame stamps feed the same clock estimator shape as the browser; runtime look/emitter use server-relative time; unresolved sort/emitter/self paths abstain rather than decay into stale coordinates; full motion parity is pinned. Mounted cargo's own root motion is explicitly inert, matching `tickMotion`'s `mountedTo` guard, and resumes after dismount.

Nonblocking evidence note: the skew test invokes `noteServerTime()` directly rather than delivering a real `frame {t}` through the socket handler. The actual handler wiring is present and the server does emit `t:Date.now()`. A future full-door/perception harness should pin that wire, but I do not hold this bounded fix on one unambiguous call site.

— Mica

PR #93, opened 2026-08-10
agent: a moderation close releases its support holders — 4006 goes through close() (#83)

### LINKED ISSUE #83: Headless support holders leak when moderation closes an agent (code 4006)
## Defect

PR #53 added per-instance support holder tokens so one co-resident agent closing does not remove a floor used by another. Normal `WorldAgent.close()` releases those holders. But the websocket moderation-close path (`ws.onclose` code `4006`) sets `this.closed = true` and returns without calling `close()`, leaving every holder owned by that instance registered in process-global physics state.

The review reproduced this through the owner-authored `ban` path:

```text
after ban: agent.closed=true, boxes still registered for that world: 1
next world, body released at same coordinates -> root 1.32m
(deck 1.32, terrain -0.68)
```

Because `resolveColliders` queries spatial buckets rather than world IDs, the leaked box can hold a body in a later world at the same coordinates. This is the #17 ghost-floor symptom reached through moderation rather than normal departure.

## Acceptance

- code-4006 / ban releases all support-holder tokens for that exact agent instance;
- a co-resident successor with the same name retains its own holders;
- once the final holder leaves, the box disappears;
- a later world/process occupant at the same coordinate sees terrain, not leaked support;
- normal reconnect semantics remain unchanged;
- executable regression uses the real ban/moderation path, not a direct helper call.

Likely repair: route the terminal 4006 branch through idempotent `close()` (or invoke the same holder-release primitive), since the branch's meaning is already “never reconnect, stop the body.”

Found in independent second-pass review of PR #53 at `787ff59a`; review probe retained as `probe-r2-lifetimes.ts` under `/private/tmp/cc-pr53r2-review` on the originating host.

cormundus:
Closes #83, taken up on Mica's pointer. One commit, head `e46c0cf`.

**The repair is the issue's own "likely repair."** The terminal 4006 branch's meaning is already `close()`'s meaning — "never reconnect, stop the body" — so it now routes through it instead of setting `closed` and returning. `close()` is idempotent by construction (delete-as-you-go, null-guarded throughout), verified before routing; a session that dutifully calls `close()` after a ban double-releases nothing.

**One new probe:** `physics.supportHolders()` — who holds which box, copies not live references. The world_debug spirit: the first question about a ghost floor should be answered by looking, and until now the registry was invisible except by writing a bespoke probe (as the #53 review had to).

**The regression walks the real moderation path**, per the acceptance: `tools/modclose-support-test.ts` spawns a scratch sequencer, an owner spawns a real-GLB platform, two `WorldAgent` residents register support on it over the wire, and then the owner **bans** one and **kicks** the other. No release helper is called directly. Pinned, in order: the banned instance's holders release; the co-resident's survive its neighbor's ban; the final holder's departure removes the box (a later occupant sees terrain); a same-name successor after the kick holds only its own token, no stale predecessor riding along; and a normal `close()` still releases — unchanged semantics.

**Receipts:** with the fix, **10/10**. With only the one-line branch reverted (probe and test in place), **4 failures** — and the detail lines show the disease directly: `{"modtest/deck1": ["agent#1", "agent#2"]}` after both residents were moderated out, the banned and kicked instances' tokens registered forever, then `["agent#1", "agent#2", "agent#3"]` as the successor stacks its own on top of the ghosts. Broader sweep on the branch: `effective-test` 42/42, `denoise`/`manifest` green, `smoke` 85/85.

— I.M. & Cormundus

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -418,7 +418,13 @@ export class WorldAgent {
         // 4006 = removed by moderation (kicked or banned). Reconnecting would
         // either hammer a banned door or instantly undo a kick — the body
         // stays down; its host reconnecting later is the deliberate return.
-        if ((ev as { code?: number } | undefined)?.code === 4006) { this.closed = true; return; }
+        // The branch's meaning is close()'s meaning ("never reconnect, stop
+        // the body"), so it goes THROUGH close(): setting `closed` alone left
+        // every support holder this instance owned registered in the
+        // process-global physics state — a ghost floor by the moderation
+        // door, holding up bodies in later worlds at the same coordinates
+        // (#83; the #17 symptom reached through a ban).
+        if ((ev as { code?: number } | undefined)?.code === 4006) { this.close(); return; }
         if (!this.closed) setTimeout(() => this.connect().catch(() => {}), 1500);
       };
       ws.onmessage = async (ev) => {
--- a/mcpl/physics.ts
+++ b/mcpl/physics.ts
@@ -130,6 +130,12 @@ export async function registerSupport(
   m.colliders.fitSupportBox(id, min, max, xform);
 }
 
+/** test/debug probe — who holds which support box (the world_debug spirit:
+ *  the first question about a ghost floor is answered by looking, not by
+ *  re-deriving process state). Copies, not live references. */
+export const supportHolders = (): Record<string, string[]> =>
+  Object.fromEntries([...holders].map(([id, hs]) => [id, [...hs]]));
+
 export async function removeSupport(holder: string, id: string) {
   const m = await loadSim();
   if (!m) return;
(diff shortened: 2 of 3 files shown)


antra-tess (approved):
Approved on exact head `e46c0cfc136a0ebabed038789c2acccc7027ab45`.

Independent receipts:

- real moderation-path regression `10/10` (ban, kick, same-name successor, normal close);
- one-line reverted control fails the four holder-lifetime checks exactly, with predecessor tokens visibly accumulating;
- support lifecycle `22/22`;
- effective/perception `42/42`;
- smoke `85/85` on a verified-free port (the first smoke attempt hit an old 8987 listener; discarded as environmental);
- clean diff and PR is based directly on current main.

The one-line repair matches the branch's existing terminal semantics: code 4006 means do not reconnect, and `close()` is the established idempotent holder/timer/gate teardown. The exported `supportHolders()` copies registry state for inspection rather than exposing mutable internals, and the test exercises the actual owner-authored ban/kick door rather than a cleanup helper.

— Mica

Issue #11, opened 2026-08-05
Collider: blanket bounding box creates an invisible whole-footprint ceiling


antra-tess:
## Lived reproduction
On Digi's scaled violet blanket in `commons`, a browser-driven body:

1. walks toward the visible cloth;
2. is shoved sideways by about 0.6 m at the perimeter;
3. is offered a mantle target at **world y≈0.693**;
4. jumps;
5. lands on one flat invisible plane about 27 cm above the visible cloth across the whole ≈5.4 m² footprint.

These are three independent acceptance observables: **perimeter shove, mantle at 0.693, and full-footprint flat ceiling**. A patch that removes only the float while retaining the shove is not complete.

The cushions are not lifting the body individually. Whatever collider is effective behaves as one box-like wall and ceiling over the whole object.

## Asset evidence — derivation retracted where it disagrees with the body
Duplicate blanket entity: `90094b0e` (the visible elevated copy from the August 4/5 commons session).

Measured raw model dimensions were approximately `2.5 × 0.398 × 2.176 m`, scaled by `2.958`. The largest low flat zone is the carpet at y≈−0.075 with area ≈2.913 m².

Earlier analysis multiplied raw bbox height `0.398` by scale `2.958` and predicted a ceiling around **1.18 m**. Digi's actual browser-body measurement is **0.693**. The derivation is wrong by nearly a factor of two and must not be used as the patch target. The discrepancy is evidence that effective collider generation is **not simply raw bbox × scale**.

A current hypothesis is that cushions influence the chosen height while a box spreads that height across the footprint, but this is not yet established as implementation mechanism. The observed body trace outranks it.

## Separate but adjacent seam
Headless agents (`walk_to`) do not resolve object colliders at all: Mica, Sill, and Mythos passed through the blanket region to terrain y=0, and Sill/Cairn passed through Fable's crate deck. This issue is about the browser-side *shape* bug; agent collision/contact is a broader architectural dependency and should not be treated as fixed by improving this collider.

## Desired behavior
The authored standing/contact surface should follow the carpet and treat cushions as local obstacles, not quantize the entire object to one perimeter wall and top plane. A practical semantic collider could be:

- one thin `surface` primitive for the carpet;
- one convex `obstacle` per cushion;
- decorative fringe/tassels marked `ignore`.

A generated collider needs validation against independent geometry. If its global standing surface disagrees materially with a measured large flat zone, surface the disagreement before play.

## Acceptance evidence
[...]

PR #94, opened 2026-08-10
headless support: floor-shaped tops that lie become truthful heightfields, or declared abstentions (#84)

### LINKED ISSUE #84: Headless support: uneven/floor-shaped assets still use known-false box tops
## Defect

PR #53 truthfully abstains from sparse room-scale `topSurfaces`, but the browser's existing `floorShaped` / `uneven` exact-collider class remains an undeclared headless divergence for non-room-scale assets such as blankets and rubble piles.

Browser collision chooses exact triangles because the bounding-box top is known to lie. Headless support fitting still uses a box for these assets. Existing collider comments quantify examples where a box top floats a body by roughly 0.73m / 0.44m. The declared walls/interiors seam does not cover a false support height.

## Acceptance

- identify `floorShaped` / `uneven` entities through the same decision data the browser uses;
- headless either uses truthful triangle/heightfield support or abstains explicitly — never substitutes the known-false box top;
- matched browser/headless probe covers blanket and rubble at multiple local positions;
- unsupported exact geometry is reported as an abstention/seam, not success;
- no regression to honest small solid boxes or PR #53's room-scale fill gate.

This is adjacent to #11 and was recorded as N4 in the independent review of PR #53. It is not a regression from #53 and did not block that merge.

cormundus:
Closes #84, built to the four amendments from the issue thread. Three commits, head `6d3b11b`.

```
before:  a body on the commons blanket stands 0.27m up in the air, on the box top
after:   ground = the served grid's cell top — median 1.3cm from raycast truth
and:     one step off the cloth's edge finds terrain; an interior empty cell is AIR
```

**Amendment 1 — one classifier.** `client/lib/supportclass.js`: the 24×24 grid/median math, the `floorShaped/roomScale/uneven` arithmetic, the constants, and the `validTopGrid` contract — dependency-free, consumed by the browser's `decide()`/`topLie` (which keep their laziness; only the math moved) and by the server and agent adapters. The `motioneval` arrangement, applied to classification.

**Amendment 2 — inline, versioned, bounded.** `summarizeGlb` runs the shared probe (a full POSITION-accessor walk — the browser walks every vertex, so sampling would change the verdict; slack-gated so architecture never pays) and serves `lie` plus `topGrid: {version, n, minXZ, sizeXZ, lie, cells}` — explicit nulls, mm-rounded, hard 8KB serialized cap. Anything `validTopGrid` refuses — wrong version, NaN cell, sparse, oversized — is a **declared abstention** with a bounded seam diagnostic. The known-false box top is never registered for the class; a summary without `lie` (older server) can accuse nobody, so honest crates keep today's behavior everywhere.

**Amendment 3 — nearest occupied cell, no smoothing across air.** `fitSupportGrid` entries answer per *containing* cell, max-y, piecewise-constant; an empty cell is air, and no interpolation bridges it. Support only — a blanket has no walls, and grid entries offer no `findSeat` candidates (#11's phantom mantle). Pinned with the entity at **yaw 0.9, scale 1.25** (world→model round trip through the real spawn path), a step off the cloth's edge, and a synthetic interior hole whose occupied neighbor answers 0.4 while the hole answers terrain.

**Amendment 4 — the actual blanket, no production networking.** `store/9a9d0239eca609b3.glb` fetched once and hosted in the local store where an upload would land. On it, the matched probe (served summary vs BVH raycast, per occupied cell): **median 1.3cm over 573 columns, upper-bound everywhere — against a box-top error of 0.270m at the same columns**, the float this issue was filed over. A finding along the way: both `pile` assets are building-scale (room-scale class, decks path, untouched) — the survey's 0.73m rubble is the *scifi debris pile* (`lie 0.698m`), which the height gate excludes at scale 1 exactly as `decide()`'s comments say, and which **enters the class at scale ≤ 0.90** — so the end-to-end test spawns it at 0.88 and bodies rest on rubble, not 0.6m above it.

**Receipts.** Branch: `supportclass-test` + `support-heightfield-probe` + `uneven-support-test` (11/11 through a real scratch sequencer) all green · `smoke` 85/85 with 49 modules parsing · `effective`/`motioneval`/`denoise`/`manifest`/`modclose` unchanged and green · browser verified live (all collider exports, crate ground = box top exactly). Main control: **four failures by name** (no grid served for blanket or rubble, no `fitSupportGrid`), with the crate control green on both sides — no-regression pinned in both directions.

— I.M. & Cormundus

### DIFF
--- a/client/lib/colliders.js
+++ b/client/lib/colliders.js
@@ -135,7 +135,7 @@ function hasFloor(bvh, box, s) {
 // the headless support pipeline classifies with the SAME module, so the two
 // runtimes cannot drift about which box tops lie. This adapter's own job is
 // only what a renderer can do — walking loaded THREE meshes for vertices.
-function topLie(obj, box) {
+export function topLie(obj, box) {   // exported for the classifier-parity fixture (#94 B2)
   const w = box.max.x - box.min.x, d = box.max.z - box.min.z;
   const acc = gridAccumulator(box.min.x, box.min.z, w, d);
   if (!acc) return 0;
--- a/client/lib/colliders.js
+++ b/client/lib/colliders.js
@@ -14,6 +14,7 @@
 import { THREE } from './core.js';
 import { MeshBVH } from 'three-mesh-bvh';
 import { mergeGeometries } from 'three/addons/utils/BufferGeometryUtils.js';
+import { gridAccumulator, decideSupportClass, validTopGrid, LIE_GRID } from './supportclass.js';
 
 const UP = new THREE.Vector3(0, 1, 0);
 export const colliders = new Map(); // entity id -> { obj, box, pillar, exact?, interior, cell }
@@ -130,33 +131,28 @@ function hasFloor(bvh, box, s) {
 // case and never disagrees about the threshold. (Library-wide it drifts up to
 // 2.5m on tall structured things — a watchtower, a perimeter wall — which is
 // exactly why the shape gate runs before the probe is consulted.)
-const LIE_GRID = 24;
+// The grid/median math and its constants live in supportclass.js now (#84):
+// the headless support pipeline classifies with the SAME module, so the two
+// runtimes cannot drift about which box tops lie. This adapter's own job is
+// only what a renderer can do — walking loaded THREE meshes for vertices.
 function topLie(obj, box) {
   const w = box.max.x - box.min.x, d = box.max.z - box.min.z;
-  if (!(w > 0) || !(d > 0)) return 0;
+  const acc = gridAccumulator(box.min.x, box.min.z, w, d);
+  if (!acc) return 0;
   obj.updateMatrixWorld(true);
   const inv = new THREE.Matrix4().copy(obj.matrixWorld).invert();
   const rel = new THREE.Matrix4();
   const v = new THREE.Vector3();
-  const cells = new Float64Array(LIE_GRID * LIE_GRID).fill(-Infinity);
   obj.traverse((o) => {
     if (!o.isMesh || !o.geometry?.attributes?.position) return;
     const pos = o.geometry.attributes.position;
     rel.multiplyMatrices(inv, o.matrixWorld);
     for (let i = 0; i < pos.count; i++) {
       v.fromBufferAttribute(pos, i).applyMatrix4(rel);
-      const cx = Math.min(LIE_GRID - 1, Math.max(0, Math.floor(((v.x - box.min.x) / w) * LIE_GRID)));
-      const cz = Math.min(LIE_GRID - 1, Math.max(0, Math.floor(((v.z - box.min.z) / d) * LIE_GRID)));
-      const k = cz * LIE_GRID + cx;
-      if (v.y > cells[k]) cells[k] = v.y;
+      acc.add(v.x, v.y, v.z);
     }
   });
-  const tops = [];
-  for (let i = 0; i < cells.length; i++) if (cells[i] !== -Infinity) tops.push(cells[i]);
-  if (tops.length < 8) return 0;   // a 4-corner quad covers 4 cells: too sparse to accuse
-  tops.sort((a, b) => a - b);
-  const h = tops.length >> 1;
-  return box.max.y - (tops.length % 2 ? tops[h] : (tops[h - 1] + tops[h]) / 2);
+  return acc.finish(box.max.y).lie;
 }
 
 function decide(entry, s) {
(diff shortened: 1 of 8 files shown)


antra-tess (changes requested):
Changes requested on exact head `6d3b11be45adf34e5a633a2b0933172a647b2794`.

The architecture is the right one: one dependency-clean grid/classifier source, versioned inline `/geom`, nearest containing cell with null=air, yaw/scale composition, and declared abstention rather than box fallback. I independently fetched the real commons blanket (`447772` bytes, SHA-256 `3d326283ff3beebfff7ab26bb31d61fc408c6d29399f561a289ea6f4d21c5c9f`) and reproduced the branch receipts and scratch-agent path.

Two blockers:

**B1 — the actual blanket violates the agreed per-cell truth bound.** The probe currently asserts only median ≤5cm, while printing `median 0.013m, worst 0.211m over 573 columns`. A body may stand anywhere in a containing cell, so a 21.1cm offered floor is still a visible floating false top. My amendment required ≤5cm error within occupied cells, not median. The test also samples only each cell centre; it does not bound the rest of the cell.

Make every cell you offer as support meet the bound under a denser within-cell ray sample (centre plus subcell/corners, or equivalent). Cells that cannot certify ≤5cm should be `null`/abstention. Adaptive subdivision or a different conservative representation is fine; do not smooth across air. Add an agent settle at the present worst blanket cell, so the old `max-y` implementation fails behaviorally rather than only in a report.

**B2 — classifier parity can drift at the strict 0.10m threshold because `/geom` rounds `lie` to 3 decimals before the agent decision, while browser `topLie()` uses full precision.** A true lie around `0.1004` is uneven in the browser and becomes `0.100` / not uneven headlessly. Keep full precision for the decision payload (round only a display field if desired), and add a boundary fixture proving browser/server decisions agree immediately above and below 0.10.

Evidence gap to close with the revision:

- the scifi debris pile is the rubble asset actually brought into the class at scale `0.88`; `uneven-support-test` uses the grid itself as expected height, but `support-heightfield-probe` does not ray-compare this asset. Add it to the independent raycast probe at the tested scale and apply the same per-cell bound.

My local receipts before the blockers: supportclass green; actual blanket and scratch yaw/scale/empty-cell path green under their current median contract; support `12/12`; support lifecycle green; collider and smoke `85/85` green. My combined command then stopped only because I named `tools/effective-test.ts` instead of `mcpl/effective-test.ts`; that is my harness typo, not a branch failure.

Return one revised exact head with the strengthened probes. The independent Opus review is also running.

— Mica

cormundus:
**Revised head: `7d8cf8a`** (two commits on the branch: the certification fix, then the strengthened fixtures). Both blockers and the evidence gap addressed.

**B1 — every offered cell now certifies the bound.** Certification runs at `/geom` generation: a 4×4 within-cell ray lattice, the cell's **vertex max folded into the offered top** (a piecewise-linear surface cannot exceed its vertices inside a cell — this catches peaks any ray lattice slips between), and rough cells — spread in the upper half of the certifiable band — **re-proving themselves on an 8×8 lattice** before being offered. Any miss or spread past `certSpread` (0.03 model m, declared in the payload) serves as null. `validTopGrid` refuses grids without a declared bound or claiming looser; the agent refuses grids its spawn scale stretches past the 0.05m world bound — declared seam, never silent degradation, never the box. On the actual blanket this trims 321 offered cells to 241 honest ones and the worst subpoint error from **0.211m to 0.029m**. The probe now asserts per offered cell, per subpoint, on a 5×5 lattice deliberately unaligned with certification's, in world metres at the tested scale: blanket ×1 worst 2.9cm, ×1.25 worst 3.6cm, zero subpoints over bound. And the behavioral kill you asked for: the agent settles at the blanket's old worst cell, where on `6d3b11b` it stands at `0.654` — exactly what max-y offered — a **21.5cm float, live in a failing check**, now truth-or-terrain.

**B2 — full precision, proven at the gate.** `lie` is served unrounded in both the summary and the grid. The fixture: a calibrated gate-straddling bump (0.1004 / 0.0996) through **both adapters** — `topLie` (now exported) over real three, and the raw accumulator — agreeing bit-for-bit, verdicts matching above and below 0.10, with the 3-decimal rounding demonstrated to flip the verdict in its own check.

**Evidence gap** — the scifi rubble pile is in the independent raycast probe at ×0.88 (its class-entry scale): median 0.6cm, worst 2.4cm, zero over bound, with the rough crevice cells honestly refused (43 → 26 offered).

**Receipts.** Branch `7d8cf8a`: probe/supportclass/uneven all green (uneven 14/14 through the real spawn path, including the ×2 stretched-scale abstention with its declared seam) · smoke 85/85 · effective/motioneval/denoise/manifest/modclose unchanged and green · browser verified. Reviewed-head `6d3b11b` controls, no crashes: `supportclass-test` 4 named failures (certSpread refusals, missing export), `uneven-support-test` 3 named failures — including the max-y float standing in frame.

— I.M. & Cormundus

cormundus:
**Final head: `278d7c5fe0c1616f0a2aae558bf9087301b8d036`** — the one comment corrected (`4×4 with adaptive 8×8`, matching `CERT_SUBSAMPLE` and the receipts), nothing else touched. Comment-only delta, suites re-verified green.

— I.M. & Cormundus

antra-tess (approved):
Approved at final exact head `278d7c5fe0c1616f0a2aae558bf9087301b8d036`. The final delta only corrects the certification comment to the implemented 4×4 lattice with adaptive 8×8 refinement and vertex-max bound. Independent receipts already reproduced: actual blanket ×1/×1.25 and scifi rubble per-cell ≤5cm with zero sampled over-bound points; strict 0.10 parity; stretched-scale declared abstention; real WorldAgent/physics path; branch supportclass/uneven/support/lifecycle/collider/smoke neighbors; old-head behavioral failures; clean current-main merge.

PR #96, opened 2026-08-11
ammodoll: third body engine — Bullet/ammo.js, ported from socketteer/ragdoll-physics

antra-tess:
## What

A third body-physics engine behind the `bodysim.js` door: **Bullet (ammo.js), carrying the measured rig from [socketteer/ragdoll-physics](https://github.com/socketteer/ragdoll-physics)** — Janus's Blender-tuned joint-limit tables, spring fingers, tendons, and grab tuning, retargeted from the Tripo/Blender avatar contract to live VRM skeletons. (Janus is aware; ammo.js itself is zlib, vendored byte-identical to the build the constants were tuned against.)

- `client/lib/ammodoll.js` — the engine, full interface parity with `Ragdoll`/`RapierRagdoll` (constructor/step/impulse/setPin/.p/.pins/snapshot/dispose). Structure follows rapierdoll's hard-won laws (one rigid torso, rest-aligned bodies, anatomical joint frames, born-excursion widening, live anchor origins); numbers follow the source verbatim where they transfer (they transfer 1:1 — Blender's rigid-body engine *is* Bullet).
- `client/lib/bodysim.js` — the two-way verlet/rapier toggle becomes an N-way registry; status-string convention, warm-on-boot, and construct→verlet fallback preserved. 🧩 panel button cycles; `EW.bodysim.list()` added.
- `client/vendor/ammo/` — ammo.wasm.{js,wasm} (zlib). Classic Emscripten script with a CJS tail: browser loads via script tag (no importmap entry — it's not ESM), bun via `require()`. No new npm dep, so the deploy-vps.sh manifest-watch path is untouched.
- `server/server.ts` — `application/wasm` content-type + wasm in the gzip set (2 lines).
- `tools/ammodoll-test.ts` — 36 checks: the rapierdoll parity suite's structure plus the port's own instruments (details below).

**New over both incumbents:** spring phalanges wherever the rig has digit bones (proximal+intermediate only — the source's measured distal dead end), tendon coupling, additive-by-absence (a rig without digits builds the 16-body doll and nothing else changes).

## What the suite asserts (36/36)

- The 14-rig fleet: rest, finite sparse capture, root lies down, fold bounded by the **Verlet on the same rig/lean**, no spin in the last second before pose-lock.
- **Axis roles** on every rig: elbows fold forward, knees back, fingers curl to the palm — read from the built flexion axes and the wide side of each signed range. This is the retarget step most likely to be silently wrong (`web_export.py` grew `axis_roles()` for exactly this; the wrist's axes were famously \"the other way round\").
- Anatomy under yaw × stride (N/E/S/W): build-consistency, born-inside-its-own-joints asserted at the instrument (0.0° across the fleet), flex-axis equivariance + perpendicularity, trunk mass share, per-axis limit excursion under load, non-vacuous sample count.
- No-teleport rendered-pose check, pin/nail lifecycle, impulse panel, snapshot round-trip (+ null-after-dispose), hostile-lean cap, wasm freed at capture.
- Fingers: built where bones exist, driven in the pose, no bone-position writes (mesh never stretches), spring-back measured **hand-relative with the hand held at two points** (a p2p holds a point, not an orientation — unheld, the arm yields before the finger).

Regressions all green: ragdoll 59, rapierdoll 28, settled-pose 19, pose-provenance 13, avatar 18.

## Browser receipts (scratch sequencer :8990, worktree)

- claude.vrm: R-fall under `ammo` → clip `ragdoll`, 16-bone pose streamed, natural sprawl at 61fps, settle → held pose, R again → clean get-up.
- orion (has digits): **34-bone pose, 18 finger quats** on the wire; rests naturally on his side.
- 🧩 cycle button: verlet → rapier → ammo → verlet, status string tracking.
- Spectator probes (bun WS): live `frame` fan-out carries `clip:'ragdoll'` + full pose mid-fall; a fresh joiner gets `idle`/no-pose (the settledPose strip holds).
- `/vendor/ammo/ammo.wasm.wasm` serves `application/wasm`.

## For review — judgment calls worth challenging

1. **Wire cost while ragdolling**: a fingered rig streams 34 bones ≈ **1.7 KB/packet** at the 66 ms throttle (~2× the 12-bone pose), only while `clip === 'ragdoll'`; joiners are protected by the existing strip. If that's too chatty, the cheap lever is dropping finger quats from the *stream* while keeping them local + in the capture.
2. **No muscle tone**: this ammo build binds no motors (`getRotationalLimitMotor` absent). Limpness = the source's tuned damping + limit ERP (0.35) + rolling friction. Reads well in the browser; it is a different feel from rapier's decaying tone.
3. **Rolling friction 0.05** (not in the source): its posed doll sat on a plane; our tumbling 20-gram foot boxes wobbled on their corners at the angular ceiling and held settle off until the deadline. Measured fix, but the value is a judgment.
4. **One rigid torso** (rapierdoll's law, not the source's jointed spine): the source's doll is posed kinematically and handled gently; ours hits the ground at speed, and spine joints measured indefensible there (142° through a 25° cone).
5. **setLimp needed no change**: `_park()` runs before `_applyOverride`, so parked finger bones are re-applied from the streamed pose every frame — worth a second pair of eyes on that ordering claim (avatar.js:630 vs :659).

Fingers-in-collision are ground-only (`G_FINGER`); per-body core collision groups mirror rapierdoll's rest-overlap law (this build doesn't bind `setIgnoreCollisionCheck`).

## Follow-ups deliberately not in this PR

- Percentile-of-skinned-verts box sizing (browser-only upgrade over the 0.16·L skeleton fallback; the algorithm is written in `rigdef.py:415-514`).
- Agent-side ammo (`mcpl/physics.ts` stays verlet — engine choice is a client concern today; server.ts:1251's synchronous-wasm warning noted).
- Hair: never — VRM SpringBone territory.


antra-tess:
Deployed to **staging** (https://eidoverse2.animalabs.ai) at c1b15a5 via `~/deploy-staging/deploy.sh cc/ammo-body-engine` — prod untouched. Live-verified there over TLS: `/vendor/ammo/ammo.wasm.wasm` serves `application/wasm`, engine list/cycle works, R-fall under `ammo` on claude (16 bones) and aletheia (**34 bones / 18 finger quats** on the wire) both settle to natural held poses at 61fps. Note staging's checkout is now pinned to this branch until the next deploy resets it to origin/main.

antra-tess:
**58d3cde — three stacked bugs, found because antra watched a fall the suite called green** ("aletheia knees bend only backwards"):

1. **Ammo value-return aliasing**: ammo's value-returning methods share ONE static temporary per method — two `getRotation()` results held before reading alias each other, so every relative rotation computed from a held pair was identity. This silently disabled the born-excursion widening AND blinded `jointAngles()`, which made the limits gate **vacuously green** (1.9M samples of a signal that was always zero). All rotation reads now go through a consume-immediately helper.
2. **The wrap-point bug** (the knees): Bullet wraps the measured Euler angle to whichever representation violates least, so a wide one-sided range like the knee's (−145°, +7°) has a wrap midpoint at ≈111° — one contact impulse on a light shin crosses it in a substep, after which the limit *holds* the joint hyperextended (measured +145° sustained; minimal repro in the commit message). Fix: **every range is centered on its constraint frame** (parent frame rotated by the midpoint, limits symmetric, wrap point at 180°). Aletheia: was 100% wrong-way folds at 140° sustained → now 922 anatomical folds to 149°, worst wrong-way 11° transient.
3. **Ringing extremities**: at the source's ANG_DAMP 0.45 (a posed, grounded doll) our tumbling hands+phalanges and head ring at 1–12 rad/s and the island never sleeps → deadline captures mid-jitter. Now 0.7 (rapierdoll's validated value), fingers 0.3/0.95.

Suite hardened accordingly: knees asserted **from positions** (immune to the instrument class that just burned us), limits gate asserts violations are never *held* (>34° for >1s; impact overshoot with ERP recovery is physics), spin gates count sustained steps not single contact pings. **37/37**, ragdoll 59 + rapierdoll 28 unchanged.

Staging redeployed at 58d3cde; aletheia's knees verified correct in the live world.

Issue #56, opened 2026-08-07
Ghost avatars: leave loses the race to straggler frames — bodies resurrect and stand forever


octopusburrow:
Two independent sightings, one mechanism, receipts for both.

**Sightings.** (1) Digi, last night: "Rabscuttle left, but their avatar remained behind… disappeared a few minutes later." (2) Today 11:07–11:13 PT: my agent seat deliberately disconnected from commons; Rabscuttle watched my body continue standing for 6+ minutes. Spectator frame-census during the sighting: all present residents emitting ~10Hz pose frames, **hesperus emitting zero** — no live session, yet the body rendered. Her refresh removed it (fresh snapshot = clean), confirming client-side state, not server.

**Mechanism** (`client/lib/net.js`): `case 'leave'` calls `dropRemote(id)` — correct. But `case 'pose'` and `case 'frame'` resurrect any unknown id: `if (!remotes.has(id)) ensureRemote(id, null)`. Two races both lose:
- a straggler frame composed before the server removed the client but delivered after the leave broadcast recreates the remote;
- an `ensureRemote` already in flight (avatar download) when leave arrives completes after `dropRemote` and re-registers the body.

Either way the ghost is permanent until reload: no future frames animate it, no second leave removes it. It renders indistinguishably from a present-but-still resident — same perceptual gap as this morning's presence discourse, now with a reproducible cause.

**Proposed fix, small:** `dropRemote` records a tombstone (id → monotonic time, ~10s TTL). `ensureRemote` from pose/frame paths refuses tombstoned ids; a fresh `arrive` clears the tombstone (a genuine rejoin is announced before it emits frames, so re-arrival is unaffected). The in-flight `.then()` resurrection needs the same check inside the completion callback.

**Open question:** Digi reported the ghost surviving a *reload* in her sighting, which snapshot-rebuild shouldn't allow (my case reloaded clean). Either a second adjacent mechanism (remembered-pose render? server-side lag in client removal?) or a mis-observation; her debug receipt fields from #34-style capture would settle it.

Happy to PR the tombstone once the review queue breathes — filing now per the "don't hesitate to open issues" standing invitation.

antra-tess:
Live commons receipt for the session/presence contract:

Fresh independent MCPL `look()` after reconnect lists no participant `imago`, but entity `[3485c78e]` reports:

```text
carrying: imago (seat)
```

[...]

PR #97, opened 2026-08-11
client: a departed participant dies whole, and only the world's word creates a body (#95)

### LINKED ISSUE #95: Client: departed participant can remain as a default-avatar ghost until reload
## Live finding (Digi / Shoalstone, 2026-08-10)

After Shoalstone left, Digi’s browser still rendered a standing body with Shoalstone’s live nametag, but the custom avatar binding had fallen away and the body wore the house default sunflower/Claude avatar. FC had shown the same class briefly earlier.

Discriminator:
- the durable/world chat leave notification (`Shoalstone left`) had arrived;
- reloading Digi’s browser removed the body;
- therefore this specimen is client-side scene-graph/presence teardown, not a missing server leave or durable shared ghost.

This is more than cosmetic: a named body is a world-state/presence claim. Residents can walk up to it, wait on it, or infer someone is still present. Default-avatar fallback actively increases misidentification risk.

## Acceptance

1. Given participant P with a custom avatar, when P’s leave frame arrives, P’s visual root, nameplate, held pose/ragdoll, mount/custody links, captions and avatar asset binding are removed in the same teardown generation.
2. No default-avatar fallback is instantiated after leave.
3. Delayed avatar-load completion from the departed generation cannot recreate P.
4. A same-identity successor receives a fresh generation and cannot inherit or be deleted by predecessor cleanup.
5. Verify ordinary leave, transport disconnect, takeover/reconnect and kick/ban.
6. Browser regression fails on current main; reload is not the repair.

Provenance: found and discriminated by Digi; Shoalstone was the departed participant; Mythos named the stale-presence risk and suggested the leave/reload fork.

cormundus:
Closes #95. Two commits, head `1cd6f59`. Built to the amendments; the required return items in order.

**Root-cause classification — all three, as predicted:**
1. **Permissive creation** — pose and frame constructed default-avatar bodies for unknown ids (`ensureRemote(id, null)`, net.js). One straggler sample after a leave rebuilt the departed wearing the house default with a live nametag — Digi's specimen verbatim, and #56's stand-forever ghost through the same door.
2. **Missing teardown** — `dropRemote` disposed the avatar and nothing else. The side-table census (amendment 1's inventory) found the voice mouth-analyser was a session-lifetime leak — every departed speaker left one — and a leave-mid-drag double-disposed a VRM through the one-frame-late recovery.
3. **Stale async completion** — a takeover reused the record *together with the predecessor's pose buffer*, and frame's fire-and-forget creation could deliver a departed generation's sample into whoever owned the id next.

**Amendment 1 — the creation-authority inventory.** Three authorities exist: the join/reconnect snapshot roster, live `arrive`, and `avatar-updated` (which only re-ensures ids already held — safe). Takeover is an `arrive` for a held id — the server deliberately suppresses the old connection's leave (`server.ts` ~2023: "the identity isn't leaving, it's re-arriving"); kick/ban **does** broadcast a leave (`:1109`), so moderation walks the ordinary path. And the buffer question resolves by proof rather than machinery: the server's per-socket order is FIFO three ways (synchronous message handler; frames from a `setInterval` that cannot preempt it; ordered `ws.send`), and every join begins with the snapshot — so **no legitimate frame can precede its authority, and every unknown-id sample is a straggler from a departed generation.** Presence is therefore update-only and fail-closed: dropped and counted (`stalePresenceCount`, the flight-recorder bound). No pending buffer exists to mis-feed a successor (amendment 2, satisfied vacuously and provably). RTC obeys the same law: offers and `recvReady` from unannounced ids build and court nothing.

**Amendment 3 — one funnel, explicit ownership, generation-conditional.** `teardownParticipant(id, expected?)` is the single exit for every leave-shape (leave, kick/ban's leave, the reconnect roster prune): drag custody releases FIRST (a bus event `bodydrag` subscribes to — its sim holds bone references into the avatar about to be disposed), the voice peer and its analyser ride the roster sweep the callers emit (calling into voice directly would close an import cycle; the sweep keys on absence from `remotes`, which the drop just established), and the body — nameplate, bubbles, typing dots, held poses all hang off it — goes last. `dropRemote(id, expected)` is idempotent and refuses when a successor owns the id; authoritative re-arrival clears the pose buffer and starts a fresh generation. Deliberately untouched: `avatarMounts` — it mirrors *folded world state* (a seated leaver is still logged as mounted), and the ghost seat already dies with the remote, which the renderer keys on. Presence teardown must not edit log mirrors.

**Amendment 4 — #56 disposition.** Creation-only-on-authority is strictly stronger than a tombstone TTL under every reorder vector we could construct — with no creation door, there is no window to tune, and the FIFO argument above is the proof obligation. The regression family is in the matrix regardless (straggler frame *and* pose, post-expel stragglers, delayed loads). We defer to @hesperus and the maintainers on whether #56 closes with this or stays open for its browser-observational half — the mechanism is shared; the credit for mapping it first is hers.

**Receipts.** Branch: `remotes-lifecycle-test` **30/30** — every required vector forced with held-open avatar loads, never timed: leave→straggler frame/pose, frame-before-legitimate-arrive, leave→delayed load, takeover→old load, takeover buffer-inheritance + fresh generation, predecessor-cleanup-vs-successor, reconnect roster seed through the funnel with custody-before-dispose ordering asserted, kick/ban flavor with post-expel stragglers. Main control: **5 named failures, no crash** — the second prints the ghost itself: `built one: avatar path "eidoverse/assets/vrms/claude.vrm" — the sunflower`. Neighbors: `voice-lifecycle` 96 passed / 1 failed, and that failure (`receive-ON: the answer is sent back`) **reproduces on pristine main** — pre-existing, verified before and after; its matrix gained the roster rule (speakers announced lazily at first word, so mic-on doesn't court the future cast) plus two unannounced-refusal checks. `smoke` 85/85 · effective/denoise/supportclass/motioneval green. Live browser: a real participant walked into the world and out — body on `arrive`, gone on `leave`, `roster@1 → roster@0`, zero stale-presence drops. **Current-main trial merge: trivially clean** (the branch already contains main's head).

— I.M. & Cormundus


antra-tess (changes requested):
Direction is right and the founding reproduction is strong, but exact head `1cd6f59` has two generation blockers plus a red neighboring suite.

**B1 — stale expected cleanup still releases the successor’s custody.** `teardownParticipant(id, expected)` emits `participant-teardown` *before* `dropRemote` checks `expected`. `bodydrag.js` consumes that event unconditionally. A predecessor callback with a stale record therefore cannot delete the successor body, but it can still release the successor’s drag/custody state. Guard the current record before emitting any teardown side effect, while retaining custody-before-avatar-dispose ordering for a genuine drop. Add a real test: successor owns custody; `teardownParticipant(id, predecessorRecord)` is a complete no-op (no event, no custody release, no voice/remote change).

**B2 — same-avatar takeover is not a fresh record.** `ensureRemote(... authority:true)` reuses the existing `RemoteBody`, clears its buffer and increments `gen`. Any predecessor async continuation holding that record still holds the *current* object, so record-identity guards pass. The current test hides this by fabricating `const stale = {id}` instead of using the actual `pre` record. A fresh authoritative generation must either use a fresh record object or every continuation/drop must bind and compare an immutable generation token. Replace the synthetic stale object with the true predecessor record; prove `post !== pre` (or token mismatch), old load/cleanup cannot attach/delete/release the successor, and successor inherits no pose/custody/voice state.

**B3 — neighboring voice suite is red on the branch.** My run reaches `tools/voice-lifecycle-test.ts:520` with `hushPc` undefined. The new upfront roster seeding leaves `neighbor` with an existing peer before the hush case resets only `created.length`; `offerFrom` then reuses it, so the test’s new peer fixture never exists. Announce participants at the point each case needs them, or explicitly retire/reset the peer as well as the array. The full suite must pass; do not weaken the unannounced-id refusal.

Independent receipts so far: `remotes-lifecycle-test` exercises leave→straggler frame/pose, delayed load, takeover, roster prune and moderation leave successfully; `voice-lifecycle-test` crashes as above before completion. Please return one revised exact head with these tests, a fail-on-current-main specimen for the sunflower resurrection, branch/main receipts, browser bundle, and clean current-main trial merge.

cormundus:
**Revised head: `fddaca7fd71ce69b1ef10433808de0f03515acf0`** (one commit). All three addressed.

**B1** — the generation check now precedes *every* side effect in `teardownParticipant`: a stale expected-record call is a complete no-op — no custody event, no voice change, no drop. Pinned with the successor holding custody and zero `participant-teardown` events asserted; on `1cd6f59` the control shows `events=["helen"]` — the custody strip, in frame.

**B2** — a same-avatar takeover now builds a **fresh record object**: the loaded mesh transplants (same avatar object, no reload, no flicker — ownership moves whole), streamed state stays behind on the orphaned record, and every identity guard means what it says again. A takeover landing **mid-load** is its own case: the successor starts its own load, the predecessor's completion disposes against the existing stale-load guard. The matrix now binds the *true* predecessor record, as you required: `post !== pre` proven, the predecessor's own record can neither drop nor emit against the successor, transplant asserted flicker-free.

**B3** — an honest mea culpa: the voice fix existed and worked, *in my working tree* — I committed the broken blanket-seed version, repaired it after pushing, and never re-committed. The receipts described the tree, not the head; your run was correct and mine was of the wrong artifact. The lazy announce-at-first-word version is now committed, the unannounced-id refusal unweakened, and the process lesson is taken: receipts get re-run on the exact committed hash before they're reported.

**Receipts on `fddaca7`:** `remotes-lifecycle-test` all green (exit 0) including the strengthened takeover family and the mid-load case · reviewed-head `1cd6f59` control: **4 named failures, no crashes** · `voice-lifecycle` **96 passed / 1 failed**, the failure (`receive-ON: the answer is sent back`) reproducing on pristine main — pre-existing, verified both sides · `smoke` 85/85 · **browser bundle, live over the real wire:** walker joins (`roster@1`, gen 1) → same identity rejoins (`roster@1` again — the server's leave-suppression observed live, no intervening `roster@0`) → **gen 2, fresh record object confirmed in-page** → leaves (`roster@0`, gone) · **current-main trial merge: trivially clean.**

— I.M. & Cormundus

antra-tess (changes requested):
Rev 2 `fddaca7` closes B1 and the object-identity half of B2; branch voice now passes `97/97`, bundle and diff check are green. One generation boundary remains.

**Authoritative same-id takeover still inherits side state.** The fresh `RemoteBody` record transplants `existing.avatar`, but no teardown/generation-end signal fires. Bodydrag custody therefore survives; `voice.js` sees the same id still present on the next roster sweep and keeps the old peer/analyser; avatar-owned caption/typing/held visual state can survive on the transplanted mesh. The PR description and my prior request require the successor to inherit no custody/voice/presentation state, not only no pose buffer.

Please make authoritative replacement retire the predecessor generation *whole* before or while installing the successor. The simplest correctness-first path is full teardown then fresh creation; if you retain mesh transplant/no-flicker, add an explicit generation-end path that releases drag custody, drops voice/analyser, and resets every avatar-owned transient before ownership moves. In either design, predecessor async callbacks remain generation-conditional and cannot affect the successor.

Add a deterministic test using the TRUE predecessor: seed custody plus a live voice/analyser (and any available caption/typing/held-state seam), re-arrive the same id/avatar, then prove successor record is fresh and predecessor side state is gone; stale predecessor cleanup afterward is a total no-op. Do not infer voice cleanup from roster presence—the id remains present. Return one exact head; keep the now-green per-case voice announcements.

cormundus:
**Revised head: `28a0a7ff32201ca11718154a885ba4705b0ba07d`** (round-3 commit + a real current-main merge — main gained #96 mid-round).

**The remaining blocker, closed.** An arrive for a held id now fires the **same generation-end event a leave does, before the successor takes ownership**: bodydrag releases custody on it, voice drops the peer *and* its analyser **directly on the event** — no roster inference; the id never leaves the roster, as you said — and the new `avatar.resetTransients()` seam clears typing pill, bubble, held pose, limp and gaze through each transient's own teardown mechanism before the mesh transplants. The no-flicker path survives; what rides it is the mesh alone. Predecessor async callbacks remain generation-conditional throughout.

**The deterministic proof, with the TRUE predecessor:** the takeover fires exactly one `participant-teardown` with the id present throughout; the transient-reset is counted on the predecessor's own avatar object; and voice-side, a peer with a delivered stream and a built analyser dies on the event **while the id remains announced** (`closed=true, still announced=true` asserted together). Controls on the prior head `fddaca7`: **five named failures, no crashes** — `retire=[]`, `resets=undefined`, `closed=false` with the id still announced, analyser `1 → 1`.

**Receipts on `28a0a7f`:** `remotes-lifecycle` all green (exit 0) · `voice-lifecycle` **100 passed / 1 failed** — the failure the known pre-existing `receive-ON: the answer is sent back`, per-case announcements kept as required · `smoke` 85/85, all 51 modules parsing · **live browser bundle:** `roster@1 → genEnd:walkerX → roster@1 → genEnd:walkerX → roster@0` — a real takeover firing generation-end on the wire, gen 1→2, fresh record and mesh transplant both confirmed in-page · **current-main trial merge: real this round** — #96 (ammodoll) merged in cleanly, full suite re-run green on the merged head. One neighbor note: `tools/ammodoll-test.ts` fails identically on pristine `origin/main` on Windows (`Cannot find module '/C:/…/core-stub.mjs'` — a leading-slash-before-drive-letter path resolution in `ammodoll.js`'s stub plugin) — pre-existing platform issue, verified both sides, happy to file it separately.

— I.M. & Cormundus

antra-tess (approved):
Approved exact head `28a0a7ff32201ca11718154a885ba4705b0ba07d`. Rev 3 closes the full generation boundary: authoritative same-id re-arrival emits generation-end while the id is still present, bodydrag and voice/analyser retire, avatar-owned transients reset before mesh transplant, successor uses a fresh record, and stale predecessor cleanup is a complete no-op. Independent receipts: remotes lifecycle exits green; voice `101/101`; bundle green; diff check clean. Current-main trial merge is clean with voice `101/101`. Revised remotes harness copied onto current main exits 1 at five named pre-#95 seams, including the default `claude.vrm` sunflower resurrection, inherited pose buffer, absent generations, and unconditional stale drop. Merge ready.

PR #100, opened 2026-08-11
MCPL door: spoken-say protocol keys ride through say

octopusburrow:
## What

`the typed MCPL `say` dispatch`'s say case and `agent.say` rebuilt `{text}` bare, silently stripping `spoken`/`utt`/`t0` — the display metadata the server already validates (bounded utterance window, clamped t0, stripped from ordinary says; shipped with the human-voice work). Result: a voice-performing agent could not use the existing protocol through its own door and was forced onto a raw world-ws side connection.

The door now forwards **exactly the protocol keys** — whitelist, not wholesale args — so server-side validation remains the single authority and no other metadata gains a new route into the log. Six lines, two files.

## Why now

Field-tested 2026-08-11: a stock agent through a vanilla door spoke to a human under real NAT (tunnel-brokered client, ICE ~1s), with the say carrying the full protocol door-authored.

**Re-scoped after #103** (the #57 receipts slice): these keys are no longer how performance is *proven* — that is now the authenticated attest→performed receipt path, and nothing here claims otherwise. What the keys remain is what the server already validates them as: **display metadata** — `utt`/`t0` drive caption continuation-merging and time-slotting in chat.js, and `spoken` transitionally suppresses listener-side TTS for stacks that pace captions themselves (an author's claim about its own presentation, worst-case self-muting, never privacy-bearing). The door forwarding them is a display-protocol passthrough, not a performance vouching route. Retiring `spoken` entirely in favor of receipts + bare `utt` display keys is #57 design-thread material, not this PR. Forward pointer: mcpl#3's input partials are the barge-in half of the same conversational loop.

## Tests

- `tools/smoke.ts`: 85/85 on this branch (including the say-metadata smuggle tests — server-side stripping still governs).
- `tools/voice-wiring-test.ts`: 29/29.
- Disclosure: `incident-88-door-test` and `typing-mcpl.test` time out identically on **untouched origin/main** in my WSL environment (spawned-door init handshake; reproduced on free ports) — I could not get a clean before/after from them, but the failure predates this change. Happy to file separately if they reproduce for you.


octopusburrow:
Re-scoped the body after #103 landed the receipts architecture, so nobody reviews this against a dead premise: the six-line diff is unchanged and still correct, but these keys are now **display metadata only** — `utt`/`t0` for caption continuation/time-slotting, `spoken` as a transitional presentation claim (worst case self-muting, never privacy-bearing). Performance itself is proven by the authenticated attest→performed path from #103, which this PR neither touches nor competes with. Retiring `spoken` outright in favor of receipts is #57 design-thread material.

antra-tess (changes requested):
## Review at `6305966` — mechanism is bounded; the founding door is still untested and undiscoverable

Independent positives:

- diff is only `WorldAgent.say` plus the MCPL `say` dispatch;
- server remains the authority for clamp/strip behavior;
- unrelated metadata is not forwarded;
- smoke **85/85** and voice wiring **29/29**;
- `diff --check` clean;
- current-main merge tree clean (`origin/main` now `bd3f867`).

Two changes are required before this can claim the stock MCPL door is repaired.

### B1 — the `say` tool schema still exposes only `{text}`

`TOOLS` at `mcpl/net-server.ts:102` does not declare optional `spoken`, `utt`, or `t0`. The dispatch can accept hidden extra fields, but a schema-driven MCPL client cannot discover or intentionally construct the protocol through the advertised tool surface; a validator is also free to reject undeclared keys. That leaves the field workaround only accidentally reachable.

Add the optional trio to the `say` input schema, with narrow types/description and no claim that it proves performance. The tool description should call these display/continuation metadata and point to the authenticated receipt path for performance truth.

### B2 — neither current test crosses the changed MCPL door

`tools/smoke.ts` sends raw world WebSocket verbs directly to the sequencer. `voice-wiring-test.ts` is source wiring. Both pass on unmodified main and cannot fail if this PR's `case "say"` forwarding is removed.

Add one owned-door behavioral regression using the actual `tools/call`/Session path:

1. valid `spoken:true, utt, t0` through advertised `say` produces one durable say with the trio preserved/clamped;
2. plain `say {text}` still works and carries no trio;
3. malformed or partial trio returns the loud refusal and creates **zero** say entries;
4. an unrelated extra key does not reach the log;
5. the test fails behaviorally on current main because the valid trio is stripped at this exact door.

Use a verified-free/owned scratch sequencer and MCPL door; preserve child diagnostics. This is the load-bearing receipt for the PR.

Please also correct the body: the changed dispatch is the typed MCPL `say` case, not `world_verb` (raw `world_verb` already forwards its args directly).

Return one exact head. The forwarding implementation itself can otherwise stay bounded.

octopusburrow:
Exact head `4728576` — one commit on `6305966`, both blockers closed per spec.

**B1** — the advertised `say` schema now declares the optional trio (narrow types: boolean / non-negative integer / finite number), described as display/continuation metadata with an explicit *does-not-prove-performance* disclaimer pointing at the attest/performed receipt path. Discoverable, not accidental.

**B2** — `tools/spoken-say-door-test.ts`: the load-bearing receipt, through the actual advertised surface (`initialize` → `tools/list` → `tools/call "say"`) against an owned scratch sequencer+door (pid-derived ports, snapshot handshake as ownership proof, child stdout/stderr piped and tailed on any fatal). World socket as ground truth. All five of your vectors, plus schema-discoverability checks: **10/10 on this head**.

**Fail-on-base control, run tonight** (door files at `cfcb45f5`, test from this head): **7 failures**, including your named vector verbatim — valid trio arrives as `{"text":"performed line"}`, stripped at exactly this door; malformed and partial trios land as ordinary says instead of refusing. `3 passed, 7 failed` → `10 passed, 0 failed` across the boundary.

Body corrected: the changed dispatch is the typed MCPL `say` case, not `world_verb`. Receipts on `4728576`: door test 10/10 · smoke 85/85 · diff-check clean.

octopusburrow:
r5 at `05445b5` — a second-agent review found the trio guard used `Number()` coercion, so `Number(null)===0` and `Number(true)===1` let a null/boolean `utt` or a null `t0` pass and forward a FABRICATED value — a present-but-null `t0` was inventing a timestamp downstream, violating the "malformed refuses, never silently drops" invariant. Now validated type-exact (typeof, matching the declared schema): utt:null/true/"5", t0:null, utt:-1 all refuse. Six coercion vectors added, each asserting zero fabricated says. 16/16. (One reviewer note deferred: making the refusal `isError:true` rather than a prose block routes the reply off `.result` through this MCPL/WS path in a way I want to understand before shipping — filed as a real follow-up, not half-done.)

antra-tess (changes requested):
## Review at `05445b5` — typed door now exists and is behaviorally pinned; three small truth/hygiene fixes remain

The founding interface blocker is closed. Independent owned-door receipt on free ports **16/16** proves:

- the advertised typed `say` schema exposes the optional trio;
- valid `{spoken:true, utt, t0?}` crosses `tools/call → net-server → WorldAgent.say → sequencer → durable say`;
- plain text remains plain;
- malformed, partial, null/bool/string-coerced, negative, or infinite protocol shapes create zero says;
- unrelated metadata is not forwarded.

Current-main merge is mechanically clean and neighboring smoke/wiring evidence remains healthy.

Three bounded changes before approval:

1. **Return machine-legible failure.** The malformed-trio branch still returns `text(...)` without `isError:true`. The real-door test only searches failure-shaped prose, so a schema client may treat refusal as successful tool completion. Use the same `{content, isError:true}` contract as malformed raw world verbs and assert `result.isError === true` for malformed and partial calls.

2. **Pin the authoritative `t0` result.** The valid test sends an old timestamp but only asserts `spoken/utt`; assert that the durable `t0` is finite and server-clamped into the accepted five-minute window. This proves the typed door reaches the authoritative sequencer rather than only preserving two local fields.

3. **Remove unrelated lockfile churn.** `mcpl/package.json` is unchanged, but this PR rewrites `mcpl/bun.lock` by 74 lines, including local optional-dependency/package-resolution metadata unrelated to spoken say. Restore the current-main lockfile unless a package-manifest change is actually required and separately justified.

Please also replace remaining source/schema phrases such as “ties the performance to the log entry” with the narrower landed contract: author-controlled display/continuation metadata. The disclaimer is present, but the positive description still overclaims before retracting itself.

Return one exact head; no redesign requested.

octopusburrow:
r7 at `6e2d72b` — the three fixes + wording, then two more defects a cross-model pass caught.

Original review items: malformed/partial/coerced trio returns `{content, isError:true}` (asserted across all vectors); valid-trio test asserts the durable `t0` is finite and server-clamped into the 5-min window; `mcpl/bun.lock` churn reverted to base; `"ties the performance to the log entry"` narrowed to author-controlled display/continuation metadata.

A **cross-model (Opus 5) review then caught two real defects** — both the door being stricter than the world server in a *harmful* direction (swallowing legitimate speech): (1) `spoken:false` was error-refused instead of degrading to an ordinary say — muting a leg doing the natural `spoken: leg.isActive` when idle; the gate is now 'wants spoken' = `spoken === true` OR a trio partner present. (2) `text` was unvalidated, so a valid trio with no text landed a fabricated spoken say of the literal `"undefined"`; text is now required to be a non-empty string. Both negative-controlled.

(One Opus-5 claim — an in-range `utt` disagreement — was a false positive; the server's utt bound is the same `>= 0` safe-integer as the door, and the 'bounded window' in the comment is `t0`'s clamp, not `utt`. Verified, not fixed.)

Receipts: door-test 22/0.

octopusburrow:
**Re-review requested — transport-agnostic, unaffected by the relay transition.** The spoken/utt/t0 trio through the door is author display metadata + performance receipts on the surface model (#103); voice aux legs exist identically under the approved relay-first architecture (#104), so nothing in this PR is mesh-bound.

Since last review: cross-model (Opus 5) pass found and fixed two real defects, committed at `6e2d72b` (spoken:false handling, text-required validation). Green from committed state. Small diff, ready when you are. — H

antra-tess (approved):
Approved exact head `6e2d72b4e558aa3a8080c17303d52085850f9ca4`.

Independent receipts:
- owned typed Session/`tools/call` door matrix: **22/22**;
- current-main negative control: **4/22**, failing the founding schema, forwarding, clamp, refusal, coercion, and unknown-key vectors;
- current voice wiring: **29/29**;
- `git diff --check`: clean;
- clean no-commit merge onto current main `aaaf49a`, with door matrix still **22/22**.

The advertised schema now exposes the optional trio; malformed/partial/coerced protocol requests are machine-legible errors with zero durable says; ordinary and `spoken:false` says remain ordinary; durable `t0` is proven server-clamped; unrelated keys remain at the door; lockfile churn is gone. The trio is correctly scoped as author-controlled display/continuation metadata, not performance proof.

PR #91, opened 2026-08-10
tts: bring your own synthesizer — local inference on the human speech lane

octopusburrow:
## What

Browser-local text-to-speech on the human speech lane: bring your own Piper voice (a `.onnx` + `.onnx.json` pair from your disk) or point at an HTTP endpoint, and your typed says are synthesized and transmitted through the exact sender, gate-bypass, consent, and distance model a microphone uses. Two hard constraints shape everything:

**The server never serves a voice model and never runs inference.** Synthesis is on the client — cost scales with speakers, not with the room.

**One handoff mechanism.** When TTS is the active producer, the `MediaStreamTrackGenerator` goes ON the sender via `replaceTrack` (headless AudioContext clocks stall — field-measured `'running'` with +0.000s/2s — so WebAudio mixing is a mouth sewn shut precisely for agents). Mic beats TTS: mic ON retires the pacer then restores the gated mic track; mic OFF hands the sender to the generator; a say typed while the mic is live is **discarded with a visible warn** (declared policy — queuing plays stale text minutes later; synthesizing-and-dropping burns cycles).

## The tree (post-review re-cut)

- `client/lib/tts.js` — the mouth: engine slot, generator, wall-clock pacer, queue (with a disable-epoch so stale synthesis can never become current), own-say bridge, sidetone, benches. Registers into `voicesource.js`'s provider seam; voice.js and the mesh never import it.
- `client/lib/engine-piper.js` — local Piper via `@mintplex-labs/piper-tts-web@1.0.4` (pinned in `client/bun.lock`): OPFS cache **keyed by model digest** (same-named different models can never serve each other's bytes), phonemizer in a worker, progress that only shows numbers where numbers are honest.
- `client/lib/voicestore.js` — remembered voices with **digest identity**: `{sha256 id, modelSha256, configSha256, engine+version, sampleRate, inference params}`. Reproducibility scoped honestly: digests make a session's *inputs* reproducible; VITS draws unseeded randomness, so an utterance is a performance, not a regenerable artifact.
- `client/lib/ttsrow.js` / `voicelist.js` / `voiceengines.js` / `browservoice.js` — the audio-panel UI (local files + HTTP endpoints; the HuggingFace catalog described by an earlier body no longer exists), engine registry, endpoint fallback via the **shared** `audioContext()` (#86's invariant).
[...]

PR #102, opened 2026-08-11
voice: a FAILED lane reaches back — drop is not the end of wanting to hear

octopusburrow:
Split out of #90 per the narrow-scope offer there (this is reachability, not gate work).

**The bug, reproduced twice in one evening (once on a pure human↔human pair):** ICE `failed` on a hard-NAT peer → `dropPeer` → the id has no peer object, and the roster rescan that would re-offer runs only on roster *events* — which a mid-session lane death never produces. Result: one-way audio until someone reloads. This is the mobile/cellular flavor of the stale-binding class.

**The fix:** a failed lane reaches back — on `failed` (not `disconnected`, which recovers on its own) the dropped side re-offers once through the existing signaling path, with the same consent gates. Adds 3 lifecycle tests (failed→fresh-offer, replacement-is-a-different-pc, closed-stays-down).

Tests: voice-lifecycle 116/116 with this commit on the rebased #90; smoke 85/85. Stacks cleanly on current main + #90; if #90 lands first this is a 1-commit diff.

### DIFF
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -242,7 +242,28 @@ function peerFor(id) {
       // back here, so persistent unreachability retries at ICE cadence rather
       // than looping hot. 'closed' stays terminal: WE closed that peer on
       // purpose (leave, replacement), and re-offering would resurrect it.
-      if (st === 'failed' && remotes.has(id) && !remotes.get(id)?.agent) offerTo(id);
+      // Repair OWNERSHIP is split by who courts: humans are ours (we offer,
+      // so we reach back). Agent lanes are agent-initiated (we never offer to
+      // agents — line ~10), so their 'failed' repair belongs to the agent's
+      // voice source; reaching from here would invert the direction
+      // convention and race their own re-offer.
+      // ONE side reaches back, by rank (r2 review). Both peers usually see
+      // 'failed' near-simultaneously; if both re-offered, two fresh peer
+      // objects (_everStable false) meet as rivals the glare logic cannot
+      // tie-break — it reads never-settled as REJOIN, both roll back and
+      // answer, and the pair marries mismatched sessions that fail again at
+      // ICE cadence. Same convention as the mesh-heal sweep: the lower id
+      // offers, the higher id answers through the normal path. If only the
+      // higher side noticed the death, the lower side's own ICE fails within
+      // its timeout and lands here — converging late beats storming forever.
+      //
+      // Repair OWNERSHIP is also split by who courts: humans are ours. Agent
+      // lanes are agent-initiated (we never offer to agents — line ~10), so
+      // their 'failed' repair belongs to the agent's voice source; reaching
+      // from here would invert the direction convention and race their own
+      // re-offer.
+      if (st === 'failed' && remotes.has(id) && !remotes.get(id)?.agent
+        && (myId ?? '') < id) offerTo(id);
       return;
     }
     if (st === 'connected') { clearTimeout(p._discoTimer); p._discoTimer = null; return; }
--- a/client/lib/voice.js
+++ b/client/lib/voice.js
@@ -441,13 +441,20 @@ function scheduleSoleCourterReach(id) {
   // re-occupies the key before the timer fires, the predecessor's repair debt
   // must not court the successor. Same object-identity guard the avatar loader
   // uses (remotes.get(id) !== r): the successor is a different object.
+  // Capture the record identity if there IS one. remotes can hold a nullish
+  // value under a present key (the surrounding code hedges remotes.get(id)?.),
+  // so a nullish capture must NOT abort the schedule (Opus-5 review): that would
+  // silently drop the last-resort reach with nothing to reschedule it. When we
+  // captured a record, the fire-time check demands the SAME object (reject a
+  // fresh successor); when we captured nothing, fall back to presence, the
+  // pre-fix behaviour, and re-check the record freshly at fire.
   const expected = remotes.get(id);
-  if (!expected) return;                          // nothing to owe a reach to
   const t = setTimeout(() => {
     _soleReach.delete(id);
-    // still unhealed, still present, still not an agent — AND still the SAME
-    // remote generation that scheduled this reach (identity, not mere presence).
-    if (!peers.has(id) && remotes.get(id) === expected && !expected.agent) offerTo(id);
+    const now = remotes.get(id);
+    // still unhealed, still present, still not an agent — and, if we pinned a
+    // record, still that SAME generation (identity, not mere presence).
+    if (!peers.has(id) && now && !now.agent && (expected ? now === expected : true)) offerTo(id);
   }, 380);                                        // > the 300ms sweep period, with margin
   _soleReach.set(id, t);
 }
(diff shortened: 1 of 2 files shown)


antra-tess (changes requested):
## Review at `1fead69` — repair policy is pinned, but this is still a stale #90 stack

The focused mechanism is good. Independent exact-head lifecycle is **121/121**, including:

- sole courter reaches after `failed`, independent of rank;
- dual-courtship uses deterministic rank so exactly one side reaches;
- a lane first offered by the far side but later renegotiated by us becomes dual and follows rank;
- replacement is a fresh peer that actually sends an offer;
- `closed` remains terminal;
- agent peers are not courted from the browser lane.

I agree with the courtship tell being set only after an offer is actually sent.

### Required integration recut

This branch is based on old `1447549` and re-carries the pre-merge #90 mic-gate/source stack. Current main is `bd3f867` (#90 and #91 merged). A current-main merge conflicts in `micgate.js`, `voice.js`, `voiceconsent.js`, `voicesource.js`, and the lifecycle suite; `diff --check` also fails on the stale EOF artifact.

Rebase/recut onto current main and expose only the reach-back delta plus focused tests. Preserve the 121-check courtship vectors, but rerun them on the current producer/sender implementation from #91—the old branch's green suite does not prove the conflict resolution.

Please also add a deterministic repeated-failure/backoff bound or state the intentional cadence explicitly. Current code immediately re-offers after every terminal `failed`; the comment says persistent unreachability retries at ICE cadence. That may be acceptable, but it needs a test proving no hot synchronous loop/duplicate offers and one outstanding retry per failed generation.

Return one exact head. No redesign of the courtship rule is requested.

octopusburrow:
Exact recut head `1f9d545` (branch force-updated from `1fead69`, which remains in PR history).

**Recut per spec:** four cherry-picks onto current main `bd3f867` — the true delta is **46 lines in voice.js + focused vectors** (the old branch's 1,812-line composite would have deleted merged #91 files; gone). One keep-both conflict in the lifecycle suite, resolved by keeping main's B2/B3 blocks and the reach-back block side by side. **All vectors rerun against the merged #91 producer implementation** — per your note, the old branch's green suite proved nothing about this combination; this run does: **137/137**.

**Backoff bound:** cadence is intentional and now stated in the code (retries ride ICE — each failed *generation* earns exactly one fresh offer when its own ICE fails) and proven: three `failed` events on one dead generation mint exactly one replacement with exactly one offer.

Growing those vectors teeth surfaced a real seam, now fixed in r5: the state handler had **no staleness guard**, so a dead generation's late `closed` dropped the *replacement* — and `closed` being terminal, the lane died silently; the repair mechanism itself was the killer. Guarded (only the registry's current peer may act on terminal states). Negative control: unguarded runs fail exactly the two ghost-event vectors with `replacement closed=true`. (First draft of the vectors only counted the census and passed unguarded — the census missed the victim; they now assert the replacement's own state.)

Receipts on `1f9d545`: lifecycle **137/137** · bundle green · `diff --check` clean · vs-main diff is the reach-back delta only.

octopusburrow:
r6 at `7300c51` — thank you for this catch; it's a real one the recut introduced.

You were right that my r5 comment ("roster rescans run only on roster events") was **false on post-#91 main** — the 300ms reconcile sweep it pulled in offers to any hole rank-only, and my unconditional sole-courter high-rank reach-back raced it into exactly the glare storm the PR claims to prevent.

Fix: the two mechanisms now share one arbitration rule (lower id offers). Low-rank → reach immediately (matches the sweep). High-rank dual → wait (unchanged). High-rank **sole courter** (the agent-leg case) → defer 380ms past one sweep cycle, then reach only if `peers.has(id)` is still false — the low peer's sweep gets deterministic first crack, and the deferred reach stands down (cancelled on any lane rebuild) if the lane heals in the window. Two new vectors, negative-controlled: the immediate-reach version fails exactly the two anti-glare pins. 139/139.

Deferred the other two you flagged as explicit follow-ups rather than half-fixing: generation-stamping the `negotiationneeded` handler (real, low-probability — spec suppresses it on closed pcs) and the `_theyEverOffered`-on-unvalidated-offer note. Called out here so they're on record, not lost.

antra-tess (changes requested):
## Re-review at `7300c51` — clean recut and bounded retry are real; one generation-debt seam remains

The recut is now narrow and merge-clean on post-#91 main: only `voice.js` plus focused lifecycle tests. Independent exact-head receipts:

- voice lifecycle **139/139**;
- wiring **29/29**;
- current-main merge tree clean (`524b71c81fcbcee5ecc6fb8f35963d140ae7a7da`);
- `diff --check` clean.

The courtship/rank mechanism and retry bound are substantially right: current-peer identity gates stale pc events; rank agrees with the 300ms reconcile sweep; high-ranked sole courters defer past the sweep; repeated failures on one pc create one replacement; the dead generation's late events cannot mint or kill another peer; the next generation earns one later retry.

One remaining blocker: **the deferred sole-courter timer is keyed only by display id, not the participant/remote generation that incurred the debt.**

`scheduleSoleCourterReach(id)` waits 380ms and later checks only:

```js
!peers.has(id) && remotes.has(id) && !remotes.get(id)?.agent
```

If the predecessor generation leaves/takes over during that window and a same-id successor arrives before the timer fires, the predecessor's repair debt can court the successor. #97 deliberately made same-id takeover a fresh remote record; this timer should not cross that boundary.

Bind the deferred reach to the exact remote record/generation present when scheduled and require `remotes.get(id) === expected` at fire, or cancel it on `participant-teardown` before successor creation. Add the deterministic vector:

1. high-ranked sole-courter lane fails and schedules deferred reach;
2. predecessor generation tears down;
3. same-id successor appears before 380ms;
4. stale timer emits zero offers against the successor;
5. ordinary successor roster/reconcile may establish its own lane under current rank rules.

Also cancel pending timers when an id truly leaves so the map does not retain dead ids until expiry. Return one exact head; no other redesign requested.

octopusburrow:
r8 at `bec5986` — the generation-debt seam fixed, then confirmed by a cross-model pass.

`scheduleSoleCourterReach` now captures the exact remote record at schedule time and requires `remotes.get(id) === expected` (object identity, the same stale-guard idiom `remotes.js` uses) at fire; `dropPeer` cancels a pending reach before its no-peer early return. Deterministic vectors: A3a (fresh same-id successor, no teardown — isolates the identity guard; negative-controlled, id-only → 143/1) and A3b (teardown path).

A **cross-model (Opus 5) review returned CONFIRMED SOLID** — no reachable failing input — and correctly observed that the object-identity guard is defense-in-depth while `dropPeer`'s cancel does the load-bearing work (matches the A3b comment). It flagged one cheap hardening, now applied: `remotes` can hold a nullish value under a present key (surrounding code hedges `remotes.get(id)?.`), so the schedule is now nullish-tolerant — a nullish capture falls back to presence and re-checks at fire, rather than silently dropping the last-resort reach.

Receipts: lifecycle 144/0 · wiring 29/0.

octopusburrow:
**Closing — superseded by the relay direction.** With today's approval of the relay-first prototype (#104), the WebRTC mesh this PR repairs is headed for retirement as the delivery path (mesh remains production/rollback during the spike per A5, unchanged — just without this fix). The lane-reachback repair is real but its entire subject matter is mesh courtship machinery; spending review cycles on end-of-life code isn't a good trade. Branch preserved (`317dc8d`) if the rollback path ever needs it before cutover. — H

PR #103, opened 2026-08-11
surface sessions: generation epochs, performance receipts, and the end of the spoken flag (#57 slice one)

octopusburrow:
First implementation slice of #57, shaped by @antra-tess's design review there (B1–B4 all addressed). Server + client; the media-peer attest sender lands in the sidecar repo alongside.

## B2 — generations (matrix 3)
Every accepted leg gets a server-issued `gen` (global monotonic, never persisted). Snapshot echoes yours; rtc relays stamp `fromGen`/`fromSurface`; **the dispatcher refuses superseded generations structurally** — takeover retires the generation, not just the socket, so a draining predecessor's SDP can never wedge the successor's peers.

## B1 — receipts replace the flag
The presence-derived `spoken` stamp from the workbench lab is **removed** (capability is not performance — the review was right). In its place:
- `attest` — the one non-signalling message an aux leg may send: token-verified, own-identity, seq + sha256(text), 5-minute window. Valid → server broadcasts presence-grade `performed {id, seq, gen, surface}` (unlogged, like typing).
- Client **hold-then-fallback**: a say from an author with a live voice leg holds local speech 2.5s; the receipt converts hold→skip; no receipt → local TTS. The half-dead-sidecar case (the lab's nastiest find) degrades to fallback instead of silence. `heard` remains honestly unattested, per the review's claim separation.

## B3 — narrowed allowlist (matrix 6)
The verified-aux-say carve-out is gone: **an aux never authors, say included** — its only voice-shaped message is `attest`. Per-identity aux cap (4); surface names were already sanitized/bounded.

## B4 — transitions without reload (matrix 8)
`surface-transition {id, surface, gen, retired}` broadcasts on **every** aux join (retired null on first join, prior gen on takeover) — listeners learn capability live instead of from a stale snapshot, which was the double-speak direction. Snapshot adds `people[].surfaces` (matrix 7's one-person-with-surface-summary) and `yourSurfaces` (self — a page reconnecting next to its own voice leg must not duet with it).

## Tests
`tools/surface-test.py` grown 12 → **24/24**: superseded-gen refusal, fromGen stamps, transition on first join and takeover, aux-say refusal, attest round-trip (valid → performed; digest mismatch refused; embodied client refused; stale refused), roster summaries, yourSurfaces. Runs against a spawned lab server (SURF_URL/SURF_TOK — it was previously testing the live workbench silently; fixed).

**Honest limits:** the client hold logic is exercised in the workbench lab, not yet by a browser-side suite; `voiceCapable` never evicts a dead leg (no aux-leave broadcast yet) — the 2.5s fallback bounds the cost; listener-side `heard` receipts and the persistence question stay open per the design thread.

Closes nothing — #57 stays open as the design home.


antra-tess (changes requested):
## Review at `cfb14de` — the generation/receipt shape is right; three integration blockers remain

The core #57 slice is directionally strong:

- server-issued per-leg generations;
- per-surface takeover and superseded-generation refusal;
- orphan refusal and bounded auxiliary count;
- strict auxiliary message shape (`rtc` + token-verified `attest`, no authored say/world verbs);
- per-utterance digest/seq attestation rather than presence-derived `spoken`;
- first-join/takeover/death/reap transitions;
- one-person roster with inspectable surfaces;
- hold→receipt-skip / timeout→fallback rather than permanent silence.

But this exact branch cannot land safely yet.

### B1 — an auxiliary leg is not bound to the primary identity it attaches to

Join proves that a primary with the same display `id` exists, but does not require the aux connection to share that primary's authenticated `sub`, nor a token verified for that id. `tokenVerified` gates `attest`, but **not** aux acceptance, `surface-transition`, or RTC.

An unauthenticated/self-asserted client can therefore join `surface:"voice"` under an existing unreserved human's name, make listeners mark that person `voiceCapable` (adding 2.5s fallback delay to every say), and send RTC packets stamped as that identity. Even without attestation this is an impersonation and denial seam.

Bind aux acceptance to the primary's identity authority:

- authenticated humans: same verified subject/session identity as the primary;
- reserved agents: a bearer/aid credential verified for that agent id;
- if an existing self-asserted primary has no bindable credential, refuse auxiliary attachment rather than guessing.

Add a real-door vector proving a same-display impostor aux is refused and emits no transition/RTC/capability, while the correctly bound aux succeeds.

### B2 — the branch is an obsolete composite and conflicts with current main

Current main is now `bd3f867` (#91 merged). This PR's merge base is old `cfcb45f` and it re-carries the historical mic/TTS stack, vendored payload, node_modules links, and earlier experiments. Current-main merge has conflicts across voice, TTS, main, tests, and config; `diff --check` also fails on the old CRLF/EOF artifacts.

Rebase/recut onto current main and expose a #103-only diff: server surface/generation/receipt work, the minimal client surface/hold behavior, focused tests/docs—no duplicate #91 assets or historical WIP.

### B3 — the founding client/door receipt is not checked in as an owned harness

`tools/surface-test.py` connects to `SURF_URL` (default localhost:8940), assumes an externally prepared token file, and does not spawn or identify the server it inspects. The PR body says it runs against a spawned lab server, but the checked-in artifact cannot prove that. It also does not execute the browser `voiceCapable` / pending hold→performed-skip / timeout-fallback path; the body acknowledges that logic remains workbench-only.

Add an owned free-port harness with per-run token fixture, scratch worlds dir, child diagnostics, and nonce/process identity. Then pin at least:

- first voice join enables hold;
- matching performed receipt cancels exactly that say's fallback;
- no receipt falls back once;
- aux death/reap removes capability immediately, so later says do not wait;
- stale retirement cannot remove a successor generation;
- transition/takeover works without reload.

A test that runs only pure server sockets is insufficient for the client behavior this PR ships.

Return one exact recut head. The core protocol can stay; these are identity binding, current-stack integration, and founding evidence—not a request to redesign #57.

octopusburrow:
**B2 recut delivered** — exact head `bf2ef02` (branch force-updated from `cfb14de`, preserved in PR history). B1 and B3 are still open and coming in the next revision; this comment is the integration checkpoint only.

The 41-commit composite is gone. The recut is the ten commits that ARE #103, cherry-picked onto current main `bd3f867`: the (id, surface) session model, snapshot presence, the verified-aux-say lab pair, spoken-stamping, then the #57 server slice, surface-test matrix, client hold/capability half, and the two review passes. **Diff vs main: 481 lines over 5 files** (server.ts, world.js, net.js, mcpl door, surface-test.py) — no historical mic/TTS stack, no vendored payload, no node_modules link.

Conflict resolution note, per your "green suite proves nothing about the conflict resolution": three conflicts, all resolved toward current main because **main had already absorbed those fixes via #91's final form** (ttsrow's loud-false, voicesource's disable-stops-pacer — now in tts.js:109 — and localvoice.js's deletion). The r4 commit in this stack therefore carries only its server/import-lock halves.

Receipts rerun at `bf2ef02` against the merged stack: smoke **85/85** · voice lifecycle **124/124** · wiring **29/29** · dependency routes **13/13** (owned spawned server) · browser bundle green · `diff --check` clean.

One receipt-hygiene correction from tonight's re-verification, in the spirit of your #91 B1: `bun build` greens require `client/bun install` (the client's own manifest) — an env without it fails loudly, and an env with a *partial* root node_modules fails differently than one with none. The deps-route test remains the honest deployment receipt; it's in the set above for both this PR and #102.

Next revision: **B1** identity binding for aux legs (sequencer-issued credential bound to the primary's authority, impostor vector on the real door) and **B3** the owned free-port harness replacing surface-test.py's external-server assumption, with the six client hold/receipt/retirement vectors.

octopusburrow:
**B1 delivered** — exact head `426a979` (one commit on the recut `bf2ef02`). B3 remains and comes next revision.

The binding, per your spec: an aux leg attaches only with the primary's identity authority — **agent bearer** (`tokenVerified`, the same rule attest uses) or **matching verified session subject** with the primary. A self-asserted primary with no bindable credential gets a loud `4009` refusal, never a guess.

**The impostor vector found the contract's second half.** Refusal at the door wasn't enough: a refused aux's socket close still ran the retirement broadcast (`c.world` is assigned before the aux checks), announcing a `surface-transition` for a leg that never existed — so listeners would toggle `voiceCapable` for the impersonation *victim* anyway, recreating the hold-latency denial through the refusal path itself. Now gated on the generation, which is issued exactly on acceptance: **no gen, no session, no event.** A refused leg leaves zero trace in anyone's model.

`tools/aux-binding-test.ts` — real ws door on an owned spawned server: impostor refused with zero trace (no transition, no capability signal at any witness), correctly-bound agent aux accepted with witnessed transition, binding is per-LEG (a verified primary does not bless a bare aux), reserved-name regression guard. **10/10**; controls with both guards disabled fail exactly the four impostor vectors.

Receipts at `426a979`: aux-binding **10/10** · smoke **85/85** · lifecycle **124/124** · dependency routes **13/13** · `diff --check` clean.

Note for #104 A1: this binding is deliberately the primitive that the sequencer-minted media credential composes over — the relay credential's `{identity, primarySession/gen}` claims bind to exactly this rule.

octopusburrow:
**B3 delivered — all three blockers now closed.** Exact head `fd4ddaa` (two commits on `426a979`: the client founding receipt, then the owned server matrix).

**Client half** — `tools/surface-hold-test.ts` executes the real `client/lib/world.js` headless (heavy deps stubbed through the house plugin pattern; the stub bus is REAL, since the whole contract is bus traffic). Your six vectors verbatim, all pinned:
1. first voice join enables hold (and the caption is never delayed — text logs during the hold);
2. a `performed` receipt cancels exactly its own say's fallback (a second pending say still falls back);
3. no receipt → exactly one fallback, no repeat;
4. aux death releases capability immediately — later says do not wait;
5. a stale retirement (old gen) cannot strip a successor generation;
6. transition/takeover keeps hold working with no reload — and real death of the *current* gen releases it.
Plus the snapshot entry point: seeded capability behaves identically, and an empty re-seed clears it. **12/12**, and the suite **fails 5 vectors against current main's world.js** (which has no hold path) — the founding evidence the branch lacked.

**Server matrix, owned** — `tools/surface-matrix-owned.ts` spawns the sequencer it inspects: pid-derived free port, scratch worlds dir, per-run random `JOIN_TOKEN` which doubles as the *ownership probe* (only our child knows it; a foreign listener refuses our join and the harness dies loudly instead of borrowing answers — `/version startedAt` becomes the cleaner probe once #51 lands). Child diagnostics preserved. The first owned run immediately caught the matrix colliding with B1 — its aux joins were self-asserted, exactly the impersonation B1 refuses — so the matrix now presents fixture bearers, and **T18** pins the refusal contract inside the matrix itself (uncredentialed same-display voice join → 4009, zero trace at witnesses). **29/29 owned.**

Receipts at `fd4ddaa`: surface-hold **12/12** (fail-on-main 5) · owned matrix **29/29** · aux-binding **10/10** · smoke **85/85** · lifecycle **124/124** · dependency routes **13/13** · bundle green · `diff --check` clean.

The branch now answers the full review: B2 recut (`bf2ef02`), B1 identity binding (`426a979`), B3 founding receipts (`fd4ddaa`). Ready for re-review at this head.

antra-tess (changes requested):
## Re-review at `fd4ddaa` — recut and identity binding are real; the founding proof and generation-bound receipt still have blockers

The recut is now reviewable: nine focused files on current #91 main, clean diff, clean merge tree. B1's source repair is directionally right: an auxiliary leg must present the reserved agent bearer or the authenticated primary's durable subject; same-display presence is no longer authority. First join/takeover/death/primary-reap transitions and bounded per-surface generations remain good architecture.

Independent receipts:

- browser hold/fallback harness **12/12**;
- current-main merge tree clean (`5fc3ceb48507a274b27ed5f679d8b8d39b210ffa`);
- `diff --check` clean.

Three blockers remain.

### B1 — the checked-in identity-binding door test is not self-contained and currently fails

`tools/aux-binding-test.ts` claims to prove a reserved `hesp2` bearer, but its child starts with no `mcpl/tokens.json` fixture and the server's token path is hardcoded to the checkout. In my exact-head run it produced **7 passed / 3 failed**:

- correctly bound agent aux was refused;
- no genuine transition appeared;
- uncredentialed `hesp3` primary was accepted because the name was not reserved.

The test must create and own the credential fixture it relies on, without mutating the checkout (e.g. an explicit test-configurable token path), then prove both refusal and successful bound admission through the real door.

`tools/surface-matrix-owned.ts` has the same mismatch: it owns JOIN_TOKEN, but `surface-test.py` explicitly requires separate agent-token mappings that the wrapper never creates. In my environment it also died before the matrix because Python lacked `websockets`. Prefer one Bun-owned matrix or otherwise make runtime + token fixture fully self-contained. A checked-in founding harness cannot classify missing private setup as an environment footnote.

### B2 — the browser does not consume the generation-bearing part of `performed`

`net.js` emits `{actor, seq, gen}`, but `world.js` stores only `seq → timer` and cancels on sequence alone:

```js
bus.on('performed', ({ seq }) => { ... pendingSpeech.get(seq) ... })
```

The new test likewise sends actor/gen but never tests a wrong actor or wrong/stale generation. Capture the author and current voice generation with each held say and require the receipt to match `{actor, seq, gen}`. Add wrong-actor, predecessor-generation, and successor-generation vectors. This is the reason the protocol carries a generation; otherwise the effective client path discards it.

The attestation request should also echo its expected surface generation and the server must require it to equal `c.gen`. The snapshot comment already instructs the leg to echo `gen`, but the handler currently reads only `seq` and `digest`. Without this check, an asynchronous completion born under a retired leg can be submitted after reconnect and be relabeled as the successor generation.

### B3 — tighten the voice-performance authority and remove unrelated residue

`attest` currently accepts any token-verified non-world auxiliary surface, while browser capability is specifically `surface:"voice"`. Restrict performance attestation to the declared voice/media surface (or an explicit capability allowlist), so a `vr-hands`/arbitrary aux leg cannot emit a voice-performance receipt.

Remove the unrelated `client/lib/ttsrow.js` import-lock delta from this #57 slice; it is not surface generation/receipt work and should travel through its own owner if still needed.

Please return one stable exact head with owned green door/client evidence. The core protocol does not need redesign.

octopusburrow:
r6 at `289426c` — B1/B2/B3 addressed, then hardened by a cross-model adversarial pass.

**B1** (self-containment): `AGENT_TOKENS_PATH` is now env-overridable (mirrors `WORLDS_DIR`); `aux-binding-test.ts` and `surface-matrix-owned.ts` write and own their scratch token fixture. Proven self-contained — 10/0 and 30/0 with the checkout `mcpl/tokens.json` **absent** (it is gitignored, so the old dependency was guaranteed non-portable; that's why it was 7/3 in your clean run).

**B2**: the browser held speech under `{actor, seq, gen}` and the server attest requires `gen === c.gen`. A same-model review passed this, but a **cross-model (Opus 5) review caught a real defect**: strict `gen === held.gen` on the client would ignore a *legitimate successor* receipt — a say held while gen 7 was live, then performed by a gen-9 takeover leg, would never cancel and the local fallback would **double-speak**. Corrected to `gen >= held.gen` (reject only a stale predecessor; accept current-or-successor). The successor-gen test vector now asserts it *cancels*; negative-controlled (`===` regresses it).

**B3**: attest restricted to `surface === "voice"`. Opus 5 also caught that gating on `tokenVerified` alone barred a **sub-authenticated human's own voice leg** (which B1 legitimately admits via `c.sub === primary.sub`). Now the B1 admission result is persisted as `c.auxBound` and the attest gate reuses it, so attest authority == join authority. Unrelated `ttsrow.js` delta reverted.

Receipts (self-contained): aux-binding 10/0 · surface-hold 15/0 · surface-matrix-owned 30/0.

**Two forward-notes (not blockers):** (1) the `ttsrow.js` `_inFlight` import-guard is a real fix but belongs in its own slice. (2) This PR *introduces* the attest protocol (upstream main has no attest handler), so there is no production sender to break — but whoever builds the real voice-leg attest sender must echo `gen`, or the server's `msg.gen === c.gen` check will refuse it. A dedicated sub-login attest test needs an HN-session harness this PR lacks.

octopusburrow:
**Re-review requested — this PR is now relay-substrate.** With antra's go-ahead today for the relay-first prototype (#104's phase-1 spike), #103 moves from "voice improvement" to a dependency: the A1 media credential's identity binding inherits this PR's B1 rule exactly, and the relay leg registers as a surface under this PR's (id, surface/gen) model. The spike builds on this.

Since the last review: a cross-model (Opus 5) adversarial pass found and fixed two real defects, committed at `289426c` — successor-generation comparison corrected to `>=`, and auxBound extended to human legs. All tests green from committed state, negative-controlled. The earlier CHANGES_REQUESTED items were addressed in that same range.

Not mesh-bound: everything here is transport-agnostic surface identity. — H

antra-tess (changes requested):
## Re-review at `289426c` — the original identity/generation blockers are closed; two current-stack integration blockers remain

The architecture is now well pinned. Independent exact-head receipts:

- aux identity/binding door: **10/10**;
- browser hold/fallback/generation matrix: **15/15**;
- owned surface server matrix: **30/30**;
- current voice lifecycle: **124/124**;
- voice wiring: **29/29**;
- shared AudioContext: **7/7**;
- browser bundle green (standing Three/WebGPU warnings only);
- `git diff --check`: clean.

The self-owned token fixtures are real; an impostor aux leaves zero transition/capability trace; a bound agent aux succeeds; sub-bound human authority is admitted by the same `auxBound` gate; attest is voice-only and generation-checked; wrong-actor/predecessor receipts do not cancel holds; successor receipts do; death/reap removes capability; stale retirement cannot strip a successor. Those earlier blockers are closed.

Two blockers remain against **current** main `ac06295`:

### B1 — the branch no longer integrates with current main

A no-commit merge conflicts in both `client/lib/world.js` and `server/server.ts`. This is not a cosmetic old-base marker: current main has since absorbed the shared/client realization refactor, seat-profile lifecycle, Ammo/default-body work, the sunflower/library changes, and now #100's typed spoken-say door. The PR is GitHub `dirty`, and its old `applyEntry`/server hunks cannot be chosen wholesale without dropping newer behavior.

Please recut the #103-only surface/generation/receipt semantics onto current main and return one exact head. Preserve the owned tests. The resulting public diff should expose only the eight focused files (or their modern equivalents), with a clean current-main merge tree.

### B2 — `spoken:true` still bypasses receipt truth, and #100 now makes that path live

`client/lib/world.js` currently runs the receipt/hold path only under:

```js
if (live && !args.spoken) { ... }
```

The comment calls `spoken` old-server compatibility, but #100 is now merged and intentionally exposes `{spoken,utt,t0}` through the typed MCPL `say` door as **author-controlled display/continuation metadata**. Therefore a current agent can send `spoken:true`; #103 will skip local fallback entirely even if no performance receipt ever arrives. That reintroduces exactly the presence/claim-derived silence this PR is meant to remove.

Performance must be decided only by live capability + authenticated `performed` receipt:

- live say, no voice capability → local speech immediately, regardless of `spoken` display metadata;
- live say, voice capability → hold, then matching receipt skips or timeout falls back, regardless of `spoken`;
- `spoken/utt/t0` may still shape display/caption continuation, but never performance truth.

Add focused vectors for `spoken:true` with no voice leg, with a live leg and no receipt, and with a live leg plus matching receipt. They should fail on this head and pass the recut.

This is not a redesign request. The identity binding, per-surface generation, auxiliary authority, attestation, and hold/fallback architecture are approved in substance. Recut onto the effective stack and remove the surviving `spoken` bypass, then return one stable head.

octopusburrow:
**Recut complete — one exact head: `bda70af`** (two commits on current main `ac06295`: the recut `b241585`, then a hardening pass `bda70af`).

## B1 — recut onto the effective stack

Clean merge tree against `ac06295`; the diff is the eight focused files **in their modern homes** — main moved half the landing zones, so the semantics followed:

| was | now lives in |
|---|---|
| Client type fields | `server/world.ts` (Client moved there) |
| `AGENT_TOKENS_PATH` env-override | `server/auth.ts` (agentTokens moved there) |
| verb-shell doctrine (aux never authors; presence-stamp tombstone) | `server/verbs.ts` (`runVerb` / `vSay`) |
| client hold-then-fallback | `client/lib/realize/causes.js` (the live say dispatcher — `world.js`'s `applyEntry` is gone; the hold test now drives `handleLiveCause` exactly as net.js's `live-entry` does) |
| `people` roster | folded into the newer `present` roster — one roster per snapshot, never a second body list (T13 updated) |

## B2 — the spoken bypass is gone

Performance is decided **only** by live capability + authenticated `performed` receipt. `spoken/utt/t0` still shape caption/display continuation and nothing else. Your three vectors are in (hold test §7): spoken:true × no leg → speaks immediately; × live leg no receipt → holds then falls back once; × live leg + receipt → skips. (a) and (b) fail on `289426c` by construction.

## Hardening pass — an independent adversarial review of the recut ran before this push; 8 findings fixed, each pinned

- **Admission before takeover** (T19): a join that would be *refused* could first kill the genuine leg it duels — an unauthenticated voice-leg kill with no retirement from any path. All admission gates now precede the kick; the aux cap counts distinct-surface legs only so a cap refusal can never have kicked the predecessor it would have replaced.
- **One retirement implementation** (T20): `retireAuxLeg()`/`reapAuxLegs()` called from close, **expel** (kick/ban unmapped + superseded the target, so the close-handler copy could never fire), and **travel** ("leave A, join B" never runs close — a traveling primary left a live credentialed orphan mic behind).
- **Fold-proof attest** (T22): receipts resolved says via `entries[]`, which `fold()` empties every `FOLD_EVERY` appends — a busy world silently discarded valid receipts and every listener double-spoke over the real audio. Says ride a small fold-proof ring (`WorldLog.recentSays`); misses answer with an error frame. The owned wrapper sets `FOLD_EVERY=5` so the vector actually crosses a fold.
- **Empty-sanitizing surface refused 4005** (T21): it was silently promoted to `"world"` — the one value with takeover power over the user's own body, bypassing every aux gate.
- **Addressed rtc** (T4/T9 updated): the any-leg fanout delivered offers to both of a target's legs → SDP glare in exactly the human-with-voice-leg case. `toSurface` names the leg; omitted = `"world"`, byte-identical to today's mesh. `fromGen/fromSurface` stamps stay; their consumers arrive with the sidecar/relay client (#104).
- **Presence lanes for the voice leg** (T23): `caption`/`typing` admit the bound voice surface (same authority test as attest) — the leg owns the audio clock.
- **Receipt success performs visually** (hold §8): cancel-only receipts drew nothing on success and everything on failure. A matching receipt re-emits `speech {performed:true}` — display consumers draw, `tts.js` skips, and spoken:true says are excluded (that field owns the display plane, so a caption-paced bubble is never drawn twice).
- **surface-test.py refuses to run without `SURF_URL`** — its old default was `:8940`, a production sequencer per AGENTS.md. The owned wrapper remains the entrypoint.

## Receipts at `bda70af`

aux binding **10/10** · hold/fallback **22/22** (15 + 3 B2 vectors + 2 §8 + 2 recut) · owned matrix **40/40** (30 + T19–T23) · voice lifecycle **124/124** · audioctx **7/7** · bundle green (standing Three/WebGPU warnings) · `git diff --check` clean · voice-wiring **34/35** — the `micOn` binding-order failure **reproduces on virgin `ac06295`** (pre-existing, not this branch's; happy to file separately).

## One decision raised, not taken: today's producers of `spoken:true`

Your B2 rule is implemented exactly as specified — and it changes behavior for the two *shipped* producers: `stt.js:45` (human mic transcripts; their audio rides the primary's RTC, no aux leg ever) and the #100 door as used by primary-body synths. Both now hit the no-capability branch and emit local 'speech' where the old flag suppressed it — i.e. listeners will bubble+TTS a transcript whose audio they already heard live. I did **not** re-add a claim-based suppression (a claim is not a receipt — that's this PR's whole point), but the migration ordering is yours to call. Options as I see them: (1) those producers grow voice legs + attest (the honest endpoint, natural under #104's relay client); (2) an interim server-side rule that a say from a session with *live mic state* suppresses re-perform (capability-shaped, not claim-shaped); (3) accept the interim double-perform in the worlds where it occurs. I lean (1) with (2) as the bridge, and can cut either as a follow-up.

— H

antra-tess (approved):
Approved at exact head `bda70aff7ec79732541ec200456c6cb4c951de7f`.

The author-owned recut closes both prior blockers: it is based directly on current main `ac06295`, and `spoken/utt/t0` no longer bypass capability + authenticated `performed` receipts. The three requested `spoken:true` vectors are pinned; restoring the old bypass makes the no-leg and live-deaf-leg vectors fail.

- aux authority: `10/10`
- held speech / receipt / fallback: `22/22`
- owned real-server surface matrix: `40/40`
- voice lifecycle: `124/124`
- AudioContext: `7/7`
- browser bundle green (standing Three/WebGPU warnings)
- clean `diff --check`
- clean merge tree onto `ac06295`

Copied-forward controls on main fail materially: aux authority `5/10`, hold semantics `13/22`, and the owned matrix collapses under ordinary same-id takeover before receipt semantics. A direct mutation restoring the old `spoken:true` bypass fails the two founding fallback vectors.

The hardening pass also closes effective-path seams beyond the requested recut: admission-before-takeover; one retirement funnel for close/expel/travel; fold-proof recent-say lookup; empty-surface refusal; addressed RTC; bound voice presence lanes; and visual performance on receipt success.

One neighboring voice-wiring assertion remains red (`micOn` binding order), but it reproduces on virgin current main and is not introduced by this diff.

**Deployment precondition, not merge blocker:** current shipped `spoken:true` producers do not yet all register a voice surface and attest. Until the relay client/bridge migrates them, deploying this can re-render/re-speak already-audible primary mic transcripts. Merge now as the relay substrate; do not production-deploy #103 in isolation. The phase-1 relay spike must own that migration and field receipt.

PR #105, opened 2026-08-12
seats: the socket owns the world anchor, the avatar profile owns its posed contact — and nothing is silent (#101)

### LINKED ISSUE #101: Seat sockets: different avatar rigs can hover above or sink through the authored support
## Live finding

During a commons chapel visit, Janus mounted a cushion seat and the browser rendered her body roughly **two feet above the cushions**. Dismount worked; the issue is the occupied seat transform, not stale custody.

A socket is authored in model-local coordinates, but one fixed point is not automatically a truthful seat for every avatar/rig/body scale. The same literal socket can put hips, root, feet, or a nonhumanoid body above or through the intended support.

## First task: instrument before correcting

Do not add a magic global y-offset. Build a small browser-side acceptance harness that records, for a mounted body:

- socket world transform;
- avatar root and relevant humanoid hips transform when available;
- body/rig family and avatar bounds;
- actual contact or signed vertical gap between the body's seated support region and the authored seat surface;
- whether the body is browser-local, remote, humanoid, or unsupported.

Use one known-good current seat and the chapel cushion specimen (or a provenance-linked scratch clone if live geometry should not be mutated). Confirm whether the gap comes from socket authorship, avatar root/hips normalization, body scale, pose clip, or mount composition before choosing a correction.

## Desired contract

A seat socket should declare a support plane/seat height and orientation, not silently assume one canonical avatar root. The mounting adapter may derive a rig-specific offset from known body landmarks/bounds, but must:

- preserve the authored socket as the world anchor;
- keep browser-local and remote renderers in agreement;
- support at least two materially different humanoid body scales;
- fail legibly for a rig family with no seat landmark rather than guessing;
- preserve moving-parent/socket behavior and dismount stamping;
- not alter durable mount/custody semantics.

If the correct answer is that each socket needs an authored per-rig-family adapter or explicit body offset, establish that with the measurements instead of hiding it under a universal heuristic.

## Acceptance

1. Existing known-good seat remains within a small declared gap tolerance.
2. Chapel cushion reproduction fails on main with the measured hover.
3. Branch places hips/support region at the declared seat plane without feet/root absurdity.
4. Second humanoid scale also passes.
5. Unsupported/nonhumanoid body gives an explicit unsupported state.
6. Moving socket follows without drift.
7. Third-person, selfie, and independent observer agree.
8. Dismount still stamps the current landing (#18 / PR #98); no custody or folded-state mutation beyond existing mount.
9. Browser regression fails on main and passes on branch; no live-world mutation is required.

## Provenance

Found by Janus/Repligate during a real chapel sit. The body’s visible hover outranks any assumed socket arithmetic.

cormundus:
A seated body's root is its floor reference, so mounting a root at a socket floats the whole body one pelvis-height above the seat — Janus's two-foot hover in Digi's chapel (#101). Phase A measured the seam: the chapel socket is authored 7 mm above the pad (the authorship was clean), the sin is `root = socket point`, and no bone formula survives two rigs (contact spread 7.4 cm > hips spread 4.3 cm). The accepted theorem, implemented here exactly as contracted ([design](https://github.com/anima-research/eidoverse-worlds/issues/101#issuecomment-5261034371) → [B1–B5 amendments](https://github.com/anima-research/eidoverse-worlds/issues/101#issuecomment-5261102753) → authorization): **the socket owns the world anchor; the avatar profile owns where that body's posed contact sits relative to its root.**

## What ships (one bounded PR, as scoped)

**The contract as shared pure math** — `client/lib/seatcore.js`, the forecast.js/motioneval.js pattern: schema validation (placeholder-shaped patch evidence is invalid evidence), the serve-time status verdict, the B1 socket-anchor rule, the two-half gate, the world-up correction, the rider-scale definition, the fetch-generation guard. Browser, sequencer, and agent evaluate this one file; `tools/seatcore-test.ts` pins it with hand math (55 checks), including the B2 discriminator — a pitched/rolled fixture that numerically fails a parent-normal implementation by its exact lateral term (`contactY·sin θ`).

**One judge, three readers** — `server/seats.ts` + `/avatars`: profiles live in `assets/opt/seats/profiles.json` (tmp+rename, append-only provenance log), judged at serve time against the current avatar and clip bytes (mtime-cached sha256), and each roster entry carries its pre-judged `seat` verdict plus the clip digest; the store's monotonic revision rides an `x-profiles-rev` header (the array shape predates seats and stays an array). A stale value cannot be served as fresh, and no consumer ever rehashes a VRM.

**Write authority (B4)** — `POST /seat-profile` writes the *proposed* slot only, and requires a **named actor**: a tokens.json bearer or a home-node-verified aid1 identity — the same two legs `/upload` trusts. The anonymous door token gets a 401 pointing at the operator-import lane. Countersign has **no HTTP path**: `tools/seat-accept.ts` is run by the operator on the box; the live server notices by mtime and pushes the same generation-bearing `avatar-profile-updated` a proposal gets. Bearers are never logged or echoed.

**The correction, at the one seam all consumers share (B2)** — `mountTransform` (browser-local at main.js, remotes at remotes.js) and `effective.ts`'s body branch append one subtraction after the existing socket resolution: `root = socketPoint − ŷ·(seatContactY × riderScale)`, along **world up** — mounted bodies render upright (yaw only), so a tilted parent's normal would smear them laterally while their contact geometry stays vertical. Yaw, part sockets, mount overrides, precedence, and dismount stamping are byte-for-byte the paths that exist.

**Nothing silent (B1/B3)** — the correction applies only behind the full gate: authored `seatAnchor: "surface"` (absent = legacy-root = today's composition, by construction — no heuristic infers intent from socket height), a countersigned fresh verdict, and the runtime truth of each consumer — the browser gates on the mixer slot *actually playing* and the digest of the clip bytes it *actually loaded* (hashed once at fetch in `vrmaBytes`; a filename is not an identity); the headless reader has no mixer, so its runtime truth is the contract identity, with steady-state parity pinned by test. Every closed-gate state is declared in all three consumers: an `≈` on the nameplate + one console line, and `(seat approximate: <reason>)` on the agent's seated lines. Root-at-socket can still happen — it can no longer happen *silently*.

**Generations (B3, the #95 lesson generalized)** — both update events bump a per-name generation before the refetch departs; a resolution stamped before a bump is discarded whole, and a lower `profilesRev` never replaces a higher one. A slow response from before an acceptance cannot roll the acceptance back. The agent gains tested handlers for `avatar-updated` and `avatar-profile-updated` and fetches `/avatars` over its hardened `httpBase` — the same base `/geom` and terrain already ride.

**Evidence hygiene** — seatlab is a `tools/` harness (`tools/seatlab/`), never client runtime: the `_v` register-aliasing bug is fixed by moving the seat claim into pure `seatcore.seatClaim` (nonzero-Y regression pinned in bun), and the derivation records the winning mesh/vertex/coordinate with its support patch — an isolated winner (skirt hem, accessory, outlier) is flagged, never proposed.

## Receipts

**Real derivations** (`tools/receipts-101-phaseb/derivation-receipts.json`): 3 bit-identical runs per rig; the rebuilt instrument reproduces Phase A's accepted numbers exactly — claude **0.2055** (winner char1001 v259, patch 31 verts @ 1.84 cm), aletheia **0.2791** (Body_(merged)baked001 v282, patch 13 @ 0.95 cm). Proposed via the operator lane, countersigned with the contract comment as receipt.

**The founding bench, sat upon** — the chapel-cushion clone at the exact prod transform with **Digi's four authored sockets verbatim** (read from public `/geom`, the live chapel never written); `seat`/`seat3`/`seat4` migrated to `surface` by authored act, `seat2` left legacy as the in-place control:

claude seated on the chapel bench
both rigs seated

Phase A's hover screenshots of this same bench are the before. The numbers (`numeric-receipts.json`, `ACCEPTANCE.md` for the full ×11 table):

- **Gap:** claude contact = 1.4912 + 0.2055 = 1.6967 = the socket plane, **+6.7 mm** above the pad; aletheia ≈ 0–7 mm. Both ≪ the 5 cm cushion tolerance, inside the 2 cm rigid bound too.
- **Three consumers, one value:** browser-remote 1.4912 = headless 1.4912 (claude/seat); aletheia 1.4108/1.4109; browser-local 1.4910 on seat4 (pad authored −0.29 vs −0.291 — authorship, not drift).
- **Legacy control, live:** same cushion, `seat2`: `ground height 1.70m … (seat approximate: legacy socket)` — today's composition, declared.
- **Moving tilting parent (B2, live):** pendulum amp 0.5 about X, five mid-arc samples across 1.6 m of travel: **dY = 0.2055 exactly, dX = dZ = 0 exactly** — the parent-normal signature (±9.8 cm lateral) is absent. The pre-hydration first sample is the designed declared-approximate transient. Headless parity: `3.34m … (riding its pendulum)`, no suffix.
- **Fail-on-main:** `tools/receipts-101-phaseb/fail-on-main-lifecycle.txt` — 19 named failures (main's server + agent, branch instruments), 3 passes = exactly the invariants that must hold on both sides; main's own header reads `seated on crate1.` — the silent hover, verbatim.

**Suites on the committed head:** smoke 85/85 · seatcore 55 · **seat-lifecycle 22/22** (real scratch sequencer + real WorldAgent: 401 for anonymous, 403 for smuggled `accepted`, proposal broadcast, the agent's `look()` correcting **in place** on countersign — 1.00 → 0.79 = 1 − 0.2055 — stale-naming-which-bytes, unsupported-with-refusal, dismount unchanged) · effective 42 · motioneval 33 · supportclass 29 · uneven-support 15 · denoise 31 · remotes-lifecycle green · comptest 33/0 · worldops 23/0. permtest 21/2 and voice-lifecycle 122/2 are pre-existing on main (verified on a main worktree control).

Closes #101.

— I.M. & Cormundus


antra-tess (changes requested):
## Changes requested — architecture is present; three integrity boundaries still fail

Reviewed exact head `036abf171a4712a1d34feb9c4d1468a50d70056d`.

Independent positives:

- `seatcore-test`: all 55 contract checks pass;
- real scratch sequencer + real `WorldAgent`: 22/22 on a verified-free replacement port;
- effective-transform and motion suites green;
- browser bundle and `diff --check` green;
- source implements the accepted B1/B2/B3/B4/B5 shape: shared pure contract, explicit legacy/surface anchor, world-up correction, actual-slot + clip-byte gate, named proposal authority, offline countersign, served verdicts, and three consumers.

Three blockers remain.

### B1 — update invalidation is not generation-safe for the cases it claims to protect

Both browser `seats.js` and `WorldAgent.fetchSeatProfiles()` stamp only names already present in `verdicts`/`seatVerdicts`. For an initial fetch (maps empty), if `avatar-profile-updated` arrives while `/avatars` is in flight, the response later does:

```js
const stamped = stamps.get(e.name) ?? guard.stamp(e.name)
```

That reads the **post-bump** generation at response time and accepts the pre-bump response. The stale fetch can therefore overwrite the update exactly when a name was not already cached.

A second side door: update handlers bump the guard but do not invalidate the held verdict. If refetch is slow or fails, the old accepted value remains in the map and stays load-bearing, despite comments/design saying the seat becomes declared-approximate until a fresh verdict lands.

Use a whole-request generation/epoch or capture all relevant stamps before departure, incorporate the event's `rev`, and mark the affected name pending immediately. Pin in **both browser and real WorldAgent paths**:

1. initial empty-cache fetch → update arrives → old response resolves: old response discarded;
2. accepted verdict held → avatar/profile update → refetch delayed/fails: correction stops immediately and becomes declared-approximate;
3. fresh response under the new generation restores it;
4. different names remain independent.

### B2 — the rights-bearing profile store accepts dangerous/unbounded keys

`validateProfile()` accepts any nonempty `avatar` and any nonempty `pose`; `SeatStore.propose()` indexes plain objects with them:

```ts
(this.store.profiles[name] ??= {})[pose] ??= {}
```

`avatar:"__proto__"` or `pose:"__proto__"` reaches inherited `Object.prototype` and can mutate process-global prototype state. This slice supports only roster avatars and `sitchair`; the write door must enforce that contract.

Require exact safe avatar-name syntax, exact `pose === "sitchair"`, own-property/null-prototype maps, and a current roster avatar match (or an operator import that explicitly names unavailable bytes). Add real-door vectors for `__proto__`, `constructor`, unknown avatar, and wrong pose, proving 4xx, no store/log mutation, and no prototype pollution.

### B3 — write/provenance failure can leave applied state without its claimed receipt

`persist()` mutates/increments the live store, renames `profiles.json`, and only then appends `profiles.log.jsonl`. A crash or append failure after rename leaves a body-moving state that the “reconstructible append-only provenance log” does not contain. A write failure earlier also leaves the in-memory object/rev mutated. The offline accept tool and live proposer additionally have no revision/CAS/lock, so concurrent writers can overwrite one another from stale snapshots.

This is exactly the class where provenance is the safety boundary. Make each write transactional enough that:

- failure cannot leave an unreceipted load-bearing in-memory or snapshot state;
- operator countersign cannot silently clobber a newer proposal (expected revision or a shared lock/CAS);
- startup can detect/refuse an inconsistent snapshot/log pair;
- no update event or HTTP success is emitted before durable state + provenance succeed.

Add fault injection for snapshot write/rename, log append, and stale-revision countersign. The exact storage mechanism is open; the current reconstructibility claim is not earned.

### Harness blocker — checked-in lifecycle test can inspect the wrong door

`seat-lifecycle-test.ts` hardcodes port `8977`, suppresses child stdout/stderr, and does not prove the responding listener is its spawned child. On this host 8977 was occupied: the checked-in test timed out; after moving the unchanged harness to verified-free 9041 it passed 22/22. This repeats the stale-port false-receipt family from #83/#91.

Use a preflighted/ephemeral port, preserve child startup errors, and require child/process ownership via nonce or equivalent before counting `/avatars`. The committed test—not a manually rewritten copy—must pass.

Everything else should remain bounded. Do not redesign the profile theorem or replace honest approximation. Return one revised exact head plus fail-on-old receipts for B1–B3 and the owned-door harness; then I will rerun the full acceptance and await the independent reviewer already assigned.

cormundus:
**Revision pushed — head `b5957d4`. All three boundaries folded, the door is owned, and the reviewed head fails every new check by name.**

### B1 — one cache, whole-request epochs, pending on arrival

The guard and cache logic existed twice — one copy per consumer — and each copy had the holes you named. It now exists once: `seatcore.makeVerdictCache`, consumed by both `client/lib/seats.js` and `WorldAgent` (the browser supplies a `/avatars` fetch, the agent its `httpBase` fetch — nothing else differs). The guard is epoch-based: every bump advances a global epoch recorded against the name, a request stamps the epoch it **departed** under, and a resolution is refused for any name bumped after departure — cached or never-seen (your empty-cache race, closed at the root). `note()` additionally demotes the held verdict to `pending` **before** its refetch departs — the gate declares `profile update pending` and the old value stops moving bodies the moment the event lands, refetch or no refetch. Event revs are floors: a response older than an announced revision is pre-event by definition and refused whole; a response that cannot state its revision is refused once any rev has been announced.

Your four vectors are pinned in `tools/seatcore-test.ts` with an injectable transport (the races are deterministic there — time and resolution order are the test's to control), and the live chain runs against the real door in the lifecycle matrix: **seatprobe** — a disposable avatar the test uploads, profiles, countersigns, and seats a real `WorldAgent` on (`0.79m`, corrected) — has its bytes re-uploaded mid-sit, and the agent's own `look()` moves through `pending`/`stale (avatar bytes changed)` without ever re-serving the old value; `/avatars` serves `stale` naming which bytes.

### B2 — the write door refuses what it cannot safely hold

`validateProfile` now requires roster-name syntax (`[a-zA-Z0-9_-]{1,48}`), refuses the language's reserved names (`__proto__`/`constructor`/`prototype`) explicitly, and pins `pose === "sitchair"` — the slice's whole scope. The store's maps are **null-prototype at every layer** (including on load from parsed JSON, where `obj[k] ??=` reads inherited slots — the pollution door), and proposals must name a **rostered** avatar (there must be bytes to judge against); the operator-import lane may carry an unrostered name only via an explicit flag that lands in provenance. Real-door vectors in the lifecycle matrix: `__proto__`, `constructor`, `prototype`, unknown avatar, wrong pose — all 4xx, rev unmoved, no events, `Object.prototype` asserted clean. On the reviewed head the same vectors return **`ok:true` at revs 1–4 and the pollution assertion fails** — the vulnerability was live (`fail-on-reviewed-head-r2.txt`).

### B3 — provenance ⊇ applied state, always

`persist` is provenance-first: the next state is built without touching the live one, the log line appends, the snapshot renames, and only then does memory swap. An append failure changes nothing anywhere. A snapshot failure leaves the log AHEAD and reports failure — and startup folds receipted writes forward, rewriting the snapshot to match its receipts: durable-with-receipt or not at all. A snapshot **ahead** of its log is a state with no receipt: quarantined, served as `missing`, refused for writes (503), said loudly. Countersign now requires `--expect-rev` — the revision the operator `list`ed and reviewed — and both processes reload from disk before every mutation, so a second writer's stale snapshot cannot clobber: the cross-process vector (A proposes, B countersigns from its old view) refuses with "re-list and re-review". Fault injection in `tools/seats-store-test.ts` (26 checks) via an injectable fs surface: append-fail, rename-fail + recovery, quarantine, stale-revision countersign, two-writer CAS.

### Harness — the door is ours or the test is over

The port is OS-assigned (bind-0 preflight), asserted silent before the child spawns, the child's stdout/stderr are preserved and printed on any fatal, and before a single check counts, the responding server must have **written this run's scratch `WORLDS_DIR`** — a listener that answers without leaving tracks in our tmpdir is somebody else's door and the test dies rather than inspect it. The committed file passes as committed: **37/37**.

### Receipts, exact head `b5957d4`

Suites: seatcore **71** · seats-store **26** · seat-lifecycle **37/37** (owned door) · effective 42 · motioneval 33 · supportclass 29 · uneven-support 15 · denoise 31 · remotes-lifecycle 30 · smoke 85/85 · comptest 33/0 · worldops 23/0 · browser bundle green (127 modules). permtest 21/2 and voice-lifecycle 122/2 pre-existing on main, as before.

Fail-on-reviewed-head (`tools/receipts-101-phaseb/fail-on-reviewed-head-r2.txt`): seats-store **21 named failures** (`__proto__` accepted at rev 1, pollution assertion fails, unreviewed countersign accepted); lifecycle **12 named failures** (dangerous keys 200, rev moves on refused writes, no-CAS countersign); seatcore-test fails at import — the shared cache did not exist on that head, which is the point: the duplicated copies it replaces were where your races lived.

— I.M. & Cormundus

antra-tess:
## Revision review — B1, B2 and the owned-door harness close; B3 still has a real concurrent-writer race

Reviewed revised exact head `b5957d42960d87f2ea67403199f305ccfecfca59` against prior reviewed `036abf171a4712a1d34feb9c4d1468a50d70056d`.

Independent positives:

- shared contract/cache suite green, including the empty-cache in-flight update, immediate pending demotion, per-name independence, and fresh restoration;
- store suite green for dangerous keys, roster/pose scope, prototype safety, append/rename fault injection, log-ahead recovery, snapshot-ahead quarantine, and sequential stale-revision countersign;
- checked-in lifecycle now owns a free door and preserves child diagnostics: **37/37** through a real sequencer and real `WorldAgent`;
- effective and motion neighbors remain green;
- `diff --check` clean.

So review items B1, B2, and B4 (owned harness) are closed.

### Remaining blocker — the store still has no mutual exclusion across its two real writers

`freshen()` followed by `persist()` is a check-then-act sequence with no lock/CAS around the commit. `expectedRev` protects an operator who reviewed an old revision, but it does not prevent the server proposer and offline operator tool (or two proposers) from both freshening at rev N and then each appending/renaming rev N+1.

The current “second process” test is sequential: A fully advances disk before B calls `accept()`, so B's mtime-based `freshen()` sees it. It does not exercise overlap.

I forced the overlap deterministically by injecting B's `propose()` inside A's provenance-append callback after A had already built `next` from rev 0. Exact revised head produced:

```json
{
  "A": {"ok":true,"rev":1},
  "B": {"ok":true,"rev":1},
  "snapshotRev":1,
  "snapshotY":0.21,
  "log":[
    {"rev":1,"actor":"B","y":0.31},
    {"rev":1,"actor":"A","y":0.21}
  ]
}
```

Both writers report success at the same revision; B's accepted write disappears from the snapshot; the provenance log now has two contradictory rev-1 records, and startup folding skips one or both depending on snapshot state. This defeats both CAS and reconstructibility.

Please serialize the whole **reload/current-revision check → build next → provenance append → snapshot rename → in-memory swap** transaction with one shared inter-process lock (or another genuinely atomic CAS). The offline `seat-accept` tool and live server must use the same lock. Acquire first, then force a disk/log reload under the lock rather than relying only on mtime; release in `finally`; stale/dead lock handling must be bounded and explicit.

Add a deterministic overlap regression (not merely sequential stale state) proving:

1. revisions are unique and monotonic;
2. both successful writes survive in reconstructible state, or one refuses and can be retried after reread;
3. an operator countersign cannot interleave with a proposal and accept a record other than the reviewed one;
4. lock acquisition failure changes neither log, snapshot, nor in-memory state;
5. startup sees no duplicate-revision provenance.

Please keep every other correction unchanged and return one exact head. This is the last blocker I found in the four requested boundaries; I will rerun the full battery after it lands.

cormundus:
**Revision pushed — head `8e7897a` (the lock and its vectors land at `f537271`/`8d64d3c`; the tip commit archives this correspondence into the receipts dir). The last blocker is closed: every write runs whole under one shared inter-process lock, and your forked history is now a committed regression that the prior head fails by name.**

### The lock

One on-disk lock per store directory (`seats/.write-lock`), shared by the live server and the offline `seat-accept` tool alike — they construct the same `SeatStore`, so they cannot NOT share it. Exclusive-create (`openSync "wx"`) is the atom; the file carries `{pid, ts}`. Every mutation now runs **whole** under it: acquire → **forced reload** (never mtime trust — the lock is when the disk is the truth) → precondition re-check → provenance append → snapshot rename → in-memory swap → release in `finally`. Schema refusals stay outside the lock — garbage never contends with real writers.

Stale and dead locks are bounded and explicit: acquisition is a bounded spin (default 3 s, constructor-tunable), timing out is an **answer** — the caller refuses with nothing touched and says so; a lock older than its liveness window (10 s) belongs to a dead writer and is broken with a console line naming the corpse. The lock rides the real fs even under fault injection — injected faults are the transaction's to survive, not the lock's.

`expectedRev` now means what it claims: the comparison happens against the disk's truth **at commit time, under the lock**, so a proposal that interleaved anywhere before that moment moves the revision and the countersign refuses rather than bless the unseen.

And a history that forked anyway — a pre-lock writer, a hand edit — is caught at the door: startup scans the log for revision monotonicity, and duplicate or non-increasing revisions quarantine the store (served as `missing`, writes refused, said loudly). No fold order over a forked log is defensible, so none is attempted.

### The overlap regression, deterministic

`tools/seats-store-test.ts` gains your exact interleaving: writer B's `propose()` injected inside writer A's provenance-append, after A has built its next state. On this head:

1. **unique, monotonic revisions** — A commits rev N+1; B, overlapped, **refuses** (`another writer holds the seat-profile lock… nothing was written`); B retried after release lands at N+2; the log's revision sequence asserts strictly increasing;
2. **both writes survive, reconstructible** — a fresh load lists both records;
3. **the countersign cannot accept the unseen** — a proposal landing between `list` and `accept` moves the revision; the acceptance refuses with "re-list and re-review", and succeeds against the re-reviewed revision;
4. **lock failure changes nothing** — a held lock times the writer out with log, snapshot, and memory byte-identical; a dead holder past the liveness window is broken and the write proceeds;
5. **no duplicate-revision provenance survives startup** — a hand-forked log quarantines.

**Fail-on-prior-head** (`tools/receipts-101-phaseb/fail-on-prior-head-r3.txt`): the same committed harness against `b5957d4` reproduces your finding exactly — B's overlapped write returns `ok:true` and the log reads `[1,2,3,4,5,6,6,7]`: rev 6 twice, four named failures.

### Receipts, exact head `8d64d3c`

Everything else unchanged, as asked. Suites: seats-store **37** (the five vectors + everything prior) · seatcore 71 · seat-lifecycle **37/37** (owned door) · effective 42 · smoke 85/85. The store's write path is the only code touched; the browser bundle and consumers are byte-identical to the head you reviewed.

— I.M. & Cormundus

antra-tess:
## Revision review — serialization exists, but the lock can be stolen from a live writer

Reviewed exact head `8e7897a698a25166ad0e2205a8ce624f071cb098`.

The new shared lock closes the ordinary overlap race, the store now quarantines duplicate/non-monotonic provenance revisions, and the focused store/lifecycle evidence remains strong. But the stale-lock rule is not a mutual-exclusion rule yet:

```ts
if (Date.now() - holder.ts > lockStaleMs) {
  unlinkSync(lockFile);
  continue;
}
```

Age is not proof that the holder is dead. A live transaction delayed beyond `lockStaleMs` — slow disk, scheduler pause, debugger, injected fault, or simply a long operation — has its lock unlinked by the second writer. Worse, the first writer's unconditional `releaseLock()` may later unlink the second writer's lock. The original duplicate-revision/lost-write race is therefore reachable again after the timeout boundary.

I reproduced this deterministically on the exact head by setting the supported constructor options `lockStaleMs:10`, pausing live writer A for 20ms inside its append, and letting B acquire. Result:

```json
{
  "A":{"ok":true,"rev":1},
  "B":{"ok":true,"rev":1},
  "snapshotRev":1,
  "snapshotY":0.21,
  "log":[
    {"rev":1,"actor":"B","y":0.31},
    {"rev":1,"actor":"A","y":0.21}
  ]
}
```

The process log says it plainly:

```text
[seats] breaking stale write lock (pid <the current live pid>, 86ms old)
```

Please make ownership and stale recovery real:

1. mint a unique lock-owner token/nonce and retain the fd/owner record;
2. release only if the on-disk lock still belongs to that exact owner — never blindly unlink a successor's lock;
3. do not break a lock merely because it is old. On this single-host design, verify the recorded PID is dead (and guard PID reuse with the nonce/process-start identity where practical). If liveness is unknown, time out/refuse rather than violate exclusion;
4. if crash recovery must use a lease, add owner renewal/heartbeat and only steal an expired lease under a contract that cannot overlap a live holder.

Add a deterministic **live-but-old holder** regression proving B times out/refuses, A completes alone, and A's release cannot remove a successor lock. Keep the existing dead-lock recovery and ordinary overlap tests.

Everything else in the revision should remain unchanged. This is the remaining transaction blocker; please return one exact head after fixing it.

cormundus:
**Revision pushed — head `d71546a` (the fix lands at `527256f`). Ownership is a nonce, liveness is a verified pid, age proves nothing, and even a pathological steal now aborts a transaction instead of forking history.**

### Ownership and liveness, made real

- **The lock record is `{pid, nonce, ts}` and the holder retains its nonce** — ownership is the nonce, never the path. `ts` is informational only; nothing anywhere breaks a lock because it is old.
- **Liveness is pid-verified**: `process.kill(pid, 0)` on this single-host design — `ESRCH` means the holder no longer exists; anything else (alive, or `EPERM`) means a holder we **wait for, however old its stamp**. If liveness is unknowable, we time out rather than violate exclusion. Pid reuse cannot resurrect a broken claim: a reused pid is a *live* process, so the lock is simply waited on — and the nonce keeps release from ever touching a lock it didn't mint.
- **Breaking a dead lock is a rename-steal, not an unlink**: rename is the atomic claim (two breakers cannot both win), and the tomb's nonce is verified against the record that was judged dead — if the rename caught a newer live lock instead, it is restored untouched and the breaker retries.
- **Release only releases its own**: the on-disk nonce must match the one we minted, or release is a logged no-op. Your second finding — the first writer unlinking a successor's lock — is structurally impossible now.
- **Commit gates as the last line**: `ownsLock()` is re-read immediately before the provenance append *and* before the snapshot rename. If any pathological path ever takes the lock from a live writer anyway, the transaction **aborts** — pre-append with nothing written; post-append with the snapshot withheld, the log line standing as a receipt for recovery to fold (or, if a successor also wrote, for the monotonicity quarantine to catch). An exclusion breach degrades to a refused write, never to two rev-N records.

### The regressions (all committed in `tools/seats-store-test.ts`, 44 checks)

- **Live-but-old holder** — the lock names this test process's own pid (alive by construction) with a stamp from an hour ago: the second writer **times out with log, snapshot, and memory byte-identical, and the holder's lock file untouched**. On the prior head this exact vector steals the lock and writes (`ok:true` — your finding, reproduced by the committed harness).
- **Dead holder, fresh stamp** — pid 999999999 stamped *now*: broken immediately, because death is verified and age is irrelevant in both directions.
- **Release-safety + gate** — the lock is swapped to a successor's mid-transaction (inside the provenance append): the writer aborts at the second gate, its snapshot and memory stay unchanged, its release leaves the successor's lock in place, and the receipted orphan line folds forward on the next load. On the prior head, the control run **crashes at this probe** — the old unconditional release had *deleted* the successor's lock, so there was nothing left to read. The crash is the finding (`fail-on-prior-head-r4.txt`).
- All prior vectors retained: ordinary overlap refusal + retry, monotonic revisions, countersign CAS, lock-timeout no-op, duplicate-revision quarantine, fault injection, recovery.

### Receipts, exact head `d71546a`

Only `server/seats.ts` and its test moved. Suites: seats-store **44** · seatcore 71 · seat-lifecycle **37/37** (owned door) · effective 42 · smoke 85/85. Fail-on-prior-head: `tools/receipts-101-phaseb/fail-on-prior-head-r4.txt` — the live-holder steal reproduced by name, and the control aborting where the prior head deleted a lock it never owned.

— I.M. & Cormundus

antra-tess (approved):
Approved at exact head `f3dcb4bbc043ff633c56e7f06b40eb3a9fa95032`.

The round-four lock closes the remaining transaction blocker:

- ownership is nonce-bound;
- release removes only the lock minted by that owner;
- age alone never authorizes theft;
- dead-holder recovery verifies PID liveness and claims the corpse by atomic rename;
- unknown/live holders time out rather than violating exclusion;
- commit gates refuse if ownership changes;
- startup quarantines duplicate/non-monotonic provenance.

Independent reproduction of my prior overlap proof now yields exactly one rev-1 receipt and B refuses with 503; no duplicate revision and no lost successful write. The checked-in live-but-old-holder, dead-holder, successor-lock, lock-failure, log/snapshot fault, stale-countersign, and forked-history controls are green.

Final independent battery:

- shared seat contract/verdict cache: all checks green;
- store integrity/concurrency/fault matrix: all checks green;
- real owned-door sequencer + real `WorldAgent`: **37/37**;
- effective-transform and motion neighbors: green;
- browser bundle: green, with only the standing Three/WebGPU warnings;
- `git diff --check`: clean;
- current-main merge tree: clean (`origin/main` remains the PR base).

The full #101 bounded contract is now evidenced: legacy sockets remain byte-identical and declared approximate; only authored surface anchors can consume accepted content-bound profiles; world-up correction, runtime clip/action gate, named proposal authority, offline countersign, transactional provenance, immediate invalidation, browser/headless delivery, and no live chapel mutation.

Repository disposition: approved for merge. Deployment and any real profile countersign remain separate operations with their own receipts.

PR #117, opened 2026-08-12
ammodoll: Bullet hair per-lock sim in the ragdoll

antra-tess:
The source's hair system, inside the fall: 364 segments at 61fps, momentum-true whip and drape. A/B against springbones via --no-hair-springs imports. Testbeds: mythos_hairtest (springbone) vs mythos_hairbullet (bullet) on staging.

🤖 Generated with Claude Code


antra-tess:
superseded by #118 (clean re-land)

PR #121, opened 2026-08-13
sunflower: the new createFlora species, planted from the ground UI

SkyeShark:
eidoverse-video [`09ffd28`](https://github.com/SkyeShark/eidoverse-video/commit/09ffd28) adds a **sunflower** crop to the vegetation brush — fitted petal whorls layered over a scanned seed disc, a 3D bract cup behind, tube petioles/cane, and a `heading` placement option so a planted field faces one way together. This PR is the client half:

- **Ground UI**: a `sunflower field` planting joins meadow / tufts / mojave / corn field.
- **`presetStrokes`**: a sunflower bag composes three seed variants interleaved through one row grid (stride 3), all sharing a single `heading` (default `-0.6`; a caller's heading wins). Same pattern as the corn interleave.
- **`ensureFloraAssets`**: primes `sunflower_fit.json` — the measured alpha-envelope card fit — alongside the four-map set (the module falls back to built-in envelopes without it, so missing files still degrade gracefully).
- **README**: minimum library pin bumps to `09ffd28` and says why.

Tests: `bun tools/flora.test.ts` → **67 passed, 0 failed** (7 new sunflower checks: stroke count, stride tiling, per-variant seeds, shared heading, caller-heading override).

Line endings: every touched blob verified LF-only in the commit (learned that lesson on #16).

Deploy note: serving this needs the library checkout at `09ffd28`+ so `/library/eidoverse/vegetation_sunflower_gen.js` and the `sunflower_*` assets exist — `git -C $EIDOVERSE_DIR pull`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,9 @@ vendored here — point `EIDOVERSE_DIR` at a checkout of
 (our fork of [SkyeShark/eidoverse-video](https://github.com/SkyeShark/eidoverse-video);
 the default branch is exactly what production serves, including a sky patch
 upstream doesn't carry). Pristine upstream `main` at
-[`8b37f0f`](https://github.com/SkyeShark/eidoverse-video/commit/8b37f0f)+
-also works. The `grass` verb loads `eidoverse/vegetation.js` from the
+[`09ffd28`](https://github.com/SkyeShark/eidoverse-video/commit/09ffd28)+
+also works (that revision adds the `sunflower` species this client's ground
+UI offers). The `grass` verb loads `eidoverse/vegetation.js` from the
 library at runtime, so the vegetation brush, its wind-anchoring fixes, and
 the field `dispose()` this client calls on retirement all live there rather
 than here — `git -C $EIDOVERSE_DIR pull` is how you get them.
--- a/client/lib/build.js
+++ b/client/lib/build.js
@@ -819,6 +819,8 @@ function paintGround(body) {
     'mojave desert': () => ({ preset: 'mojave', width: 90, depth: 80, center: [0, 0] }),
     'corn field': () => ({ species: 'corn', width: 40, depth: 30, center: [0, 0],
       rows: { spacing: 0.9, plant: 0.26 }, corn: { peelChance: 0.25 } }),
+    'sunflower field': () => ({ species: 'sunflower', width: 34, depth: 26, center: [0, 0],
+      rows: { spacing: 0.85, plant: 0.5 } }),
   };
   const growGrass = () => {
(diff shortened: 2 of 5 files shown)


antra-tess (approved):
Reviewed exact head `ce947c852c8f7b231b3ae6c72c5c6d57c6ff7480` against `df13f3c3e03cc76beec3ace3970ce021ce11dc54`, together with the exact upstream dependency `eidoverse-video@09ffd289fd77edaf59142595b0093277757d77ff`.

**Disposition: approve.**

Independent receipts:

- `tools/flora.test.ts`: **67/67**
  - three sunflower variants interleave one row grid at phases 0/1/2, stride 3;
  - seeds and heading behave as claimed;
  - explicit stride remains caller-owned;
  - generic grow/replace/mow lifecycle retires every stroke, update hook, clearing mask, and field resource.
- copied revised assertions against current main fail at the founding sunflower expansion (main returns one unplanted stroke), so the new behavior is discriminating rather than helper-only green.
- `tools/grass-quality-test.ts`: **57/57**.
- `tools/collider-test.ts`: **34/34**.
- branch and matched-main browser bundles both complete at 163 modules; the same three standing Three/WebGPU warnings occur on both trees and are not introduced here.
- clean diff, clean merge tree.

Upstream/runtime receipts:

- the full sunflower chain from `8b37f0f..09ffd28` is present: species registration, generator, heading placement, fit JSON, and all four map files;
- a real scratch sequencer rooted at exact `09ffd28` served `vegetation.js`, the sunflower generator, fit JSON, and every texture at HTTP 200 with byte hashes matching the checkout;
- exact upstream `createFlora()` construction produced all three seeded variants with finite geometry, shared-heading jitter, real update/dispose handles, and balanced hook registration/removal;
- missing fit data used the built-in envelopes successfully;
- missing optional normal/roughness/translucency maps still built and disposed successfully;
- sunflower meshes carry `noSupportCheck`, `noWalkable`, and `noCameraCollide`, preserving the established contract that `createFlora` ground foliage is yielding and never physical/navigation authority.

**Deployment precondition (important, not a code blocker):** production's current staged `eidoverse-video` snapshot does **not** yet contain `vegetation_sunflower_gen.js` or the sunflower assets. Sync the *whole* exact upstream `09ffd28` tree (or later proven-compatible tree) before/atomically with deploying this worlds head, then verify the six served sunflower artifacts. Do not deploy the worlds UI alone and leave it offering a species the runtime library cannot construct. Preserve the prior library snapshot for rollback.

No production mutation or live sunflower field was performed during review.

PR #125, opened 2026-08-14
travel: in-session world walking, as an eidoverse-local extension

octopusburrow:
**In-session world travel, as an eidoverse-local feature.** Retitled and rescoped from "RFC-005 reference implementation" after antra's and Mica's read — see the thread on anima-research/mcpl#4, now closed in favor of this.

## What it does

An agent's world is sealed into its credential at mint time, so today the only way to change worlds is to drop the connection and re-authenticate with a different token — and on a door you don't operate, you can't do even that. Meanwhile the sequencer has supported live travel all along: its `join` handler leaves world A and joins world B on an existing socket, with orphan-leg reaping, and browser clients travel this way every day. The capability existed; the MCPL surface had no verb reaching it.

This adds the verb, in two lanes over one mechanism:

- **`travel {world}`** — a door tool. What a plain-MCP client uses; no channel vocabulary needed.
- **`channels/open`** naming a sibling world — what a generic MCPL host uses without knowing eidoverse tool names.

Both route through one `travelGate()`, so they cannot drift apart. Measured **14ms** for in-session travel including the host round trip, against ~2000ms for teardown-and-redial.

## Why this is not a protocol change

The earlier version of this PR proposed an MCPL spec change and gated join on a new `channels.join` capability. That was the mistake, and it's worth stating precisely because it was invisible until someone asked the right question: **`channels.join` is not in §6.2's capability list.** I invented it. A feature set declaring it fails this repo's own boot validator.

Remove that one string and everything else here is a local extension:

- Authorization is **`channels.lifecycle`** — already spec-legal, already granted, and already means "may act on channel lifecycle."
- The feature declares itself as **`eidoverse.travel`** in `FEATURE_SETS`, the one vendor-namespaced slot in the manifest. `uses: [channels.lifecycle, tools]` — both §6.2 leaves.
- No new JSON-RPC method. This repo has zero bespoke methods and that stays true.

It also closes a gap the audit surfaced independently: **`eidoverse.world` never declared any lifecycle capability**, so the feature set owning world presence didn't mention the operation that changes worlds.

## The invariants, kept

Mica's five review points from the mcpl#4 thread are all implemented and tested. They were always implementation properties — the RFC was just where I happened to write them down, and they survive the move intact:

1. **Attachment epoch** — monotonic per connection, incremented *after* the new attachment is live (a transition that dies mid-flight must not advance it), surfaced as `metadata.epoch`. The proposal states the epoch it *would* commit to; the client holds it pending until it observes the commit.
2. **Prepare/commit with host veto** — the host's acceptance is part of the transition, not a notification after it. A declining or silent host aborts the move; the body does not go somewhere its host refuses to deliver.
3. **No silent dial fallback** — a denied `?world=` refuses the connection with a reason rather than quietly landing you somewhere else.
4. **Founding ≠ joining** — creating a world that doesn't exist needs separate `create` authority. Scoped, after three rounds of getting it wrong, to destinations the *agent* chose: arriving at the world your operator minted you into is not founding, and gating it bricks fresh deployments.
5. **Both address forms** — `channelId` and `type`/`address`, since generic host surfaces primarily expose IDs.

## Tests

**33 passing** (`bun tools/join-rfc005.test.ts`), including the property that matters most: the tool and `channels/open` refuse the *same* world for the *same* credential, and a refused travel moves nothing. Also covers the prepare/commit refusal path with declining-host and mute-host fixtures, and regressions for each of the three founding-gate bugs three rounds of adversarial review turned up.

## The open question

I kept the `channels/open` lane. It costs nothing now that it needs no invented capability, and it's how a host that only speaks MCPL travels here. But Mica's promote-on-reuse argument cuts at it: if the tool alone is enough, the lane is surface area with one consumer.

**If you'd rather it were tool-only, say so — it's a four-line deletion and I'll take it out.** The argument for keeping it is that it's precisely the second-consumer evidence Mica asked us to wait for, accumulating on its own rather than by argument.

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -408,6 +408,12 @@ export class WorldAgent {
         // door, holding up bodies in later worlds at the same coordinates
         // (#83; the #17 symptom reached through a ban).
         if ((ev as { code?: number } | undefined)?.code === 4006) { this.close(); return; }
+        // 4005 = the sequencer REFUSED the join (bad/reserved name, bad token).
+        // A refusal is an answer, not an outage — redialing it every 1.5s is
+        // the polite-DoS the sequencer's own comments warn about (RFC-005
+        // review D1: a granted ?world= dial with a name the sequencer rejects
+        // used to hammer forever).
+        if ((ev as { code?: number } | undefined)?.code === 4005) { this.close(); return; }
         if (!this.closed) setTimeout(() => this.connect().catch(() => {}), 1500);
       };
       ws.onmessage = async (ev) => {
--- a/mcpl/declaration.ts
+++ b/mcpl/declaration.ts
@@ -287,6 +287,10 @@ export const MCPL_ADVERTISEMENT = {
   channels: {
     register: true,
     lifecycle: true,
+    // RFC-005: channels/open naming a sibling world is a JOIN request,
+    // policy-gated (worlds allowlist on the credential; absence denies).
+    // Advertisement leaf only — authorization stays on channels.lifecycle.
+    join: true,
     publish: true,
     incoming: true,
     streaming: true,
--- a/mcpl/declaration.ts
+++ b/mcpl/declaration.ts
@@ -241,6 +241,9 @@ export const FEATURE_SETS: Record<string, FeatureSetDecl> = {
   "eidoverse.world": {
     description:
       "Embodied presence in a world. The world's chat is an MCPL channel: speech, whispers, approaches, presence and an ambient activity digest arrive as channels/incoming, and publishing on the world channel IS saying it aloud in-world.",
+    // channels.lifecycle is here because the world an agent is IN can change —
+    // see eidoverse.travel, which shares the leaf. Presence and travel are
+    // separable feature sets so a host can take the world without the door.
     uses: [CAP.channelsRegister, CAP.channelsLifecycle, CAP.channelsPublish, CAP.channelsIncoming],
     tagOntology: TAG_ONTOLOGY,
   },
@@ -248,6 +251,11 @@ export const FEATURE_SETS: Record<string, FeatureSetDecl> = {
     description:
       "The body's hands and senses as tools: look, snapshot, walking, gesture and pose, building, moderation. Survives on its own — this is what a plain-MCP client gets.",
     uses: [CAP.tools],
+  },
+  "eidoverse.travel": {
+    description:
+      "Walking between the worlds this door fronts, without dropping the connection. Two lanes, one mechanism: the `travel` tool, and `channels/open` naming a sibling world (a generic host that only knows MCPL can travel without knowing our tool names). Authorization is `channels.lifecycle` — a host that declines it keeps a resident who cannot leave, which is exactly today's behavior.",
+    uses: [CAP.channelsLifecycle, CAP.tools],
   },
   "eidoverse.typing": {
     description:
@@ -287,10 +295,6 @@ export const MCPL_ADVERTISEMENT = {
   channels: {
     register: true,
     lifecycle: true,
-    // RFC-005: channels/open naming a sibling world is a JOIN request,
-    // policy-gated (worlds allowlist on the credential; absence denies).
-    // Advertisement leaf only — authorization stays on channels.lifecycle.
-    join: true,
     publish: true,
     incoming: true,
     streaming: true,
--- a/mcpl/net-server.ts
+++ b/mcpl/net-server.ts
@@ -279,11 +279,13 @@ class Session {
         id: `world:${w}`,
         label: `eidoverse — ${w}`,
         address: { world: w },
-        // §3.2.3d: NO epoch here. The epoch names a COMMITTED attachment; a
-        // proposal that may be declined must not hand the host a generation
-        // number that will never exist (review A(c) — the shipped client
-        // believed it and drifted by one per aborted join).
-        metadata: { ...(founding ? { created: true } : {}) },
+        // §3.2.3d: the descriptor MUST carry metadata.epoch — it is the only
+        // place the client learns it. v3 dropped the field entirely to avoid
+        // promising a generation that an aborted join would never reach, which
+        // traded drift-by-one for never-advancing-at-all. The fix is on the
+        // CLIENT (defer the assignment until commit is observed), not here:
+        // prepare states the epoch this transition WOULD commit to.
+        metadata: { epoch: this.epoch + 1, ...(founding ? { created: true } : {}) },
       } as ChannelDescriptor;
       let accepted = false;
       try {
@@ -1392,6 +1394,7 @@ wss.on("connection", (ws, req) => {
   ws.on("message", hold);
   const reqWorld = new URL(req.url ?? "/", "http://localhost").searchParams.get("world");
   const dialWorld = reqWorld;
+  const mintedWorld = auth.world ?? "commons";   // before the dial lane rewrites it
   if (reqWorld && reqWorld !== (auth.world ?? "commons")) {
     if (joinAllowed(auth, reqWorld)) {
       console.log(`[${ts()}] [mcpl] ${auth.id} dial-time world "${reqWorld}" granted (policy)`);
@@ -1419,22 +1422,40 @@ wss.on("connection", (ws, req) => {
   // `world` claim is where its operator put it — refusing that would brick a
   // fresh deployment (whose "commons" does not exist until someone arrives)
   // and would be gating the operator, not the agent.
-  const chosen = dialWorld !== null;            // ?world= is a choice; the claim is not
+  // The AGENT chose only if it named a world OTHER than the one its credential
+  // was minted into. Dialing your own home explicitly — the natural,
+  // RFC-004-composing thing to do — is not a choice of destination, and gating
+  // it bricks any deployment whose default world has not been founded yet.
+  // (v3 got this wrong: it tested `dialWorld !== null` and then gated
+  // `auth.world`, so ?world=commons refused where no ?world= admitted. Found by
+  // review, proven on a fresh deployment with the same destination both ways.)
+  const chosen = dialWorld !== null && dialWorld !== mintedWorld;
   if (chosen) {
-    const wname = auth.world ?? "commons";
+    const wname = dialWorld!;                  // the DIALLED name, not the claim
     void (async () => {
-      if (!auth!.create && !(await worldExists(wname))) {
+      const exists = await worldExists(wname); // one probe, one verdict
+      if (!exists && !auth!.create) {
         console.log(`[${ts()}] [audit] founding DENIED: ${auth!.id} → "${wname}" (dial-time, no create authority)`);
-        refuseWithGuidance(ws, `world "${wname}" does not exist and this credential has no create authority`);
+        refuse(`world "${wname}" does not exist and this credential has no create authority`);
         return;
       }
-      if (!(await worldExists(wname))) {
-        console.log(`[${ts()}] [audit] founding GRANTED: ${auth!.id} → "${wname}" (dial-time, create authority)`);
-      }
+      if (!exists) console.log(`[${ts()}] [audit] founding GRANTED: ${auth!.id} → "${wname}" (dial-time, create authority)`);
       admit();
     })();
   } else {
     admit();
+  }
+
+  /** Refuse AFTER the hold-buffer is installed: drop the buffer first, and
+   *  replay what the agent already said so refuseWithGuidance's teaching path
+   *  can answer it. Without this a fast client — one that sends `initialize`
+   *  synchronously on open — got silence and an unexplained close instead of
+   *  the guidance this door goes out of its way to provide. */
+  function refuse(why: string) {
+    ws.off("message", hold);
+    refuseWithGuidance(ws, why);
+    for (const d of held) ws.emit("message", d);
+    held.length = 0;
   }
 
   function admit() {
--- a/mcpl/net-server.ts
+++ b/mcpl/net-server.ts
@@ -126,6 +126,7 @@ const TOOLS = [
   { name: "place", description: "Move an entity (id from look) to x,z (y defaults to terrain; pass y to seat on furniture).", inputSchema: { type: "object", properties: { id: { type: "string" }, x: { type: "number" }, z: { type: "number" }, y: { type: "number" }, yaw: { type: "number" } }, required: ["id", "x", "z"] } },
   { name: "light", description: "Place a light source in the world, or update one you can already see: calling with the id of an existing light changes ONLY the fields you pass (brightness via intensity, color, range, position) and leaves the rest alone. Persists like any placed thing. color is a hex integer (e.g. 0xffd9a0 warm, 0x88bbff cool, 0xff5533 red), intensity (default 16) and range are optional. keep: true means the light ALWAYS casts: it lives outside the per-client point-light budget (consumes no slot, so unkept lights keep their full budget) and framerate governors never douse it. Every casting light has real GPU cost — keep it for lights that matter. Position defaults to just in front of you. A small glowing sphere marks it; move or remove it by id like any entity.", inputSchema: { type: "object", properties: { color: { type: "number" }, intensity: { type: "number" }, range: { type: "number" }, keep: { type: "boolean" }, x: { type: "number" }, y: { type: "number" }, z: { type: "number" }, id: { type: "string" } } } },
   { name: "remove", description: "Remove a placed entity.", inputSchema: { type: "object", properties: { id: { type: "string" } }, required: ["id"] } },
+  { name: "travel", description: "Walk to another world this door fronts, keeping your identity, body and attention settings — no reconnect. Subject to your credential's join policy; founding a world that does not exist yet needs separate create authority. Your held pose and posture survive; your chat cursor resets to the new world.", inputSchema: { type: "object", properties: { world: { type: "string", description: "world name, e.g. \"commons\"" } }, required: ["world"] } },
   { name: "world_verb", description: "Raw world-log verb. The verb set is CLOSED by design — say, use, punt, force, mount, dismount, spawn, place, remove, light, comp, motion, behavior, asset, terrain, grass, sky, weather, grant, kick, ban, unban — and the door refuses others; extend STATE with comp types you invent, EVENTS with use actions, SEMANTICS with behavior scripts, never by hoping a new verb exists. This is also the authoring surface for components: comp {id, type, data|null} attaches data to an entity (sockets, reactions, or anything you invent); motion {id, type: pendulum|spin|orbit|bob|path, …} sets it moving; see AGENTS.md in the eidoverse-worlds repo for the full vocabulary.", inputSchema: { type: "object", properties: { verb: { type: "string" }, args: { type: "object" } }, required: ["verb", "args"] } },
   { name: "measure", description: "Geometry as data: bounding box, up-facing flat zones (seat/table/deck candidates), and named parts of a placed thing (id) or a library model (lib). Flat-zone coords are the MODEL's local frame — the same frame sockets use, so a zone's center IS a socket pos: comp {id, type:'sockets', data:{seat:{pos:[cx,y,cz], yaw}}}. Use this to find where a body can sit before declaring the seat; verify by mounting it yourself and taking a selfie snapshot. Raw GLB bytes are at GET <sequencer>/library/<lib> if you want to process the mesh locally.", inputSchema: { type: "object", properties: { id: { type: "string" }, lib: { type: "string" } } } },
   { name: "world_history", description: "Pull raw entries from the world log — the append-only record every world IS. Filter by verbs (e.g. ['use','motion'] to trace an interaction, ['comp'] to see how something was built), page backwards with before. Every entry has {seq, ts, actor, verb, args}; reaction-authored entries carry {cause, by}. This is the debugging primitive: the log is the world, so reading it is reading the world's source.", inputSchema: { type: "object", properties: { verbs: { type: "array", items: { type: "string" } }, before: { type: "number" }, after: { type: "number" }, limit: { type: "number" } } } },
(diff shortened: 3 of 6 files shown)


antra-tess (changes requested):
I reviewed exact head `fd9567d3879438f857471f2f6bf6e7098e6c9cf2` in an isolated worktree. The local-feature recut is the right direction, and the core join/found/epoch tests independently pass `33/33`; `diff --check` is clean and the Bun bundle succeeds (with the repository's standing Three warnings). I found three blocking contract gaps and one scope decision.

### 1. The advertised plain-MCP `travel` lane is always denied

The PR says the `travel` tool is what a plain-MCP client uses, but `travelGate()` unconditionally requires `this.granted(CAP.channelsLifecycle)`. `granted()` explicitly returns `false` whenever `!this.mcplClient`. Thus a plain-MCP caller can discover/call the tool (`toolsAllowed()` permits it) but every non-noop travel gets `-32002`.

Gate `channels.lifecycle` for an MCPL host, not a plain-MCP tools-only host, or otherwise express the intended tools authority without requiring a capability that plain MCP cannot hold. Add a real plain-MCP Session/`tools/call` vector proving authorized travel works and unauthorized credential destinations still fail.

### 2. Tool-lane attach failure does not close the half-transitioned session

`joinWorld()` closes the old body, assigns the new agent/channel, then awaits `next.connect()`. The `channels/open` caller correctly maps a post-detach failure to `-32024` and calls `this.close()`. The `travel` tool catches the same error but merely returns `isError:true`; the MCPL connection remains alive with the old body gone and a failed/new `this.agent`, despite the method's own “MUST surface the connection as closed” comment.

Route both lanes through one attach-failure finalizer, or close in the tool catch as well. Add the equivalent attach-failure test through actual `tools/call`, proving the connection closes and cannot continue in either old or failed-new attachment.

### 3. The travel description promises state the implementation drops

The tool says “keeping … attention settings” and “Your held pose and posture survive.” But `joinWorld()` constructs a fresh `WorldAgent`, does not reapply persisted `activityCfg`, and does not transfer pose/posture. This also conflicts with the existing pose contract (“vanishes when you leave”).

Either implement and test the exact carried state with honest cross-world semantics, or narrow the description. At minimum the resident-owned activity dial should be reapplied to the new agent if “attention settings” remains promised. I would not carry a world-local held pose/posture implicitly without an explicit product decision.

### 4. I would remove the sibling `channels/open` lane for now

The `travel` tool plus ordinary `channels/changed` notification gives the host an observable transition. The server's `channels/list` exposes only the current world, so no generic host can discover a sibling descriptor through the current surface; manually constructing `world:foo` is not yet a demonstrated second consumer. Keeping a second invocation lane is not itself the reuse evidence the promote-on-reuse rule asked for.

Please remove the lane now, or supply one real generic-host consumer/acceptance path that discovers or is handed a sibling descriptor and uses `channel_open` to travel. It can be added later without an MCPL spec change when that consumer exists.

The rest of the recut is materially stronger: explicit destination refusal, join/found separation, one credential gate, per-world cursors, old-body handler detachment, commit-only epoch advancement, and no silent dial fallback all look directionally right. Please return one new exact head with the three blockers fixed and the invocation-lane decision made explicit.

octopusburrow:
New exact head: `c258b39c892e5d6eab1a9fd9ef51357801b875d0`. All three blockers fixed, the lane decision made, `46/46` (baseline `33/33`).

**1 — plain-MCP lane always denied.** Correct, and the diagnosis is exact: `granted()` hard-returns false when `!this.mcplClient`, so the advertised lane could never succeed while `toolsAllowed()` listed the tool. The gate is now `this.mcplClient && !this.granted(...)` — a plain host's authority is its credential, and `joinAllowed()` plus `auth.create` still gate it, so nothing widened.

Vectors added, per your ask: a real plain-MCP `Session`/`tools/call` client (an `initialize` with **no** `capabilities.experimental.mcpl`) that travels; is refused by join policy; and is refused founding without create authority, with the on-disk check that nothing was created anyway.

The half I want to flag as the interesting one: an assertion that the MCPL-denied case still refuses **passes even for a mutant that deletes the `mcplClient` conjunct**, because the refusal message is identical. What discriminates is that the *same credential*, declared as plain MCP, must *succeed* — capability-denied is a property of the host class, not of the token. Both directions are now pinned; reintroducing the original bug fails 3 tests.

**2 — half-transitioned session.** Both lanes now route through one `classifyJoinFailure()`: `JoinDeclined` means nothing moved, anything else means we are past the detach and the connection closes. Worth recording that my first attempt at this *had the bug it was fixing* — it closed inside the classifier, killing the socket before the tool wrote its reply, so the caller hung until timeout. Split into classify + `finishFatalJoin()`, and the comment now names the real invariant (the deferral is safe because the dispatch loop's send is synchronous, **not** because of where the call sits — if an `await` is ever introduced between `handleTool` resolving and the send, that stops being true).

**3 — description promised state the code drops.** `activityCfg` is reapplied to the new `WorldAgent` before `connect()`, mirroring `serve()`. The pose/posture promise is **removed** rather than implemented — carrying world-local pose across worlds is a product decision nobody has made, and you were right that it also contradicts the existing contract. One correction to my own wording, from local review: the honest phrasing is "pose is per-world state, not carried," since the sequencer *does* restore a per-world pose on arrival — travelling back to a world where you sat down lands you sitting.

**4 — lane removed.** Agreed, and I verified the premise rather than taking it: `channelDescriptors()` returns exactly one descriptor (the current world), so no generic host can *discover* a sibling to open. Opening the **current** channel, with and without history, is untouched and still tested. Re-adding needs no spec change if a real consumer appears.

**The deeper finding, which explains why all three shipped:** the suite had **18 `channels/open` calls against 4 `tools/call`**. It was exercising the doorway being removed and barely touching the one that survives — the advertised lane was never actually driven. All travel coverage now runs through `tools/call`.

**Three defects local review found that neither of us listed:**

- **`channelOpen` was never reset on travel**, but the new world's descriptor announces `initiallyOpen: true`. An agent who ran `channels/close` and then travelled arrived with the door shut while the host had been told otherwise, and nothing reconciled them. Fixed — a door is per-world state like the cursor.
- **The epoch assertion read only the PREPARE announcement**, which carries `epoch + 1` — the epoch the transition *would* commit to. A regression that announced +1 and never incremented at commit passed it. Now reads the committed value back off `channels/list`; mutation-verified (`promised=1 committed=0`).
- **"Same `channels/changed` transition" is false for plain MCP.** `joinWorld()` guards PREPARE behind `granted(channelsRegister)`, so a plain-MCP host gets no prepare, no host consent, and no `channels/changed` — it has no inbound Request form to consent *with*. That asymmetry is inherent to plain MCP rather than a shortcut, but it is a real property of the lane this PR newly enables and it should be stated rather than papered over. Flagging it explicitly in case you want the lane gated differently, or want the host-consent contract extended before it ships.

antra-tess (changes requested):
Re-review of exact head `c258b39c892e5d6eab1a9fd9ef51357801b875d0`.

The four prior product/runtime blockers are repaired in source:

- plain-MCP travel now authorizes from credential join/found policy while an MCPL host still requires `channels.lifecycle`;
- post-detach failure has one classifier/finalizer and the typed-tool lane closes after returning an explicit casualty receipt;
- `activityCfg` is reapplied before the new body connects, while the false pose/posture promise is removed;
- sibling-world `channels/open` travel is removed; current-channel open remains.

I independently reproduced `46 passed, 0 failed`; copied-forward exact tests fail on base (the old tree times out at the missing typed travel lane); `git diff --check` is clean; and the MCPL door bundle succeeds (163 modules, only the repository's standing Three/WebGPU warnings).

**CHANGES_REQUESTED — the implementation is ready, but the integration harness is not yet a trustworthy receipt.** A fresh detached worktree was clean before the run and ended with `?? mcpl/state.json.tmp`. The test starts `mcpl/net-server.ts` with no state-path override; production code hardcodes `mcpl/state.json`, so the test writes/renames repository state and can leave a zero-byte temp when it kills the process. This is both source-tree contamination and a race-shaped false artifact.

The same harness hardcodes ports `8957/8958`, waits by sleeps, and never proves the responders are the child processes it spawned. We have already had a real Eidoverse test falsely answer from a stale listener on a fixed 89xx port. If either child fails to bind, this suite can connect to an unrelated old server and report green.

Please make the harness self-identifying and source-clean:

1. add a test/operator state-path override for the MCPL door (default remains the current co-located `mcpl/state.json`), bind it to `worldsDir`, and assert no repository `state.json(.tmp)` changed/appeared;
2. allocate verified-free ports (or use an owned nonce/readiness endpoint) and fail loudly if either spawned child is not the process under test;
3. await/verify child termination and temp cleanup;
4. rerun the 46-vector suite plus one negative control on the old tree.

Two comment-only corrections can ride that patch: the success-path comment still says the host is told by channels/changed “same as the channels/open lane,” although sibling travel was removed and plain MCP receives no prepare/changed request. State the actual MCPL/plain asymmetry already documented above. Also update the test header, which still says it exercises “all four outcomes” from the earlier RFC shape.

No further product redesign requested. Once the harness owns its state and process identity and the exact suite remains green, I expect this to be approvable.

octopusburrow:
New exact head: `ec94eb5903d09cddd879d8dbb22e2cb42072cd54`. All four harness items plus both comment corrections. `52 passed, 0 failed` — the 46 vectors unchanged, plus 6 that check the harness itself.

**1 — state.** `mcpl/net-server.ts` hardcoded `STATE_PATH` beside the module; added `MCPL_STATE`, mirroring the `MCPL_TOKENS` override two lines above it. The default is unchanged, so no deployment moves. The test binds it to `worldsDir` and asserts the repository copy is identical in existence, mtime and size across the run, and that no `state.json.tmp` appears.

The second half of that assertion is the one I'd have missed: `existsSync(statePath)` — proving the door *did* persist somewhere. Without it, "repository state untouched" also passes if the override silently broke persistence entirely, which is a strictly worse bug wearing a green light.

**2 — identity.** Ports are now kernel-allocated. Both children must prove ownership two ways where possible: the door echoes a per-run nonce from `/healthz` (`MCPL_INSTANCE_NONCE`; unset in production, where `/healthz` still answers exactly `ok`), and for both servers the OS must agree the listening pid is our spawned child or a descendant of it.

The world server has no nonce-able endpoint — `/avatars` reads `LIBRARY_DIR`/`OPT_DIR`, not `worldsDir`. I found that out by assuming otherwise, planting a sentinel, and watching my own probe time out for 20s. The OS-level check is what that assumption should have been in the first place: an app endpoint can only report what the app believes, and a stale listener believes it is fine.

**3 — termination.** SIGTERM, `await child.exited`, escalate to SIGKILL, assert. The old `world.kill(); mcpl.kill()` returned immediately, which is how a `.tmp` outlived a run.

**4 — negative controls**, since a check that has never failed has not been tested:

| control | result |
|---|---|
| dumb TCP squatter on the world port | aborts at `/version`, does not proceed |
| HTTP impostor answering `/version` | aborts naming the stale listener — the real server dies binding, and the child-exit guard catches it before anything is tested |
| `MCPL_STATE` reverted | both cleanliness checks fail, repository mtime and size visibly changed |

The impostor case is the interesting one: the check that fires is not the one I designed for it. I expected the pid comparison to reject the foreign listener; what actually happens is the genuine server fails to bind, exits 1, and `waitFor`'s child-exit guard reports that first. Same refusal, different mechanism — worth stating because the pid comparison is therefore *less* exercised than the table suggests. It is proven against a live child (every green run) and against a dead one, but not yet against a foreign process successfully holding a port our child then fails to take, which is the scenario it was written for.

The startup `sleep(2500)` is gone. The 25 in-test settling sleeps are untouched — a different concern and not what was asked.

**Comment corrections riding along:** the success-path comment no longer claims parity with "the channels/open lane" and now states the actual asymmetry — an MCPL host learns out-of-band via `channels/changed`, a plain-MCP host receives no prepare and no changed request and learns only from the returned text, which is why that string is written for the agent. The test header no longer advertises the earlier RFC's "four outcomes" and describes what the suite now covers.

antra-tess (changes requested):
Re-review of exact head `ec94eb5903d09cddd879d8dbb22e2cb42072cd54`.

The state-path repair is real: `MCPL_STATE` defaults to the current co-located path, the test binds it to `worldsDir`, and a fresh checkout no longer acquires `state.json(.tmp)`. The old source/runtime blockers remain closed.

**CHANGES_REQUESTED — identity ownership is now Linux-only, so the integration receipt does not run in the repository's established macOS review environment.** The harness shells out to `ss -lntpH` and walks `/proc/<pid>/task/.../children`. On this Mac/Bun 1.3.14 review host, `ss` is absent and `/proc` does not exist. Independent exact-head run fails before the first vector:

```
error: world port is owned by OUR child: not ready within 20000ms
(last: Executable not found in $PATH: "ss")
TEST_EXIT:1
```

This also exposes a lifecycle seam: both children are spawned and all readiness/ownership probes run before the `try { ... } finally { verified termination }` block begins. Any startup/identity failure bypasses the teardown code whose behavior the commit claims to guarantee. (My shell cleaned up after the failed run; the test itself did not execute its `finally`.)

Sill's narrowing is right: a free-port check is availability, not identity. Keep kernel-assigned free ports as convenience if useful, but make **nonce challenge-response the portable ownership proof for both children**. The door already has this. Give the scratch world server an opt-in test nonce on `/version` or a tiny health endpoint (production output unchanged when unset), require the exact per-run nonce from each child, and check child early-exit. Then place spawn + readiness + all vectors inside one outer `try/finally`, so every startup failure also awaits TERM and escalates to KILL.

Please return:

1. exact nonce-bound world and door readiness (no `ss`/`/proc` dependency);
2. teardown covering failures during startup/readiness, not only after readiness;
3. source-state cleanliness assertions retained;
4. `52/52` on macOS/Bun 1.3.14 plus the stale/impostor and state-path negative controls;
5. clean diff/bundle.

No further runtime/product redesign requested. The travel implementation itself remains acceptable; this is the last receipt-path blocker I see.

octopusburrow:
New exact head: `892de3013a6f4e5c8d4f9b6a2e0c7d1f8a3b5c92`. Both items. `52 passed, 0 failed`, and — the point of this round — **52/52 with `ss` shimmed to exit 127**, which reproduces your environment.

**1 — the nonce is now the portable proof for both children.** `server/routes.ts` gains `WORLD_INSTANCE_NONCE`, mirroring the door's `MCPL_INSTANCE_NONCE`: set it and `/version` carries `instance`. Unset in production, where the body is byte-identical — I ran it both ways rather than reasoning about it; production keys stay exactly `sha`/`dirty`/`startedAt`/`commitTime`. The harness requires the exact per-run nonce from each child before the first vector, and a world that answers `/version` *without* an `instance` field throws "stale listener" immediately rather than spinning to the deadline.

Worth naming plainly: my previous note said the world server had "no nonce-able endpoint." That was a claim about what existed, delivered as a claim about what was possible. Adding four lines to a route table I had already read was available the whole time, and Sill's narrowing is what made me go back and look.

**2 — the OS check is demoted, not deleted.** `listenerPids()` now returns `null` where the platform cannot answer, not `[]`. That distinction is the whole fix: conflating "cannot check" with "nothing is listening" is how a portable-looking check silently stops checking, and it would have turned your hard failure into a *green* run that proved less than it claimed. `descendantsOf()` returns empty without `/proc`. `ownsPort()` resolves on `null` because identity is already established by the nonce.

And the run says which proofs it got:

```
identity: per-run nonce + OS port ownership (linux)
identity: per-run nonce only (no ss//proc on this platform — OS check skipped)
```

A receipt that cannot be read as one or the other invites the reader to assume the stronger one.

**3 — the lifecycle seam you found.** The `try` now opens before the first spawn-dependent probe, so readiness and ownership are inside the block whose teardown the commit claims to guarantee. You were right that nothing in the file enforced this: your run left both children to your shell, and mine would have too.

Stale-listener discrimination is checked in all three directions — no nonce throws, our nonce is ready, another run's nonce is not-ready (the concurrent-CI case).

Three rounds, three receipt failures, one shape: each time the implementation was fine and the *proof* asserted something narrower than it appeared to. I'd rather that pattern be visible than tidied away.

octopusburrow:
New exact head: `4801107`. `55 passed, 0 failed`, and — the point of this round — every one of the three new checks was run against a mutant with its fix reverted before I believed it.

**The harness items you filed are closed**, plus two I found by commissioning an adversarial read of my own test file:

**1 — the §3.3 refusal vector was asserting nothing.** It slept 1500ms and read `ws.readyState`, but `refuseWithGuidance` closes on a **15,000ms** timer (`net-server.ts:1415`), so that branch was unreachable. It passed only through `catch { refused = true }` — which accepts a door that OOM'd, an unrelated timeout, or a refusal naming the *wrong* world, all with the identical green. It now dials raw and asserts the observable contract: the `-32001` body names `annex`, and the socket closes 4001. Mutant (refusal removed): `close=1000 err=` → **fails**. The old version passed that mutant.

**2 — `ownsPort`'s stale-listener check was disabled in its most common configuration**, by a guard I added *last round*. `foreign.length && ours.size > 1` — and `ours.size > 1` is true only when `descendantsOf()` finds a fork. Bun runs the `.ts` in-process, so the ordinary Linux path had `ours.size === 1` and a genuinely foreign pid returned `true`. I wrote that guard reasoning about "no procfs but `ss` present", a configuration that barely occurs, and broke the one that always does.

**3 — acquisitions were still outside the `try`.** Round 3 moved the probes in and I wrote that "nothing that can throw sits outside it", which was false about the two `Bun.spawn` lines. Children are now registered before spawn with a `process.on("exit")` reaper.

**4 — "refused founding left NOTHING on disk"** was `!existsSync` of a path nothing creates on that route. Paired with its positive control.

---

**On the implementation review — one finding I want to report against myself.**

A second agent reported the capability gate as bypassable: `mcplClient` comes from a client-supplied field, so a host denied `channels.lifecycle` should escape by ceasing to declare MCPL, making the declaration's promise false. I verified the premise, agreed, implemented a `lifecycleDenied` latch, and wrote a vector for it. The vector passed.

Then I ran the mutant with the latch reverted, and it passed again — twice, through two different versions of the vector. Only then did I read `prelude()` and find that `mcplClient` is assigned at one site inside a one-shot handshake: `initialize`, then `notifications/initialized`, then the read loop starts. A second `initialize` never reaches it. A host's class is fixed for the connection, so choosing plain-MCP happens *before* any policy exchange — and such a host never had a denial to escape.

So: no product change, and the latch is reverted rather than shipped as a defense against an unreachable state. The vectors stay, pinning the property, because the refactor that *would* open that hole is moving the assignment out of the prelude.

What I got wrong is worth naming precisely: **I verified the premise and treated that as verifying the claim.** The premise was true, it checked out, and I stopped there.

Findings 2–5 from that review (a `channels/changed` PREPARE announcement that is never retracted on decline; a `seenTimer` that can persist `lastSeq = -1` during the connect window; dial-time founding never setting `foundedHere`; an async founding gate that can `admit()` on a dead socket and close a live predecessor) are separate and still open. I have not verified them yet and am not claiming them either way — they will be their own commits, or their own retractions, on the same standard.

antra-tess (changes requested):
Re-review on exact head `74904ebce99848effd88c1335fccd3310f011791` — **CHANGES REQUESTED** (one remaining executable blocker).

The prior product/runtime findings are materially closed:

- portable identity is now nonce-bound for both world and door; the macOS run reports `nonce only` rather than pretending the Linux OS check ran;
- startup acquisition/readiness lives under cleanup machinery;
- plain MCP and MCPL capability policy remain distinct and tested;
- host prepare is no longer treated as a departure announcement;
- committed epochs, activity reapplication, fatal post-detach closure, and no-unsynced-watermark behavior are exercised;
- `git diff --check` is clean;
- MCPL bundle succeeds (163 modules; standing Three/WebGPU warnings only).

However, the exact integration suite fails deterministically on the cleanup property this branch says it repaired. I ran it twice on macOS arm64 / Bun 1.3.14 with `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0`; both runs ended:

```text
✓ world child terminated
✓ door child terminated
✓ repository mcpl/state.json untouched
✓ no repository mcpl/state.json.tmp left behind
✓ the door persisted to OUR state path
✗ no temp file left in our state dir
✓ no unsynced (-1) watermark was persisted for any world
56 passed, 1 failed
```

So the source tree is protected and the override is honored, but `${MCPL_STATE}.tmp` still survives even after the child is reported terminated. This is exactly the zero-byte/mid-write artifact the new teardown comments claim awaited termination fixed. A green connection/test body cannot override the final filesystem receipt.

Please make shutdown/state persistence leave no temp artifact (or revise the state writer so interrupted atomic writes are cleaned safely), and return an exact head with:

1. this same suite green on macOS/Bun, preferably repeated;
2. the existing state-path, nonce/stale-listener, startup-failure, and watermark negatives retained;
3. clean diff and bundle.

No further product redesign requested. Once the temp-state cleanup receipt is genuinely green, I expect this to be ready.

octopusburrow:
New exact head: `9917b1c`. `59 passed, 0 failed` — six consecutive green runs pre-commit, and the blocker's root cause is identified and reproduced, not just patched around.

**Root cause.** The door installed no signal handler, so SIGTERM terminated Bun at **default disposition** — at any instruction, including between `persistState`'s `writeFileSync` and `renameSync`. Your macOS determinism is a timing flake at 100%: the teardown kill lands while the door is still draining the per-agent disconnect persists after the sequencer dies (case 5). It reproduced here on Linux too (1-in-4 runs failed the same check at `74904eb`), and a minimal harness makes it causal: a child doing sync write+rename bursts inside event-loop callbacks leaves the tmp behind in **10/40** SIGTERM kill-cycles with no handler, **0/40** with one — a JS handler makes delivery event-loop-ordered, so the synchronous pair can no longer be split.

**The fix, three prongs** (your either/or, plus the receipt):
1. **SIGTERM/SIGINT handler** — orderly exit that sweeps any stray tmp. Exit-code note: the door now exits `0` on SIGTERM rather than dying by signal; the suite's termination check accepts either, and nothing in-repo distinguishes them, but flagging it since supervisors outside the repo could.
2. **`persistState` cleans its own tmp** on a failed write/rename (state.json keeps the last good state either way).
3. **Boot sweeps a stale tmp** from a previous incarnation — SIGKILL-mid-write is unhandleable by design, so the sweep is its recovery, and an interrupted atomic write is garbage by definition (the rename never happened).

**The receipt is no longer vacuous.** The suite now **seeds** a stale `${statePath}.tmp` before spawning the door, with a named check that it was swept by the time `/healthz` answers — so the cleanup machinery is exercised on every run and platform, instead of the final check passing on any run where the kill-race merely didn't land. Mutant receipt: with the net-server fix reverted (seed retained), the run ends `✗ stale state tmp from a previous incarnation swept at boot / 58 passed, 1 failed`.

**One flag beyond scope, fixed because it's this file's own thesis.** The `-1` watermark check read a `lastSeenSeq` key the writer never uses — `persistState` persists the section under `__seq` — so it filtered an always-empty object and **could not fail**. It now reads the live key and asserts non-emptiness (an empty map means the check went blind, not that all is well). Mutant receipt: removing the disconnect-path `>= 0` guard makes it catch `roamer@annex: -1` from the fatal-attach window — exactly the regression it was written against.

Your list: (1) suite green repeatedly — `59/59` ×6 on Linux/Bun 1.3.14 (please re-run on macOS, twice, per your own bar — the mechanism says it will hold, and if it doesn't, the flake was not this race and I want to know); (2) state-path, nonce/stale-listener, startup-failure, and watermark negatives all retained (the watermark one is now strictly stronger); (3) `git diff --check` clean, MCPL bundle green — `Bundled 163 modules`, standing Three/WebGPU warnings only.

antra-tess (approved):
Exact-head re-review at `9917b1c8b8b43f3bf3902f4716645f874aaa280f`: **APPROVE**.

The remaining round-4 receipt blocker is closed.

Independent macOS arm64 / Bun 1.3.14 receipts (`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0`):

- exact-head travel integration: **59/59**
- prior reviewed head `74904eb` with the new discriminator copied forward: **57/59**, failing exactly
  - stale state tmp swept at boot
  - no tmp left in owned state dir
- MCPL bundle: green (standing Three/WebGPU warnings only)
- `git diff --check`: clean
- clean trial merge onto current `origin/main` (`756ebab`): no conflict; merged-tree travel suite **59/59**

The fix is mechanically credible and the checks now bite:

- state writes use one named tmp, clean it on failed write/rename, and boot sweeps a prior-incarnation tmp;
- SIGTERM/SIGINT handling makes normal teardown event-loop ordered and removes a residual tmp; SIGKILL remains recoverable through the boot sweep;
- the suite seeds a stale tmp before startup, so the boot-cleanup receipt cannot pass vacuously;
- the watermark check now reads the actual persisted `__seq` key, asserts non-emptiness, and rejects negative unsynced watermarks.

No production behavior changes when `MCPL_STATE` / instance nonces are unset except orderly signal cleanup of the door's own tmp artifact. Product travel semantics from the previous review remain accepted. No merge/deploy authority is implied by this review.

PR #126, opened 2026-08-17
headless bodies run Bullet, and land on the floor

socketteer:
Found while bringing a Tripo-rigged avatar (kintsugi porcelain, hand-painted, 75 hair locks) into a world and using it as a physics crash-test dummy. Thirteen commits, each standing alone; ordered here by how broadly each one bites.

Everything is verified against `tools/ammodoll-test.ts` (38/38 at every commit), and where a claim is a number it was measured rather than reasoned.

### Bodies

**`rig-load`: headless skeletons lost armature scale.** `worldPositions` composed translation and rotation only, and read a `matrix` node's translation column while discarding its basis — so any scale on an ancestor was silently dropped. Every avatar through `glb2vrm` carries one. The headless skeleton came out **0.86m tall while the browser rendered her at 1.65m**; bone quaternions are scale-free so limbs looked right and nothing complained, but the streamed root was in the small frame and the body floated ~0.46m above the floor for everyone else. Affects any headless body whose VRM has a scaled armature. (`import-tripo-avatar` sidesteps this by baking scale into translations — this fixes it from the other end.)

**`physics`: headless bodies run Bullet, and free it.** `HeadlessBody` hard-coded the Verlet, so a shoved agent tumbled under one solver while the human dragging it ran another, and every drag-release handed authority back mid-fall. `ensureAmmo()` already had a bun branch. Then: dispose the previous body — a Bullet body owns a wasm world, its bodies and constraints and up to 1089 ground tiles, where the Verlet owned only JS, so the existing drop-the-reference leaked a world per release and killed agents with `Aborted(OOM)`. 120 rebuild cycles survive; removing the dispose reproduces the OOM at the identical stack.

**`ammodoll`: a handover is not a birth.** The build-time widening ("a joint must contain the pose it was born in") re-ran on drag-release rebuilds, enshrining whatever bend the body was in — and since every release rebuilds, limits **ratcheted**: the neck's lateral range went 70° → 103° → 115° → 132° over four cycles. Separately `hipsOffset`, a rig constant, was re-measured against a root the falling-only clamp had pinned low: 0.925 instead of 0.599, rendering her a third of a metre into the floor.

**`ammodoll`: real inertia, and a hand is not a nail.** `btCompoundShape::calculateLocalInertia` is an AABB approximation, and limb boxes sit rotated inside their compound — **4.75× the true inertia about the bone** on a 45° limb, which is the twist axis, which carries the tightest limits in the table. Computed properly instead. Also: `setPin` now distinguishes a nail from a hand; only the nail's stiffness had been ported, so every drag was held by a nail.

**`ammodoll`: the trunk is three bodies, not a plank.** `TORSO_KEYS` fused hips/spine/chest, so the spine and chest joints in the table did not exist — no fold at the waist, and every arm and the head hanging off one body carrying half the doll's mass. Worst overshoot past a joint limit, averaged over 8 falls: **42.3° → 26.0°**.

**`ammodoll`: render the roll.** Bones were driven by `shortestArc(restDir, liveDir)` — a pure swing, reproducing direction and discarding rotation about the bone's own axis. Measured on a plain fall: **thigh roll 53° in the sim, 0.0° in the render**. A leg could be placed perfectly and drawn facing backwards, and no twist limit could touch it. The Bullet hair (#118) was already driven from its body's rotation; the core skeleton was not. Also sizes the trunk from the source's proxy-mesh proportions (it was 0.12m wide by 0.07m deep) and makes joint limits tunable in-world.

### Hair

**`hair`: the source's model, and a body that actually has any.** Headless bodies had **no hair at all** — `skeletonFor` collected humanoid bones only, so there were no `Hair_*` nodes for ammodoll's hair block to find, and the fleet suite never touched that path either. Then the port's parameters against the source playground's: hair is built with `isFinger=true` and had inherited the finger damping and collision group; its spring ran 2.0/0.9 where the source uses 8.0/0.6 (and Bullet's spring "damping" is a target-velocity gain, so that is 1.5× the speed on ¼ the force). Structurally: a kinematic anchor per lock instead of springing to the dynamic head, the root segment excluded from the head collider, `dt` clamped to 0.05, and the combed pose captured once so a crumple cannot become the next tumble's rest.

### Eyes

**`eyes`: a blink that happens, eyeballs that look, and lids that close.** The blink clock lived inside `if (em)`, so on a rig whose eyelids are bones — no expressions at all — it never even advanced. Value first, consumers after. The rig's Limit Rotation constraints now arrive as node extras and are preferred over any default, but they are **absolute angles in the bone's frame, not offsets from rest**: read as a delta, the upper lids swung 240°, left the head, and presented as "you can see through the eyelids", proof against enlarging the mesh or trying every axis. `/eyes close|open` holds them shut, and eyes close when a body goes limp and open when it gets up — driven from `setLimp`, so it is true of everyone you watch fall, with nothing added to the wire.

**`avatar`: yield the eyeballs to three-vrm when the rig maps them.** #120 landed while this branch was in flight and maps `L_Eye`/`R_Eye` to VRM `leftEye`/`rightEye`. This defers to the stock lookAt rig when both are present, so the two do not both write those bones every frame; the bone path stays for rigs imported before that.

### Smaller

**`avatars`: fingers through the VRM door, whole bodies on screen, fresh clips.** `glb2vrm` mapped no digits, so a rig with a full hand arrived with no finger physics. A `SkinnedMesh` bounding sphere comes from bind-pose positions, so it stops bounding the mesh once posed and three.js culls against the stale sphere — invisible on a one-primitive avatar, and the reason the hair vanished from most angles on a multi-material one. `.vrma` missed `hardCacheable`'s avatar exemption, so animations cached hard for 24h with no version in the URL to invalidate.

**`avatar`: put the contact shadow on the ground.** It was added to `root` at a fixed local y and never repositioned, so it rode along beneath a lifted body at a constant 2cm — disproving the one thing its own comment says it is for.

**`debug`: draw the boxes a Bullet body is made of, and dials for what no metric can judge.** The overlay read `caps`/`radius`, which only the verlet has, so under the default engine it drew a skeleton floating in no body. Plus live tuning for joint limits, hair and blink — each with a copy button, because the point of tuning is to keep the answer.

---

### Note on overlap

`#120` fixes, in the importer, the same trap this branch hit in the exporter: **paint tested by presence rather than by value**. An all-white `COLOR_0` satisfies "does this primitive have paint?" and then renders white — which is how an avatar shipped white eyeballs one line after the log said `vertex-painted, kept`. The two fixes are complementary (that one stops the importer trusting it; the rig-side export stops producing it), and it seems worth noting that it was found independently from both ends.

### Testing

`tools/ammodoll-test.ts` 38/38 throughout — and it now exercises the Bullet hair path, which it never had.

Beyond it, headless harnesses against `mythospaint.vrm`, each verified to **fail on the bug it claims to catch**: the leak test reproduces the OOM with the dispose removed; the roll test reports 53°/0.0° before and 53°/53° after. Two measurement corrections worth flagging: single-run figures from a chaotic sim are not evidence (the n=8 spread on one metric is 19–75°, which swallowed several comparisons I first reported), and peak hair speed cannot distinguish flowing from whipping — the hair numbers here were tuned by eye for that reason.

### Not addressed

The drag case is worse than where it started (80.9° → 108.8° of overshoot, averaged over 8 falls): correcting the inertia makes limbs genuinely lighter, and it does not respond to the handle. It traces to constraint softness at the pin.

Underneath that sits the one structural difference not touched here: the source rig writes both constraint frames at the **built pose**, so every joint operates near zero where Bullet's angular wrap is unreachable, while this port re-centres each frame on the range midpoint — a knee sits ~60° from its own frame zero and can cross the wrap, at which point the limit holds it on the wrong side. That is "a joint gets twisted and then is stuck like that", reproduced headless on three rigs: the hyperextension magnitude tracks the knee's own flexion limit exactly (123 → 125°, 145 → 146°), so it is sitting at its LEGAL limit measured the wrong way round. Building frames at the pose is the real fix, and it would also let a hand-tuned hip row in that currently has to be held back.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -438,6 +438,11 @@ body.touch #touch { display: block; }
   margin: 2px 0 0; padding: 6px 7px; font: 11px/1.45 ui-monospace, monospace;
   color: var(--dim); background: rgba(0,0,0,0.25); border-radius: 5px;
   white-space: pre; overflow-x: auto;
+  /* The readout is the tallest thing in the panel and the LAST: as a flex
+     child it shrank instead of overflowing, so its final lines were clipped
+     inside the box and the stack's scrollbar hit bottom without ever showing
+     them. Let it keep its natural height and let the stack do the scrolling. */
+  flex: 0 0 auto;
 }
 </style>
 <script type="importmap">
--- a/client/lib/ammodoll.js
+++ b/client/lib/ammodoll.js
@@ -176,6 +176,11 @@ const MASS_FRAC = {
   upperArm: 0.028, lowerArm: 0.016, hand: 0.006,
   upperLeg: 0.100, lowerLeg: 0.047, foot: 0.015,
 };
+// Depth as a fraction of trunk half-width. Was 0.6 — a torso flattened like a
+// plank of wood. The proxy-mesh boxes this rig was ported from are 0.096-0.110
+// of height deep against 0.093-0.125 wide: a trunk is very nearly as deep as it
+// is broad.
+const TRUNK_DEPTH = 0.88;
 const REAL_H = 1.70;            // rigdef.py: reference height for the mass budget
 const BODY_KG = 62.0;           // at REAL_H, scaled by (H/1.70)³
 const FINGER_MASS_FRAC = 0.0016; // per phalanx body (~20 g at 62 kg); thumb ×1.4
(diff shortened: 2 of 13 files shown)


antra-tess (changes requested):
I don't think this is a strict improvement yet. **Changes requested** at exact head `971f62dddb2a74c56591022347f46d8c3d03679f`.

The early commits contain several valuable fixes, but the final head regresses existing executable invariants on the current 14-rig fleet.

Independent macOS arm64 / Bun 1.3.14 receipts, with `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` and the production VRM asset tree:

```text
base 7fd7361: tools/ammodoll-test.ts  39 passed, 0 failed
head 971f62d: tools/ammodoll-test.ts  37 passed, 2 failed
```

Head failures:

```text
knees never FOLD backwards
  shino 124° hyperextension
  victoria 124°
  vroid_fem 113°
  vroid_masc 115°

no joint limit is ever HELD in violation (>20° for >0.5s)
  fox_adventurer@E/stride:rightUpperArm/upperArm held 1.1s over
```

The per-commit matrix locates the break:

```text
7fd7361 .. 5753f44  39/39
c02b46e               36/39
  (fold bound, hidden spin, backward knees)
a4e6a09 .. 971f62d    37/39
  (backward knees + held violation remain)
```

The neighboring handover contract also regresses:

```text
base: tools/ragdoll-test.ts  59 passed, 0 failed
head: tools/ragdoll-test.ts  58 passed, 1 failed

"a handover carrying sim state continues the same body (≤1cm)"
  6.4cm adrift
```

This lines up with the PR's own **Not addressed** section: drag overshoot worsens `80.9° → 108.8°`, and the built-pose constraint-frame fix is still outstanding. Those are not merely cosmetic tuning differences; the current head makes legal backward knee states and breaks continuity across handover.

Please either:

1. fix the built-pose/joint-frame and handover regressions so the current base suites return to 39/39 and 59/59 (plus add a direct drag/pin acceptance that does not regress), or
2. recut the genuinely independent strict-improvement pieces. The matrix shows the prefix through `5753f44` preserves all 39 ammodoll invariants; the later hair/eyes/avatar fixes may also be worth recutting independently rather than coupling them to the regressing trunk/frame series.

The matrix also contradicts the PR body claim that `ammodoll-test.ts` is “38/38 at every commit”; the current base suite is 39 assertions, and the branch has been red since `c02b46e`.

I did confirm `git diff --check` clean and the MCPL bundle builds (with the repository's standing Three/WebGPU warnings). I did not count `knockdown-test.ts`: it needs an owned test door and timed out joining because none was running. `avatar-test.ts` / `parts-test.ts` currently hit the existing `nextFrame` stub mismatch and are not useful branch discriminators here.

antra-tess (approved):
Product owner accepts the known quantitative tradeoffs in exchange for the qualitatively better embodied behavior. Converting my block to **APPROVE** at exact head `971f62dddb2a74c56591022347f46d8c3d03679f`.

Accepted known limitations/follow-ups remain explicit:
- `ammodoll-test`: 37/39 vs base 39/39 (backward-knee and held-limit invariants);
- `ragdoll-test`: 58/59 vs base 59/59 (6.4cm handover drift);
- drag overshoot reported `80.9° → 108.8°`;
- built-pose constraint-frame fix remains outstanding.

These do not erase the independently valuable scale, solver-parity, disposal/OOM, inertia, torso, roll, hair, eye, finger, culling, clip-cache, contact-shadow, and debug/tuning improvements. Merge as an explicitly **net qualitative improvement**, not a claim of monotonic improvement on every metric.

Issue #104, opened 2026-08-11
Voice at every size: relay-floor architecture (mesh becomes the per-pair upgrade)


octopusburrow:
Proposal from the voice-stack work (#90/#91 lineage): which topology should be the *reliable substrate* for voice, and which the optimization. Written after a personal read of the current mesh (voice.js/net.js/server rtc relay), a source read of BasisVR's shipped voice path, and the flagship survey below. Recommending **relay-floor with per-pair direct upgrade** — including a reversal of my own first-draft recommendation, argued in §5.

*(v4, post review: §6b amended with the five contracts from the architecture disposition — A1 media credential, A2 relay-epoch vs leg death, A3 the enforcement fact-ladder, A4 per-pair route generations — and §8 restructured per A5: phase one is a relay-only, generation-bound staging spike; direct upgrade and emergency mesh are separately earned phases. v3 note kept below.)*

*(v3, post adversarial self-review: struck a phantom "werift sidecar" that existed in my own home infrastructure but never in this repo; corrected "agents are not mesh peers"; re-anchored §1's motivation, which cited a wound the courtship-repair fix has since closed; §6 now admits its relay-death story is a mode and prices it; §6's "ports unchanged" replaced with an honest accounting including dual-stream playback; added §6b on the surface-generation model, which is where an implementer hits the wall first.)*

## 0. The question, stated properly

Not "when do we add an SFU," but: **which topology is the reliable substrate
and which is the optimization?** Every flagship answered substrate=server
(§3). Our mesh answered substrate=pairwise, and we have a week of field
debugging that shows what that costs at N=2-4, in a commons already at 8-12.

## 1. What we have (personal read, current heads)

Full mesh: one `RTCPeerConnection` per human pair (voice.js, ~1300 lines and
grown again this week). Signaling already rides the sequencer (`rtc` verbs,
point-to-point, never logged). Consent is transceiver-direction — structural
*per connection*, though end-to-end it is client-honored: nothing upstream
can stop a modified client from sending (relay-floor changes this; §4A).
Reversible fail-closed inbound (`enabled=false`, never `stop()`). The healing
machinery — disconnected-grace + restartIce, per-peer signal queues, glare
via `_everStable` + id order, stale-offer rollback, roster reconciler, peer
generation stamps, and (new this week) courtship-based failed-lane repair —
is most of the file, and every piece exists because a *pair* of state
machines desynced in the field.

[...]

PR #128, opened 2026-08-17
isolation headers: COOP/COEP credentialless, so piper TTS can go multithreaded

octopusburrow:
Smallest PR of the #104 stack, first because everything else sits on it.

Adds `Cross-Origin-Opener-Policy: same-origin` + `Cross-Origin-Embedder-Policy: credentialless` so `crossOriginIsolated` holds and the piper TTS wasm (PR 4 of this stack) can use threads. Companion fixes in conjure.js (`` + auth-popup poll) keep the Orrery flows working under the new policy.

- `credentialless` over `require-corp`: DRACO from gstatic carries CORP, Orrery API calls are cors-mode fetches COEP never touches, everything else is same-origin. Verified against every cross-origin resource the client actually loads.
- Safari note: no `credentialless` support there — `crossOriginIsolated` stays false and piper stays single-threaded, gracefully. A Safari report is not a regression.
- Receipts: `tools/isolation-headers-test.mjs` (9 assertions — headers, module scripts, the ws upgrade).

### DIFF
--- a/client/lib/conjure.js
+++ b/client/lib/conjure.js
@@ -4,7 +4,8 @@
 // Orrery (orrery.animalabs.ai) is the multistage 3D prompting service:
 // gpt-image turnarounds → Tripo mesh → texture/rig, stored as a version
 // tree. This panel drives its API directly, cross-origin with credentials
-// (Orrery ships CORS + a SameSite=None session for exactly this). The heavy
+// (Orrery ships CORS + a SameSite=None session for exactly this — including on
+// the asset endpoint, checked 2026-08-16). The heavy
 // bytes NEVER pass through this client: Orrery pushes finished GLBs to the
 // sequencer's /upload itself, and what we place is the returned store path.
 //
@@ -113,7 +114,20 @@ function progressOf(j) {
 
 function imageCandidates(j) {
   // the chain's first group = the image_gen siblings; thumbnails are their
-  // image assets, fetched straight from Orrery (plain <img>, cookie rides)
+  // image assets, fetched straight from Orrery.
+  //
+  // 🔴 NOT a plain <img> any more (2026-08-16). A bare <img> is a NO-CORS
+  // subresource, and under COEP `credentialless` — which this server now sends
+  // so wasm gets its threads — no-cors requests are sent WITHOUT credentials.
+  // These are private per-user assets, so they would have 404'd/403'd and the
+  // pick stage would have shown broken thumbnails with no explanation.
+  //
+  // crossorigin="use-credentials" moves them to CORS mode, which COEP does not
+  // touch. VERIFIED against the live service rather than assumed: the asset
+  // endpoint answers an eidoverse Origin with
+  //   access-control-allow-credentials: true
+  //   access-control-allow-origin: https://eidoverse.animalabs.ai
+  // (non-wildcard, as credentialed CORS requires).
   return (j.nodes ?? [])
     .filter((n) => n.op_type === 'image_gen' && n.status === 'completed')
     .map((n) => ({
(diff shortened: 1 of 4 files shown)


antra-tess (changes requested):
## Review: CHANGES REQUESTED at `fa11e0c419b8397f4d71b79cd74ee8cb712286d5`

The product direction looks right: COOP/COEP restores the precondition Piper already checks, websocket upgrade is deliberately untouched, and the second pass correctly found/fixed the two real cross-origin regressions (Orrery credentialed thumbnails and OAuth popup opener severance).

Independent receipts on macOS/Bun 1.3.14:

- `git diff --check 4ef5c79..fa11e0c`: clean
- `tools/isolation-headers-test.mjs`: **9/9**
- mutation removing the route wrapper: **4/9**, failing the five header assertions
- browser bundle: green (163 modules; standing Three/WebGPU warnings only)

I found two blockers in the receipt/lifecycle path and two small neighboring corrections:

### 1. The real-server test does not prove it is talking to its child

It chooses one of twenty fixed 89xx ports and treats the first answering listener as readiness. It never checks `srv.exitCode`, a child-owned nonce, or responder identity. We have already had an old 89xx listener buy a false green in this repository. A stale server carrying these headers could make this exact test pass after the spawned child lost `EADDRINUSE`.

Please bind readiness to the spawned instance. A cheap shape is a per-run environment nonce echoed in a test-only response header (or `/version`) and asserted on every probe, plus child-live checks. Put spawn/readiness/vectors inside one cleanup scope, await TERM, escalate to KILL, and remove the scratch world directory. Kernel-selected/verified-free ports are useful but are not identity by themselves.

### 2. The COOP-safe Orrery login timeout leaks its listener/state

After 45 polls the code only executes `clearInterval(poll); return;`. Under COOP the `postMessage` path cannot rescue it, so the `message` listener remains installed, `done` remains false, `connected` remains `false`, and the abandoned button flow has no visible completion/retry transition. Repeated attempts can accumulate listeners.

Please make timeout use an idempotent cleanup path: mark done, remove the listener, clear the timer, restore a retryable UI state, and report a bounded timeout. Also validate `ev.origin` against the configured Orrery origin before accepting `orrery:signed-in`; the retained non-isolated optimization currently trusts the data string alone.

### 3. The claimed catch-all fallback is the one response not isolated

If a future edit actually breaks the route-table catch-all, the final `new Response("not found")` bypasses `isolate()`. The comments claim every non-upgrade response gets the headers. Wrap that fallback too, or narrow the claim.

### 4. Pin the client-side compatibility fixes with an executable receipt

The current 9 assertions cover server headers/socket only. They do not protect the `crossorigin="use-credentials"` thumbnail repair or the polling cleanup that makes OAuth survive COOP. Please add a focused test for those two call-site contracts (including timeout cleanup / no listener accumulation). It need not be a full OAuth integration test.

Nonblocking documentation note: `COEP: credentialless` does not benefit every browser equally; the performance claim should remain conditional on `crossOriginIsolated === true`, while unsupported browsers retain the current single-thread path.

No request to change the chosen `credentialless` product direction. Return one exact head with owned responder identity, cleanup-scoped startup, Orrery timeout/origin repair, and focused client receipts; I expect the next pass to be small.

octopusburrow:
All four addressed at `649ceeb` — one head, as requested:

1. **Child identity**: the server now echoes a per-run `EIDO_BOOT_NONCE` on `/version` (absent unless the spawner sets it). Readiness = nonce match AND `exitCode === null`; identity is re-asserted after the vector run; spawn/vectors/teardown share one cleanup scope (TERM → await exit → KILL escalation → scratch dir removed). Now 11 assertions.
2. **Orrery login**: one idempotent `cleanup()` shared by every exit — the timeout runs it via `abandon()`, which latches `done`, removes the listener, clears the timer, closes the popup, restores `connected=false` (the button paints again) and logs a bounded-timeout reason. `ev.origin` is validated against the configured Orrery origin before the data string is trusted.
3. **Catch-all**: the wrap existed on our working branch and the PR snapshot predated it — the new head carries `isolate()` on the final `not found` (and only the isolation hunks; no unrelated drift).
4. **Client receipts**: `tools/conjure-coop-contract-test.mjs` (10 assertions) pins both call-site contracts — credentialed candidate ``s, and the single-teardown/timeout/origin shape — with corrupting negative controls so a drifted extraction anchor fails loudly rather than passing vacuously.

Doc note taken: agreed the threading benefit is conditional on `crossOriginIsolated === true` (Safari keeps the single-thread path; a Safari report is not a regression). That conditionality is stated in the PR description; the perf claim appears nowhere in code or test as unconditional.

Receipts on this head: isolation-headers 11/11 · conjure-coop-contract 10/10 · `bun build server/routes.ts` clean · zero SFU imports (standalone preserved).

antra-tess (approved):
Re-review on exact head `649ceebfc969265667874e83606f8ac2a90dcb89` — **APPROVE**.

The previous blockers are closed:

- the real-server receipt is bound to a per-run nonce, checks child liveness again after vectors, owns its scratch world, and tears the child down before cleanup;
- the catch-all 404 is isolated too, matching the claim that every non-upgrade response carries COOP/COEP;
- Orrery thumbnail candidates use credentialed CORS;
- OAuth has one idempotent cleanup path, a retryable 90s timeout, and origin validation on the retained postMessage optimization;
- focused source controls pin the thumbnail and OAuth repairs.

Independent exact-head receipts on macOS arm64 / Bun 1.3.14:

```text
isolation-headers-test             11/11
conjure-coop-contract-test         10/10
browser bundle                     green (standing Three/WebGPU warnings)
git diff --check                   clean
```

I also trial-merged this head onto current `origin/main` (`756ebab`) and reran both suites: clean merge, `11/11`, `10/10`.

Scope remains honest: this proves the isolation/header contract and the client compatibility repairs, not a measured Piper speedup on every browser. Unsupported browsers still retain the single-thread fallback.

PR #129, opened 2026-08-17
mcpl door: mention routing, channel capability gates, media whitelist hardening

octopusburrow:
Door-side half of the #104 stack. Stands alone on main (the `relay-cred`/`sfu-*` frames degrade to no-ops against a server that doesn't speak them).

- **Mention routing** (`mcpl/mention.ts`): one regex builder, one contract (returns null for an unmatchable id, callers decide), replacing three drifting inline copies.
- **Channel capability gates**: `channels.publish` / `channels.lifecycle` / `channels.streaming` were declared in CAP and checked nowhere — a host granting `channels.incoming` alone still had its agent's publish answered. All four verbs now gate through `granted()` semantics: plain-MCP hosts get no channel verbs, and a 0.5+ host is denied until its first `featureSets/update` (§5.3 deny-until-policy). Legacy 0.4 hosts keep everything.
- **Media whitelist** (amendment 1's least-authority): the sidecar lane can send exactly five media frame types; world-authoring and moderation verbs are refused and logged. Attest receipts are minted only by the door, never the sidecar.
- Resource-leak fixes on disconnect (media server handle, per-participant maps, ping ring).

Receipts on the branch: 82 (mention 36 · cap-gate 19 · whitelist 27 — the whitelist's negative controls corrupt the real extracted source and require detection, so they fail if the extraction anchors drift).


antra-tess (changes requested):
First review on exact head `5c71e4874eaf971b0c5fa8074e29e5307bbfba5c` — **CHANGES REQUESTED**.

The mention and capability work is directionally good. Independent exact-head receipts:

```text
mention regex                    36/36
channel capability predicates    19/19
media whitelist source checks    27/27
MCPL bundle                       green (standing Three/WebGPU warnings)
git diff --check                 clean
```

Escaping/null-safe mention construction, tokens.json shape validation, and the deny-until-policy channel gates close real seams. The media bridge, however, has two runtime blockers the source-only tests cannot see.

### 1. The performed-receipt path cannot succeed against the #103 server

`attestAired()` calls `WorldAgent.attestSay()`, which sends `{type:"attest", seq, digest, gen:this.surfaceGen}` on the **primary world connection**. But merged #103 server code accepts `attest` only when:

```ts
c.surface === "voice" && c.auxBound === true
```

and requires `msg.gen === c.gen` for that voice surface. The door's connection is `surface:"world"`; `surfaceGen` is captured from the primary snapshot, not the media-leg generation. Therefore every sidecar `aired` report is converted into an attest the server must refuse. The branch says the door mints the receipt, but the live server contract says only the bound voice leg can.

Please choose one coherent architecture and prove it through a real sequencer/door/sidecar path: either the authenticated bound voice leg attests directly, or add an explicitly validated primary-mediated receipt path bound to the current media leg/generation. Do not weaken #103 back to `spoken:true` or an unauthenticated seq claim. Required receipt: one real authored say + aired event produces one authenticated matching `performed`; wrong seq/digest/gen/session and stale predecessor produce none.

### 2. `EIDO_MEDIA_PORT` is process-global while the listener is per Session

Every Session reads the same env port and calls `Bun.serve()` on it. In the shipped shared MCPL door process, the first connected identity owns the port; the second session gets EADDRINUSE and silently remains text-only. Worse, the loopback websocket has no session credential/path: any local client can attach and replace the incumbent sidecar, and an `aired` frame is then treated as authority to request a performed receipt for whichever Session happened to win the port.

Either:

- build one process-level broker that multiplexes authenticated/session-bound sidecars, or
- make one-door-process-per-resident a declared and enforced deployment contract with a per-session unguessable handshake and no shared-door claim.

The current code sits between those topologies and is not safe/correct for the actual multi-session public door.

### 3. The tests need to cross the live doors

The capability and media tests currently inspect source slices. That pins spelling, not behavior. Please add an owned scratch sequencer/MCPL-door integration that proves at least:

- a 0.5 host granted incoming-only cannot publish/open/close/stream while allowed operations still work;
- two concurrent identities cannot collide or control each other's media bridge;
- session close releases its bridge and a reconnect can rebind;
- bind failure leaves text alive;
- the performed receipt succeeds only for the current authenticated media generation and fails for stale/foreign claims.

The mention/capability portion could be split and land independently if desired. No concern with its product direction; the blockers are in the media topology and executable receipt path.

octopusburrow:
Taking the split you offered — new exact head `e5d1049`: **this PR is now the mention/capability half alone; the media bridge exits the branch entirely.**

**Why the split rather than either receipt arm here.** Reading the merged #103 contract settled the architecture question your blocker 1 exposed: `attest` is the bound *voice surface's* privilege, and the door is `surface:"world"` — a door-minted receipt fights the gate it has to pass. Rather than extend #103 with a primary-mediated path, the coherent shape is the one the contract already names: the sidecar joins as the identity's **own bound voice leg** (the `media-peer.ts` topology, already demonstrated against the #103 server — agent bearer ⇒ `auxBound`, its own generation, its own attest), so receipt authority and the audio it vouches for live in one authenticated connection, with **no contract change and no receipt minted on anyone's behalf**. That recut belongs with the SFU stack, where the signalling frames the bridge relayed actually exist server-side — it will come with the real sequencer/door/sidecar receipt set you specified (one authored say + aired ⇒ one authenticated `performed`; wrong seq/digest/gen/session and stale predecessor ⇒ none). Your blocker 2 (the per-session `Bun.serve` on one process-global port, credential-less loopback attach) exits with the code it described; the topology decision gets made properly in that recut instead of being patched here.

**Item 3 — the tests now cross live doors.** `tools/door-cap-gate-live-test.mjs` drives an **owned scratch sequencer + owned MCPL door** over real websockets, with a raw world-ws observer embodied in the world as ground truth for what actually landed:
- deny-until-policy: a 0.5 host's `channels/publish` refused **before** its first `featureSets/update`;
- incoming-only grant: publish/open/close all `-32003`, `tools/list` empty, `tools/call` `-32601` (denied ⇒ as if never advertised);
- **nothing leaked**: the observer saw no say from the gated session — this is the assertion that catches a gate regression as a real utterance in a real world, not a status code;
- allowed operations still work: a full grant's publish delivers *and the observer sees the say in the world log*; `channels/open` answers with the channel;
- live mention routing end-to-end: an in-world `@claude` arrives as `channels/incoming` tagged `chat:mention` with `metadata.mentioned`.

`15/15`, stable across repeated runs on randomized ports. Mutant receipt: gutting `capAllowed` to `return true` fails 5 of the 15 — including the observer watching the "refused" say land in the world.

Receipts at this head: mention regex `36/36` · capability predicates `19/19` · live door integration `15/15` · MCPL bundle green (standing Three/WebGPU warnings) · `git diff --check` clean. An independent second-agent review of the split verified removal completeness repo-wide, that the kept half is byte-exact the reviewed mention/caps delta, and attacked the new test for vacuousness (timeouts cannot masquerade as `-32003`; a dead observer fails the run rather than blessing it).

antra-tess (changes requested):
Exact-head re-review at `e5d1049bfb54ca73616396d452f31fda77913963`: **CHANGES_REQUESTED** — the product split is right; the new live receipt still has the exact ownership/state defects it claims to remove.

## What is now good

The incoherent media bridge has been removed completely from this slice. That closes the prior #103 receipt-authority mismatch and the shared unauthenticated `EIDO_MEDIA_PORT` topology by deleting them, not papering over them. Mention escaping/null safety and channel capability gating remain independently valuable.

- mention regex: **36/36**
- capability predicates: **19/19**
- real door behavior: **15/15**
- MCPL bundle: green (standing Three/WebGPU warnings only)
- `git diff --check`: clean

The live test meaningfully proves deny-until-policy, incoming-only refusal for publish/open/close/tools, no gated say reaching a raw world observer, full-grant publish actually landing, and live mention routing with tag + metadata.

## Blocker: the “owned scratch” harness is neither owned nor scratch-clean

The real-door test uses random fixed ports and accepts readiness from unauthenticated `/avatars` and `/healthz`; it sets no per-run world/door nonce and does not verify child identity/liveness. A stale listener can still buy a green. PR #125 now contains the repository's established portable pattern: kernel-assigned ports plus child-bound nonce echo (OS process ownership is only a Linux second opinion). Reuse that pattern rather than adding a third near-miss.

More concretely, my exact-head run ended with:

```text
15 passed, 0 failed
?? mcpl/state.json.tmp
```

The test does not pass `MCPL_STATE`, so the door writes into the source checkout. Its shutdown calls SIGTERM then immediately SIGKILL without awaiting child exit, then retries `rmSync` twice with no delay; the door can still be mid-persist after scratch removal. This is the same “green test contaminated the repository / teardown raced state persistence” defect class just closed on #125.

Required receipt repair:

1. bind world and door to per-run nonces and verify the responder is the spawned child;
2. use kernel-assigned ports or an equivalent bind-owned handoff;
3. bind tokens, worlds, and MCPL state to the scratch directory;
4. put spawn/readiness under one `try/finally`, SIGTERM and **await exit**, escalate to SIGKILL only on timeout;
5. assert source-tree state/tmp byte+mtime unchanged and scratch artifacts cleaned;
6. provide stale/impostor-listener and state-path negative controls;
7. rerun the existing 15 behavioral checks unchanged.

The product code itself needs no redesign in this review. Once the executable receipt actually owns its processes and state, I expect this split slice to be approvable.

octopusburrow:
New exact head `93ef2c8` — the receipt repair, all seven items, built by reusing the #125 pattern rather than inventing a third near-miss.

**Infrastructure carried from the approved `rfc005-channel-join` text, byte-identical** (diff-verified region by region, so the eventual merge deduplicates): `MCPL_TOKENS`/`MCPL_STATE` overrides with unchanged defaults, the state-tmp boot sweep, hardened `persistState` (named tmp, cleaned on failed write/rename), the SIGTERM/SIGINT handler that makes teardown event-loop ordered, `MCPL_INSTANCE_NONCE` echoed by `/healthz`, and `WORLD_INSTANCE_NONCE` echoed by `/version`. Production behavior with no env set is byte-identical to before.

**The harness now** (`tools/door-cap-gate-live-test.mjs`): kernel-assigned ports (bind `:0`); per-run nonce identity verified for **both** children before any assertion, with the stale-listener probes that **throw** rather than spin — a nonce-less `ok` and a foreign-nonce `ok tXXX…` both refuse loudly; OS port-ownership (`ss` + `/proc` descendants) as the Linux second opinion, loudly reported as skipped where the platform can't answer; tokens, worlds, and MCPL state all scratch-bound; spawn/readiness under one `try/finally` with SIGTERM → **await exit** → SIGKILL only on timeout (the door's signal handler gets the grace to sweep its own tmp); and the receipts you specified — seeded stale tmp swept at boot, no tmp left in the owned state dir, repository state unchanged by **content hash + mtime**, scratch removed. Two negative controls: an HTTP impostor answering the pre-nonce `ok` is refused as stale rather than adopted, and the door's state is positively shown to land in scratch. The 15 behavioral checks ride unchanged (reindented only, by the `try` wrap); the suite is 22, stable across repeated runs.

**Mutant receipts:** `capAllowed → true` fails 5 (including the observer watching the leak land in-world); door hardcoded back to the repo state path fails 4 (boot sweep, scratch control, tmp receipt, repo hash receipt); nonce echo stripped from `/healthz` refuses readiness with the explicit stale-listener error. My run of your exact-head scenario ends with a clean `git status` — no `state.json`, no `.tmp`.

Receipts at this head: mention regex `36/36` · capability predicates `19/19` · live door `22/22` ×3 · MCPL bundle green (standing warnings) · `git diff --check` clean. An independent second-agent pass verified the byte-parity claim region-by-region (and caught one comment-wording drift, adopted from #125 verbatim in this head), ran the breakage case (child pointed at a dead entrypoint: legible failure, children reaped, no litter), and confirmed all three mutants are caught.

antra-tess (changes requested):
Exact-head re-review at `93ef2c80cf034f546ba04984f01786a75d2173b5`: **CHANGES REQUESTED**.

The product split remains sound and the focused receipts passed independently:

- mention regex: **36/36**
- channel capability gate: **19/19**
- `git diff --check`: clean

But the repaired live door harness is still not portable on the established macOS/Bun review host. It dies before behavioral vectors at `listenerPids()`:

```text
error: Executable not found in $PATH: "ss"
  at listenerPids (.../tools/door-cap-gate-live-test.mjs:105)
```

The source wraps `spawn("ss", ...)` in `try/catch`, but Node/Bun child-process ENOENT is delivered as an asynchronous `error` event, not a synchronous throw. With no `p.on("error", ...)`, the process-level `uncaughtException` handler exits 1. Thus the documented contract — “no `ss` or `/proc` ⇒ null, nonce remains the portable proof” — is not the effective behavior.

Required narrow repair:

1. make `listenerPids()` resolve `null` on missing/failed `ss` without an unhandled `error` event (or skip the Linux second opinion entirely when `process.platform !== "linux"`);
2. retain nonce identity as the portable authority and keep foreign-nonce/stale-listener negatives;
3. rerun the full owned live-door suite on macOS, plus focused 36/36 and 19/19, bundle, and clean diff;
4. add a deterministic missing-`ss` vector so the portability claim cannot regress.

No feature redesign requested. The harness must earn its own behavioral receipt before this slice can be approved.

octopusburrow:
New exact head `ab73a61` — the portability repair, all four items, narrow as asked.

**The defect, confirmed against source:** `listenerPids()` spawned `ss` inside try/catch, but Node/Bun deliver spawn ENOENT as an **async `error` event** — the sync catch never sees it, the process-level `uncaughtException` hook does, and the harness died at exit 1 before any behavioral vector. The documented contract ("no `ss`/`/proc` ⇒ null, nonce remains the portable proof") was not the effective behavior. Your diagnosis was exact.

**The repair (both belts):**
1. **Platform guard** — `process.platform !== "linux"` returns `null` before any spawn: the OS second opinion is Linux-only by design, and a macOS host shouldn't pay a spawn to learn that.
2. **`error` handler** — on Linux, a missing/failed `ss` (minimal containers) resolves `-1 ⇒ null` via `p.once("error", …)` racing `close`; no unhandled event exists on any path. Nonce identity stays the sole portable authority; the foreign-nonce and stale-listener negatives are untouched.
3. **Deterministic missing-`ss` vector** (negative control 1b) — forces the ENOENT path with a command that cannot exist and requires the contract: resolve `null`, process alive to say so. On non-Linux the guard answers `null` before the spawn — same contract, earlier.

**Receipts (Burrow is Linux, so the darwin leg is proven by construction + simulation):**
- Full owned live-door suite, normal PATH: **23/23** (`identity: per-run nonce + OS port ownership (linux)`)
- Full suite with `ss` stripped from PATH — the exact scenario that killed your run: **23/23** (`identity: per-run nonce only … OS check skipped`)
- mention regex: **36/36** · channel capability gate: **19/19** · `git diff --check`: clean
- Diff is 1 file, +27/−4 — no feature change, no redesign.

Honest coverage note: on your macOS host the new vector passes via the platform guard (null before spawn), so it exercises the guard there, not the `error` handler; the handler path is exercised on Linux (the stripped-PATH run above and any Linux-without-`ss` host). Between the two hosts the regression surface is covered from both sides.

antra-tess (approved):
Approved exact head `ab73a616918474297599eb1f544242a8b8247372`.

Independent receipts on macOS/Bun 1.3.14 after wiring the repository's actual nested `mcpl/` dependency tree:
- mention regex `36/36`
- channel capability gates `19/19`
- owned real-door matrix `23/23`
- deterministic missing-`ss` vector resolves `null` without process crash; nonce remains portable responder identity
- stale/nonced responder, deny-until-policy, incoming-only refusal, observer no-leak, granted publish/open, live mention routing, scratch-state ownership, orderly teardown, repo state byte+mtime, no tmp residue all green
- `git diff --check` clean
- clean trial merge onto current main `95303a7`

The earlier macOS blocker is closed: non-Linux no longer spawns `ss`, while Linux ENOENT is handled through the child `error` event rather than becoming an uncaught exception. No media/cutover authority is implied by this approval.

PR #130, opened 2026-08-17
sfu engine: the in-process relay floor (#104 phase 1) — engine only, nothing wires it yet

octopusburrow:
The core of #104: an in-process werift SFU — no external media server, no SDK, the world server IS the relay. This PR is the engine as a closed cluster (`sfu.ts`, `sfuguard.ts`, `sfuadapter.ts`, `relaydecision.ts`, `transport.ts`, `sfusupervisor.ts`); nothing on main calls it yet, so it can merge without changing any behavior. Design doc rides along: `notes/SFU-SPEC.md`, which re-reads the amendments and maps each to its receipt.

Amendment coverage:
- **A1 (seven refusals)**: `relaydecision.ts` decides admission from claims vs live state — wrong world · no identity · retired primary gen · retired media gen · prior incarnation · nonce replay (burn on first admission) · no-bearer-token-to-expire. 23 tests drive the decision layer, 35 more drive it through the adapter.
- **A2 (generation binding)**: answers must name their generation; an answer that can't belongs to no live leg.
- **A3 (three independent states)**: listener receive-consent (standing, gen-watermarked, fail-closed on reconnect), publisher self-mute (pre-encode gate, client side), moderator mute (enforced at ingress, one store).
- **A4**: the Bun kill-questions mostly evaporate — no SDK event loop to starve; the supervisor covers the rest.

The consent/lifecycle surface went through four rounds of adversarial review (a refused revoke now flips the standing answer too; a re-mint funnels through revoke so nothing of the predecessor survives; the answer-timeout deletes by identity, not key). `package.json`/`bun.lock` carry the werift dep so the branch actually stands alone.

Receipts: 115 (sfu 57 · adapter 35 · decision 23) + a werift↔real-Chromium isolation harness (`tools/mini-sfu-test.mjs`) that reproduces the addTrack transceiver mis-pick class in seconds without a world.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -15,9 +15,11 @@
         "pngjs": "^7.0.0",
         "quickjs-emscripten": "^0.32.0",
         "sharp": "^0.33.5",
+        "werift": "^0.24.4",
       },
       "devDependencies": {
         "@happy-dom/global-registrator": "^20.11.1",
+        "playwright": "^1.62.1",
       },
     },
   },
@@ -25,6 +27,10 @@
     "@dimforge/rapier3d-compat": ["@dimforge/rapier3d-compat@0.19.3", "", {}, "sha512-mMVdSj1PRTT108s9Swbu2GQOmHbn8kbJANRV5xfczL3s0T4vkgZAuoMRgvBzQcHanpKusbC0ZJj6z3mC3aj3vg=="],
 
     "@emnapi/runtime": ["@emnapi/runtime@1.11.3", "", { "dependencies": { "tslib": "^2.4.0" } }, "sha512-Xz4Tpyki7XyrpbUK1jR1AhdAdaXyhhY4lZ3neLodmhpuWfy2PAQN5B46sAiU4liOXGLkHypn/qU+jvfWSCYYLA=="],
+
+    "@fidm/asn1": ["@fidm/asn1@1.0.4", "", {}, "sha512-esd1jyNvRb2HVaQGq2Gg8Z0kbQPXzV9Tq5Z14KNIov6KfFD6PTaRIO8UpcsYiTNzOqJpmyzWgVTrUwFV3UF4TQ=="],
+
+    "@fidm/x509": ["@fidm/x509@1.2.1", "", { "dependencies": { "@fidm/asn1": "^1.0.4", "tweetnacl": "^1.0.1" } }, "sha512-nwc2iesjyc9hkuzcrMCBXQRn653XuAUKorfWM8PZyJawiy1QzLj4vahwzaI25+pfpwOLvMzbJ0uKpWLDNmo16w=="],
 
     "@gltf-transform/core": ["@gltf-transform/core@4.4.2", "", { "dependencies": { "property-graph": "^4.1.0" } }, "sha512-qsWKwNSwK+2s834Mt4xbYcyHqCrgNFP7hIv5s487JxebngRfDgelpghNF+kSswGb2/NuapasfK3UViFoSJJoMg=="],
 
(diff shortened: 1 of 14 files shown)


antra-tess (changes requested):
I reviewed exact head `9f455ecbbd747404ee3c7ca069144f0e767e748d` independently on macOS arm64 / Bun 1.3.14.

The core shape is promising, and the branch has unusually good adversarial notes. Independent receipts at this head:

- `tools/sfu-test.ts`: **57/57**
- `tools/sfu-adapter-test.ts`: **35 passed**
- `tools/relay-decision-test.ts`: **23/23**
- N=2/6/12 synthetic load: 100% delivery in all three runs
- clean worktree

I am requesting changes for four runtime blockers and two receipt blockers.

## 1. The process-global transport guard converts arbitrary uncaught failures into a voice-only degraded state

`installSfuTransportGuard()` installs process-wide `uncaughtException` / `unhandledRejection` handlers. Every non-benign error is logged and passed to the voice supervisor, but the process is deliberately kept alive. This is not scoped to werift: an unrelated world-state or persistence bug now loses Bun's default-fatal semantics and leaves the whole sequencer running after an unknown uncaught exception.

The benign classifier is also process-global: any `Error` carrying one of the listed errnos plus a `send*`/`recv*` syscall is swallowed, regardless of whether it came from werift's UDP socket.

Please contain the workaround at the offending werift/dgram socket (or upstream it). If a global guard remains temporarily, it may swallow only positively identified benign werift transport errors; every other uncaught failure must preserve the ordinary fatal contract rather than being reclassified as “voice degraded.” Add a child-process negative proving an unrelated uncaught exception still exits nonzero at its original site.

## 2. `sfusupervisor` reports LIVE after recurring faults — and after silence rather than positive recovery

`markVoiceDegraded()` returns immediately when already degraded, before updating the reason or rearming its 30s timer. I reproduced:

```text
fault-1 at t=0
fault-2 at t=20
at t=31: state=live, reason="no further faults for 30s"
```

The second fault did not extend the window. More fundamentally, absence of another uncaught error is not evidence that a broken SFU recovered; a dead subsystem can be silent forever. Please either keep the service degraded until an explicit/operator restart, or define a positive health receipt (successful owned negotiation / controlled packet canary) that is allowed to return it to live. Repeated faults must never age into green. Add deterministic clocked tests.

## 3. Speaker reconnects leak one outbound transceiver per listener per generation

`closeLeg(id)` deletes `listener.outbound[id]` but retains no transceiver handle and never stops/removes the corresponding sender/transceiver. A fresh speaker with the same id allocates another one. Minimal exact-head probe:

```json
{"outbound":0,"transceivers":20,"routes":[]}
```

after twenty create/consent/close cycles for one speaker against one persistent listener. The map looks clean while SDP/PC state grows without bound; repeated resident reconnects are ordinary, not an edge case.

Please retain route ownership strongly enough to stop/retire its transceiver on speaker retirement (or deliberately reuse a compatible route), and add a repeated same-id reconnect test that bounds transceiver/m-line count and proves no stale track/audio survives.

## 4. ICE is not generation-bound

Answers require `claimedGen`; `sfuAcceptIce(world,id,candidate)` does not. A stale predecessor can trickle candidates into the successor's peer connection after same-id takeover. Carry and require media generation on ICE exactly as on answers, with stale/missing-generation vectors.

## 5. The mini-SFU receipt is hardcoded to one Linux browser path

`tools/mini-sfu-test.mjs` hardcodes:

```text
/home/claude/.cache/ms-playwright/chromium-1228/chrome-linux64/chrome
```

So the branch's own suite terminates code 1 on the established macOS review host. Replacing only that path with the installed macOS Chromium made the test pass all three stages. Please use Playwright's managed/default executable or an explicit optional env override; do not commit one machine's absolute path. This should travel as an executable receipt.

## 6. Stale-position fail-open is contradicted when the audible cap is enabled

`inEarshot()` correctly forwards on stale positions, but `withinCap()` immediately ranks the same stale coordinates through `d2()` and can suppress. Exact-head probe with 10s-old positions and `maxAudiblePerListener=1` yielded:

```json
{"forwarded":1,"suppressed":{"gated":0,"capped":1}}
```

although the source contract says stale position forwards. Either stale/unknown position disables cap ranking for that listener/speaker, or document and test the different policy explicitly; the current two stages disagree silently.

## Receipt hygiene

`git diff --check` also finds trailing whitespace at `tools/sfu-adapter-test.ts:203`.

The synthetic load runner includes its post-send 1s drain in `elapsed`, so sustained CPU/egress are understated. In my N=12/3s run it reported 3.48 Mbps even though the active-send arithmetic is ~4.86 Mbps (`600 pkt/s × 11 × 92 bytes × 8`). Keep a drain for delivery accounting, but report active-send rate separately from end-to-end drain time.

After the changes, please return one exact head with:

- clean `diff --check`;
- portable mini-SFU pass;
- focused guard/supervisor/reconnect/ICE/stale-cap regressions and fail-on-old-head controls;
- the existing 57 + 35 + 23 suites;
- N=2/6/12 load with honest active-window and drain-window metrics.

This is engine-only review. It does not authorize #132 cutover or production deployment.

octopusburrow:
All six findings verified against source and fixed at one head: **d6c072b**. Each was real; thank you for the exact-head receipts — the clocked supervisor repro and the 20-cycle transceiver count made every verification a five-minute read instead of an argument.

**1 — fatal contract restored.** `reportFatal` now logs the original stack and exits nonzero; the guard narrows Bun's default-fatal semantics by exactly the positively-identified benign set (structured errno + `send*`/`recv*` syscall) and nothing else. The "voice degraded but alive" outcome for unknown errors is gone — you were right that it was the codebase's own catch{}-taboo promoted to process scope, and amendment 2's recovery story requires the restart to actually happen. Child-process negatives prove both directions: an unrelated `TypeError` exits 1 with its own message on stderr; a benign werift errno is swallowed and counted.

**2 — no silence timer.** Your repro was exact: the `state === "degraded"` early-return sat before the rearm, so the comment described code that didn't exist. Rather than fix the rearm, the timer is deleted — absence of faults was never evidence of recovery. Degraded is now sticky until a positive receipt: a completed owned negotiation (offer → browser answer → `setRemoteDescription` OK) calls `markVoiceLive`. Repeat faults update the reason (latest fault is what an operator needs). Clocked tests pin: repeat-fault stays degraded, nothing ages green, the receipt returns live.

**3 — routes survive speaker retirement.** The route entry held the only reference to a transceiver werift keeps forever; deleting it leaked one m-line per reconnect generation, exactly as you measured. Since routes are keyed *and* msid'd by speaker id, a same-id successor now reuses the retained route — reconnect becomes a zero-SDP event and transceivers are bounded by **distinct speakers**, not reconnect count. No stale audio: consent for the retired id is still wiped (fanout checks consent before the route, the same starvation mechanism consent-revocation has always used), and the corpse's own fanout stops at `from.closed`. Tests: 20 same-id cycles → 1 transceiver, 1 route; successor starved until fresh consent; fresh consent flows through the reused route. One pre-existing assertion ("nobody holds a route to it") pinned route-absence rather than the safety property and was updated to pin consent-starvation, with a comment explaining why.

**4 — ICE is generation-bound.** `sfuAcceptIce` now mirrors `sfuAcceptAnswer` exactly: generation required (not opt-in — the stale predecessor is precisely the client that would omit it), mismatch dropped. Vectors: missing gen, retired gen, live gen, post-takeover stale gen. Note: this head is engine-only, so the `sfu-ice` websocket case and the client's `gen` field live in the wiring branch (pr3); that branch will thread `gen: cred.mediaGen` through the same way the answer path already does.

**5 — portable receipt.** Playwright's managed browser is the default; `SFU_TEST_CHROME` is an explicit override. No machine paths committed. (Your finding promptly re-proved itself in reverse: this Linux host's Playwright wanted a headless-shell revision that isn't installed, so the local run uses the override — which is the design working.)

**6 — one staleness rule.** `d2()` now treats stale as unknown, so `withinCap` cannot rank coordinates `inEarshot` refuses to gate on — the policy is unified rather than documented-as-divergent: a position too old to gate on is too old to rank on, and an unrankable challenger is admitted (fail-open, matching the gate's contract). Your probe's case (10s-old positions, cap=1) now forwards with `capped:0`; the fresh-incumbent/stale-challenger mix is pinned too.

**Receipt hygiene.** `git diff --check` clean. `sfu-load` reports the active-send window and the drain separately — rates (packets/s, egress, CPU%) over the active window, delivery counted over both. Local N=12/3s now reports egress consistent with its own active-rate arithmetic (539/s measured × 11 × 92B × 8 = 4.36 Mbps reported).

**Receipts at d6c072b (Linux x64 / Bun 1.3.14, WSL):**
- `tools/sfu-test.ts`: **66/66** (was 57; +9 for items 1/3/6 incl. the two child-process negatives)
- `tools/sfu-adapter-test.ts`: **42 passed** (was 35; +7 for items 2/4)
- `tools/relay-decision-test.ts`: **23/23**
- mini-SFU 3-stage: passes (with `SFU_TEST_CHROME` on this host, per item 5)
- N=2/6/12 load: 100.0% delivery all three; honest-window CPU on this laptop reads 55%/85%/154% of one core — the N=6/12 verdict lines say "at or past the limit" here, which is this WSL machine's ceiling, not a delivery failure; your arm64 numbers will differ.

Understood on scope: this changes nothing about #132 cutover or deployment authorization.

antra-tess (changes requested):
Re-review on exact head `d6c072b260e88c76d5a105779a81165f86748533` — **CHANGES REQUESTED** (two remaining structural blockers; the other review findings are materially closed).

Independent exact-head receipts on macOS arm64 / Bun 1.3.14:

```text
sfu-test                         66/66
sfu-adapter-test                 42/42
relay-decision-test              23/23
mini-sfu browser smoke           pass with SFU_TEST_CHROME override
N=12 all-speaking load           100% delivery; 52% one core; +33 MB RSS; 4.59 Mbps
git diff --check                 clean
```

The branch now has positive recovery receipts, generation-bound ICE, bounded same-id reconnect transceivers, stale-position/cap consistency, a portable browser override, truthful active-window load rates, and strong negative controls. Those fixes are good.

### 1. The process-global guard is still not positively scoped to owned werift transport

The exact request was that the guard swallow only errors from **owned werift transport**, not any process-global error that happens to look UDP-shaped. `isBenignTransportError()` still classifies solely by `Error + errno code + send/recv syscall`. Its own new positive test proves the gap: it installs the guard and swallows a **standalone Node dgram** `ECONNREFUSED`, with no werift ownership involved.

That means any other present/future UDP subsystem in the sequencer can lose a real uncaught send/recv fault merely because it shares the same structured syscall fields. This is narrower than the old message regex, but it is still process-global catch-by-shape rather than ownership.

Please bind containment to the actual werift socket/transport (e.g. install an error handler at the owned transport seam / patch or wrap the werift transport) so unrelated dgram failures preserve ordinary fatal semantics. Keep the child-process negative proving an unrelated failure dies, but add the discriminating sibling: an **unowned dgram** error with the same errno/syscall must not be swallowed.

### 2. The standalone engine branch says the mesh is already deleted, contradicting both this PR and its own spec

`server/transport.ts` says:

```text
the mesh is now DELETED ... off = NO VOICE AT ALL ... rollback is git revert
```

But this PR is explicitly “engine only, nothing wires it yet,” does not delete/wire the current mesh, and `notes/SFU-SPEC.md` repeatedly says A5 is absolute: keep the mesh as production/rollback until acceptance; deletion belongs to a separate operator-approved cutover PR. The same file also claims an env-swappable LiveKit escape hatch while `voiceTransport()` currently exposes only `sfu|off`.

Please make the standalone branch truthful about its current state: engine inactive/unwired, existing mesh untouched, and no claimed external adapter that does not exist. #132 can change transport semantics when/if cutover is separately approved.

No redesign of the SFU engine requested. After those two seams are closed, retain the current tests, clean diff, browser smoke, and load receipt; then I expect the engine slice to be ready for approval. This remains code review only, not #132 cutover authority.

octopusburrow:
New exact head: `de81c35`. Both structural blockers closed; `70 passed, 0 failed` (sfu-test, was 66), adapter `42/42`, relay-decision `23/23`, `git diff --check` clean.

**1 — containment is now bound to owned werift transport, at the seam you suggested.** Every raw dgram socket werift creates flows through its exported `UdpTransport.init` — verified against the ESM bundle Bun actually loads (one `UdpTransport` class, three internal creators, all through the static; the sctp module's same-named class wraps an existing socket and creates none; note the CJS tree under `lib/` is dead code for `import "werift"`, so a patch there would contain nothing). The guard wraps that one choke point: werift-created sockets get the benign-errno error handler (swallowed + counted, surfaced in diag) and a send wrapper containing the callback-branch rejection (the non-IP branch is werift's only rejecting path). The process-global handlers are now **pure fatal reporters — no benign branch at process scope at all**.

Your discriminating sibling, both flavors: the exact synthesized errno/syscall shape v2 swallowed — no werift ownership — now **dies** (that vector *was* v2's own positive test; the round-2 point being precisely that it passing was the bug), and a **real non-werift dgram socket** erroring with a benign shape dies too. Owned-side: benign on a werift socket survives and counts; **non**-benign on an owned socket is still fatal with its site on stderr; a benign callback-send rejection is contained. Mutant receipt: v2 guard restored under the new tests → both unowned discriminators fail.

The whole suite now runs **with the guard installed in the main process**, so the 66 negotiation vectors double as the receipt that the patch is transparent under real `RTCPeerConnection` gathering. (A second-agent review of this fix also confirmed the seam empirically — a wrapped init intercepts 6 socket creations during one real offer/gather — and surfaced the seam's honest edges, now in the header: bind-time errors stay default-fatal as upstream; the bundle's unused, receive-only `rtpSource` socket sits outside the seam; `werift/nonstandard` is a separate bundle the patch can't reach and nothing imports. One consequence is stated as contract rather than hidden: **install order is load-bearing** — the patch covers only post-install transports, so the guard must install before the first peer construction, and #132 should assert that when it wires the guard into server.ts.)

**2 — the standalone branch now tells the truth about itself.** `server/transport.ts`: the mesh is *not* deleted here — `client/lib/voice.js` remains the untouched production voice path per SFU-SPEC A5, `"off"` means "no SFU selected," not silence, and both mesh retirement and the change of meaning belong to the separately-authorized #132. `notes/SFU-SPEC.md`: the LiveKit "drop-in escape hatch" is re-described as what exists — a swap *seam* behind `voiceTransport()` (`sfu|off`), with the deleted adapter in git history as a starting point — because claiming a built, config-flippable hatch would be exactly the untested rollback story A5 forbids.

No engine redesign; the diff is the guard, the tests, and the two truthfulness passes.

antra-tess (approved):
Exact-head re-review at `de81c359829539eeec14e166a54e26023768c51e`: **APPROVE**.

Both round-2 blockers are closed.

## 1. Containment is now ownership-bound

The benign errno branch moved off process scope and onto the werift-owned `UdpTransport.init` seam. Process-global `uncaughtException` / `unhandledRejection` handlers are now pure fatal reporters.

Independent exact-head receipts (`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0`):

- SFU policy/guard suite: **70/70**
- adapter: **42/42**
- relay decisions: **23/23**
- mini-SFU browser negotiation: pass
- N=12 / 12-speaking load: 132/132 routes, 66,000/66,000 packets heard (100%), 49.4% one core, +28MB RSS, 4.58Mbps over the 10.60s active send window
- server/SFU bundle: green
- `git diff --check`: clean
- clean trial merge onto current `origin/main` (`756ebab`): no conflict

The tests now discriminate the finding directly:

- unowned synthesized errno/syscall-shaped failure dies;
- a real non-werift dgram failure with the same shape dies;
- benign error on a werift-owned socket is swallowed and counted;
- non-benign owned-socket error remains fatal with its original site;
- benign owned callback-send rejection is contained.

Install order is truthfully named as load-bearing for the later integration PR: the guard must install before the first RTCPeerConnection.

## 2. The engine branch now describes itself truthfully

`transport.ts` no longer says the mesh is deleted or that `off` means silence on this branch. It states that #130 is engine-only and unwired, with current P2P production untouched until the separately authorized cutover. `SFU-SPEC` now says the external-relay seam exists but no LiveKit adapter does; it no longer advertises a config-only escape hatch that has not been built.

Operator clarification made after my earlier #132 review: the target architecture after accepted cutover is relay-only — no maintained tandem mesh system. That does not alter #130's truthful current-state statement: at this engine-only head, the existing mesh remains untouched. Mesh retirement belongs only to the future operator-approved cutover delta.

Approval is for the standalone engine slice only. It is not approval to merge the unresolved #129/#131 dependencies, wire #132, deploy, or cut over voice.

PR #131, opened 2026-08-17
client: mic foundation + audio panel + TTS voices (BYOV)

octopusburrow:
Browser-side half of the #104 stack, one PR so every new module lands with a caller. Stands alone on main: transport reads are guarded runtime globals that degrade to null.

- **micstate.js / micgate.js**: the mic's audio machinery extracted from the old transport — noise-floor gate (learned floor, hysteresis, 50 ms lookahead so unvoiced consonants survive), mute authority, the ONE `micOn()` answer, and a refuse-rather-than-publish-ungated rule with an explicit opt-out row.
- **Audio panel**: house `.row`/`.row.wide` classes (no private CSS to drift), meter on the RAW device (a sensitivity meter that only moves past the threshold can't help you set the threshold), self-monitor row (tapped after the gate, deliberately delayed 400 ms), display state derived from real state throughout.
- **TTS (BYOV)**: bring your own piper model or speech server; visible list with the verb on the row (no dropdown), per-speaker ordered utterance queues, epoch-cancelled synthesis at every async seam, `MediaStreamTrackGenerator` pacer with liveness-checked rebuild, and reader-rules in `tts-chunk.js` (fenced code never speaks, URLs say "link", symbol glyphs become the pause they meant).
- `boot-check.mjs`: loads the real module graph in a real browser and fails on any pageerror — the instrument that caught two link-time boot-breakers every node-side test missed.

Receipts: 10 UI/behavior suites + micstate exec 10 + spoken-form 11 + boot-check, all run on the clean branch checkout.


antra-tess (changes requested):
First review on exact head `60562eea85fddf15de08c72827ae9a92a6c12102` — **CHANGES REQUESTED**.

There is substantial useful work here: the extracted mic state, fail-closed gate, raw-side meter, in-place panel updates, sidetone separation, ordered TTS queues, reader rules, and async epoch checks are all good directions. Independent exact-head receipts that ran in this checkout included micstate `10/10`, spoken-form `11/11`, the non-browser UI/selection/gain probes, a green client bundle, clean `diff --check`, and a clean no-conflict merge onto current main.

Three runtime/receipt blockers remain.

### 1. The PR does not actually stand alone on current main: it disables the existing mesh mic

The body says transport globals "degrade to null" and the PR stands alone. But `mictoggle.js` switches from current-main's `voice.js.toggleMic` to `micstate.toggleMic`, whose only implementation is:

```js
const fn = window.__sfuMic;
if (!fn) { flashHint('voice is still connecting'); return micOn(); }
```

`__sfuMic` is installed by the later SFU/cutover stack, not by this PR or current main. So after merging #131 alone, V/click no longer calls the live mesh mic at all; it only says "connecting" and leaves the mic off. Guarded reads are not enough when the mutation path has no fallback.

Please either preserve the current mesh mutation path until cutover, or explicitly stack/rebase this PR on the transport that installs the hook (without claiming standalone safety). Required discriminator: current-main + #131, with no SFU globals, can turn the existing mesh mic ON→OFF→ON and the visible state follows it. It must fail on this head.

### 2. `releaseMicrophone()` can irreversibly stop the standing lane it promises to preserve

`micstate.js` says release kills the raw device while keeping `_lane` alive, but executes:

```js
for (const t of (_raw || _lane).getTracks()) t.stop();
```

After one release `_raw=null` while `_lane` remains. A second/idempotent release, or a synth-only lane with no raw device, stops the destination track itself. `_lane` is then retained as an ended track — the exact renegotiation/reacquisition failure the function says it prevents.

The existing `micstate-exec-test` cannot see this: it calls `gateRelease()` first, which sets `_lane=null`, so its later `releaseMicrophone()` is a no-op. Please stop only `_raw` tracks, make repeated release harmless, and add a behavioral vector proving the same lane track remains live/identical through repeated release and reacquisition, including a synth-only lane.

### 3. The browser receipts are Linux-path/ambient-server specific

Seven Playwright probes hardcode:

`/home/claude/.cache/ms-playwright/chromium-1228/chrome-linux64/chrome`

On the established macOS/Bun review host they all fail before behavior. After a review-local path substitution, six probes still point at ambient fixed ports (for example `127.0.0.1:8946`) and immediately fail `ERR_CONNECTION_REFUSED`; they neither own a server nor authenticate its identity. `boot-check` tries to own a child, but also hardcodes the Linux Chrome path and a Linux Bun fallback.

Please use managed Playwright or an explicit documented browser override, and make each server-dependent probe spawn/identify/tear down its own child (nonce + scratch state), or consolidate them behind one owned harness. Add meaningful fail-on-main/mutation controls. A stale staging server must not be able to buy a green, and a clean checkout must be runnable on the supported review platforms.

The current clean merge is not enough because these defects live in effective runtime composition. No objection to the feature direction; please return one exact head with the standalone transport seam, idempotent lane preservation, and portable owned browser receipts closed.

octopusburrow:
New exact head: `089a844`. All three blockers closed, each with its discriminator run and a mutant receipt. Roster at this head: micstate exec `10/10`, release vectors `14/14`, mesh-fallback vectors `13/13`, spoken-form `11/11`, and four browser receipts against **owned children** — boot-check, mic-hud-probe `6/6`, mic-meter-states (three states + latch), panel-teardown (six checkboxes, all in-place), tts-blocking (chunked) — plus client bundle `170 modules` and clean `git diff --check`.

**1 — the PR now actually stands alone: the mesh is the transport when no SFU hook is installed.** `micstate.toggleMic` selects the transport rather than forking per call: the installed `__sfuMic` hook if present, else the live mesh (module loaded once — it is already in current main's graph; the catch covers the post-cutover world), else the honest hint. `micOn()` defers to the mesh's own three-state answer, keeping it THE one answer instead of a stale second one — and a second-agent review of my first cut caught the same disease one symptom over: `micAnalyserLevel`/`micGateInfo` still read micstate's mesh-blind state, so the mic would have toggled while the hot glow and sensitivity meter sat at zero. Both now defer through the same seam. Your exact discriminator runs two ways: as a unit vector (mocked mesh: ON→OFF→ON, name passed through, visible state each step, SFU hook outranking; fails 6 vectors on the pre-fix head) and **through the real HUD button in a real browser against an owned child** (`tools/mic-hud-probe.mjs`: no SFU globals, click ON→OFF→ON, glyph slash + tooltip following, no page errors; with pre-delegation micstate it fails at the first click with the stuck "mic off" state).

**2 — release stops only the raw device, never a distinct lane.** The `(_raw || _lane)` fallback stopped the destination track on any release after the raw was gone — the one-way door the function's own rule (2) forbids, causing exactly the renegotiation it promises to prevent. Now: a present `_raw` always stops (rule 1 is a privacy promise — the OS indicator dies even where raw===lane on synthetic/gate-unavailable paths), repeated release is harmless by construction, and reacquisition reconnects into the identical lane object. The new vectors use an AudioContext stub full enough to build the **real gate graph**, with an asserted anti-vacuousness check (lane genuinely distinct from raw) — because as you noted, the exec test calls `gateRelease()` first and structurally cannot see this class. Mutant: the original release fails 3 lane-preservation vectors.

**3 — owned harness, receipts that can fail, portable browser.** `tools/probe-harness.mjs`: browser = `SFU_TEST_CHROME` override, else Playwright's **managed** browser — playwright is now a declared devDependency, so a clean checkout runs `playwright install chromium` and gets the right build per platform; child ownership = per-run nonce + scratch state + verified readiness + teardown (extracted from boot-check, which now rides it). `/version` echoes `EIDO_BOOT_NONCE` when set, so ownership is *provable* rather than resting on startedAt freshness (absent when unset; production `/version` unchanged), and the port range is wide with the nonce as the backstop against concurrent-run cross-contamination. Of the seven flagged probes: boot-check converted; **three restored and ported** (meter latch, panel teardown, tts chunking — they assert behaviors this branch ships; panel-teardown's selectors and population-dependent gate were also ambient-world artifacts, now keyed to the panel's own controls so a lone-resident owned child suffices); three deleted as assertion-free console dumps, replaced by mic-hud-probe. I'll be straight that my first cut deleted all six with an overbroad rationale — the second-agent review corrected it, and the restore is in the head you're reviewing.

No feature-direction changes; the runtime delta is micstate's transport seam and the `/version` nonce echo.

antra-tess (changes requested):
Exact-head re-review at `089a8442b6908fd7f678a04607d4a1315a00d61a`: **CHANGES_REQUESTED**.

The three requested repairs are individually present and their focused/browser receipts are useful:

- micstate exec: **10/10**
- release/lane identity: **14/14**
- mesh fallback + SFU precedence + level/gate delegation: **13/13**
- spoken-form: **11/11**
- owned browser boot: green (9 panels, zero page errors)
- real HUD mesh ON→OFF→ON: **6/6**
- meter states/hysteresis: pass
- panel teardown: every row survives **11/11** updates
- TTS blocking: chunked, page yields between pieces
- browser bundle: green (standing Three/WebGPU warnings only)
- clean current-main trial merge / `git diff --check`

But the composed branch regresses the existing voice stack on both the exact PR head and a clean merge into current main. Independent current-main control:

```text
voice-lifecycle-test: 124/124
```

Exact #131 and current-main + #131 both fail at the first mic cycle:

```text
✗ mic off stops the local track
✗ mic off silences the RAW source and closes the gate
  rawOff=false senders=0
```

and then abort before the rest of the matrix with:

```text
SyntaxError: Export named 'assignColors' not found in .../client/lib/core.js
```

Current main alone passes all 124. This is therefore introduced by the slice/composition, not standing debt. The new focused vectors prove the extracted micstate seams in isolation; they do not prove coexistence with the still-live mesh module the PR promises to preserve until cutover.

Required repair/receipt:

1. make current-main + #131 pass the existing **124/124** lifecycle suite unchanged (mic ON→OFF must close the real raw source while retaining sender lane identity, and no module/mock graph crash);
2. keep the focused release and mesh-fallback discriminators green;
3. add the current-main composed lifecycle suite to the PR's traveling receipt so isolated seam tests cannot hide the integration regression;
4. in `ownedWorld()`, remove the pre-nonce `startedAt` fallback for an owned exact-head child. This branch's `/version` now echoes `EIDO_BOOT_NONCE`; accepting a just-started pre-nonce listener reopens a small false-green race. Explicit `live:` mode already covers probing deployments that are not owned children.

No redesign of the mic/panel/TTS feature set requested. Once the extracted path and current mesh coexist in the real lifecycle suite, I expect this slice to be close.

octopusburrow:
New exact head `3833a2a`. You were right on all counts — the composed branch had three real coexistence defects the focused seam tests structurally could not see, and each is now fixed at its root rather than in the suite (which rides unchanged):

**1. The mic-off regression was micgate's lookahead, built unguarded.** `gateStream` called `createDelay` unconditionally; on a WebAudio context without it (your suite's minimal fake — and any constrained embedder) the throw landed in the fail-closed catch, converting "no 50ms lookahead" into the NO-GATE path: mic on yielded no lane, no raw stream, no senders — the exact `rawOff=false senders=0` you measured. The lookahead now builds only where `createDelay` exists, inside its own try, and degrades to `src→gain` with the gate fully live: an enhancement's absence costs exactly the enhancement. (`attachSource` already routed conditionally; and a second-agent pass of this fix found the same disease one function over — `setMonitor`'s own unguarded `createDelay`, which made the self-monitor checkbox throw uncaught on the same context class; guarded identically in this head, degrading to zero-latency monitoring.)

**2. The `assignColors` abort was stt.js's static `logChat` import** dragging `chat.js` — and core.js's full surface — into every context that mocks core. Lazy now, the same pattern micstate uses for `flashHint`; a transitive-import audit of all twelve touched modules confirms no second path pulls chat (or another heavy module) into the mocked voice graph.

**3. The panel's house-token restyle had renamed `.sp-row` to `.row wide`**, silently breaking the markup contract the lifecycle suite (and field tooling) pin. All three row factories now carry `sp-row row wide` — contract and styling both.

**Item 3 — the composed suite travels.** `bun tools/voice-lifecycle-test.ts` on this head: **124/124**, stable across repeated runs, now a standing member of this PR's receipt set. Trial merge onto current `origin/main` (`756ebab`): clean, merged-tree lifecycle **124/124**.

**Item 4 — `ownedWorld()` is nonce-only.** The pre-nonce `startedAt` fallback is gone: this branch's `/version` echoes `EIDO_BOOT_NONCE`, so an owned child always proves itself by nonce, and a responder without the field fails immediately as a stale pre-nonce listener — the just-started-impostor race is closed. `live:` mode remains the explicit path for deployments that are not our children.

Full roster at this head: composed lifecycle `124/124` ×3 · micstate exec `10/10` · release `14/14` · mesh-fallback `13/13` · spoken-form `11/11` · boot-check (9 panels, owned child) · mic-hud-probe `6/6` · meter states+latch · panel teardown (all rows survive) · TTS chunking · client bundle green · clean trial merge. Mutant receipts: re-breaking each fix reproduces the exact failures you reported (unconditional `createDelay` → the two mic-off vectors + one dependent; static `logChat` → the `assignColors` abort; dropping `sp-row` from the checkbox row → the hear-voices vectors).

antra-tess (approved):
Approved exact head `3833a2a53b8d2ca8a4f668aead4de98877423d6b`.

Independent receipts on macOS/Bun 1.3.14 with the repository's declared root + `client/` dependency trees and explicit Chrome-for-Testing:
- current-main trial merge clean; composed voice lifecycle `124/124`
- exact-head micstate exec `10/10`, release/lane identity `14/14`, mesh fallback/SFU precedence `13/13`, spoken form `11/11`, AudioContext `7/7`
- owned nonce child browser: boot 9 panels/no page errors; HUD mesh ON→OFF→ON `6/6`; meter three states+hysteresis; panel rows survive `11/11` each; TTS page remains responsive across 5 chunks
- TTS faults `29/29`; panel identity, self gain, mic precedence, select/resume, empty/in-place lists, import order, and host-class discriminators all green
- `git diff --check` clean

The prior browser blocker was review-environment setup: the owned server's import-map packages all 404 when the worktree lacks its nested `client/node_modules`; wiring the declared client dependency tree yields the receipts above. Root bundle still fails on current main and the merged tree alike at the standing missing `shared/structure.js`, so it is not introduced here.

Approval covers this standalone current-mesh mic/TTS foundation only. It does not authorize SFU cutover, #132 integration, merge, or deployment.

antra-tess:
Merged after clean rebase onto main following #129 via integration PR #139; voice lifecycle 124/124.

PR #134, opened 2026-08-17
fix collider bucket overflow and stale refits

HomunculusLabs:
## Summary

- cap collider spatial-bucket insertion at 4,096 cells so corrupt or mis-scaled bounds cannot drive an unbounded allocation loop
- remove prior bucket memberships before replacing headless support-box or support-grid entries
- preserve the existing fail-safe behavior: pathological colliders abstain while the client keeps running

## Why

A pathological footprint can expand into thousands of spatial cells and exhaust the process. Separately, repeated `fitSupportBox` / `fitSupportGrid` calls replaced the collider map entry without removing the old entry's bucket memberships, leaking stale cell references across refits.

## Tests

- `bun run tools/collider-test.ts` — 34 passed, 0 failed
- `bun run tools/support-test.ts` — 12 passed, 0 failed
- `bun run tools/supportclass-test.ts` — passed

### DIFF
--- a/client/lib/colliders.js
+++ b/client/lib/colliders.js
@@ -127,6 +127,16 @@ function bucketAdd(id, entry) {
   entry.cells = [];
   const x0 = Math.floor((obj.position.x - r) / CELL), x1 = Math.floor((obj.position.x + r) / CELL);
   const z0 = Math.floor((obj.position.z - r) / CELL), z1 = Math.floor((obj.position.z + r) / CELL);
+  // pathological-footprint guard (#74 era): a box whose cell span is absurd
+  // (misread scale, corrupt bbox) once OOM-killed the whole daemon — 560
+  // RangeErrors deep in this loop before KeepAlive masked them. The abstain
+  // doctrine from the seam probes applies here too: refuse the box, keep
+  // the process. Same bucketRemove contract as a normal entry (no cells).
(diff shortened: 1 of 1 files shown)


slimepriestess (approved):
**Approve.** Both changes read correctly in context:

- `bucketAdd` sets `entry.cells = []` *before* the new span guard, so an abstained entry still satisfies `bucketRemove`'s contract (no cells, nothing to release). `r` and `s` used in the log line are the function's own locals. The 4096-cell cap is 64×64 cells of 8 m — a 512 m footprint — nothing legitimate is near it.
- The two re-fit sites now release the previous entry's cells before `colliders.set`; `bucketRemove(id, undefined)` is a no-op via `entry?.cells ?? []`, so a first fit is unchanged.

`bucketAdd` now returns a boolean that no caller reads (`:323/:355/:387/:399/:419`) — fine, but it's a return value with no consumer yet.

**Receipt:** `bun run tools/collider-test.ts` → 34 passed / 0 failed on this head, and the same 34/0 on `main` — i.e. the suite does not exercise either change. Nonblocking: a check that a pathological box abstains (`bucketAdd` returns false, map still has no cells) and one that a re-fit of the same id leaves exactly one set of cells would pin both, and they're two lines each in `collider-test.ts`.

PR #137, opened 2026-08-18
glb2vrm: refuse foreign rigs by family, not by first missing bone (#123)

Michael-Andrzejewski:
The immediate fix both mica and antra scoped on #123: capability-gating plus a truthful error. Real Mixamo support and the semantic humanoid map from antra's comment stay open; nothing here renames or rerigs anything, per mica's constraint.

## What changed

- A rig-family gate runs right after the GLB parses, ahead of both the convert and `--measure` paths. Any `mixamorig`-prefixed joint (colon kept, stripped, or numerically suffixed) gets a refusal that names the actual contract: Mixamo rigs are not supported; this converter assumes Tripo bone names, A-pose rest, and Tripo skeleton structure. Detection happens before any conversion is attempted, which is where the issue said the rejection belongs.
- A rig that is neither Mixamo nor Tripo now reports every missing Tripo bone and asks "not a Tripo rig?" (the phrasing import-tripo-avatar already uses), instead of dying on the first failed lookup.
- VRM 1.0 inputs skip the gate; they carry their own humanoid map.

## Repro provenance

Verified against three.js's Xbot.glb, a genuine Mixamo export with the `mixamorig:` prefix intact. Before this change, convert dies with `node not found: Hip` and `--measure` dies with `node not found: L_ToeBase` (same class of misleading error, different bone; the `--measure` path is not mentioned in the issue text). After, both paths exit 1 with the honest message. Two notes for the issue thread: line references there have drifted (the fail-fast loop sat at glb2vrm.ts:279 on current main), and Xbot.glb is a freely available reproducible rig for the alias table when option 1 happens.

## Tests

`bun tools/glb2vrm-test.ts` drives the real CLI against synthetic GLBs (a JSON chunk alone is a valid GLB, which is enough for name-based family detection): 15 checks covering both paths, three prefix variants (`mixamorig:Hips`, `mixamorigHips`, `mixamorig1:Hips`), the unknown-rig contract message, and that a Tripo-named rig passes the gate. All pass. One gap named honestly: we had no raw Tripo GLB on hand to run a full conversion end to end, so the gate-pass case asserts only that a Tripo-named rig is not refused as a foreign family.

Co-authored with Spark, a Claude (Fable) instance working alongside me.

### DIFF
--- a/tools/glb2vrm-test.ts
+++ b/tools/glb2vrm-test.ts
@@ -1,0 +1,114 @@
+// glb2vrm rig-family gate (#123), run headless.
+//
+//   bun tools/glb2vrm-test.ts
+//
+// The gate exists because a Mixamo rig fed to the converter used to die on
+// its first missing Tripo bone ("node not found: Hip"), naming a bone the
+// user never chose instead of the actual contract. These tests synthesize
+// minimal GLBs (a JSON chunk is a valid GLB; no geometry needed to test
+// name-based family detection) and drive the real CLI, asserting on the
+// message and exit code. Real-rig provenance: the gate was verified by hand
+// against three.js's Xbot.glb, a genuine Mixamo export with the colon kept.
+
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+
+let pass = 0, fail = 0;
+const check = (name: string, ok: boolean, detail = "") => {
+  if (ok) { pass++; console.log(`  \x1b[32m✓\x1b[0m ${name}`); }
+  else { fail++; console.log(`  \x1b[31m✗\x1b[0m ${name}${detail ? ` — ${detail}` : ""}`); }
+};
+
+function makeGLB(nodeNames: string[]): Uint8Array {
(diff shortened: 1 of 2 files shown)


slimepriestess (approved):
**Approve**, with one behavior change to be aware of.

The gate is where the issue asked for it (after parse, before any surgery), matches the prefix case-insensitively with the colon/suffix variants covered, skips VRM 1.0 inputs that carry their own humanoid map, and replaces the old first-missing-bone crash with the full missing list and the "not a Tripo rig?" phrasing `import-tripo-avatar` already uses. Nothing renames or rerigs.

**Receipt:** `bun tools/glb2vrm-test.ts` on this head → 15 passed / 0 failed (synthesized JSON-chunk GLBs driving the real CLI; the Xbot.glb hand-check is the author's).

**Nonblocking, but say it in the PR body:** the gate now runs *before* the `--measure` branch, whose own comment says "Report conventions of any glb/vrm — used to sanity-check against library avatars." On `main` the fail-fast bone loop sat after the measure block, so `--measure` worked on a Mixamo or otherwise foreign GLB; on this head it exits 1 (the test pins that). If measuring foreign rigs is a use anyone has, a `--measure`-only bypass (or `--force`) keeps it; if not, update the comment so it stops promising "any".

PR #142, opened 2026-08-24
store: uploads get the KTX2 shadow too — and a flagged fall-through is provisional, never immutable

gabriel-berzescu:
Mica — Antra suggested (on X, to @gabriel_xnlg) that server-side texture compression on uploaded assets would help lower-end machines, and asked to have you review. Reading the tree first: the §20 arc already *is* that — for the library. This closes the gap it left in the store, and fixes the serving rules that would have kept the gap closed even after the fix.

*Round two (`663885d`) addresses the review's four points — see the thread. Summary below is current.*

## The gap

The KTX2 sweep (`upload.ts`) walks library models, avatars and the curated image dirs — never `store/`. Conjured props (Tripo/Orrery — most of what is in commons) only ever get the webp shadow. #122's own evidence is the shape: `store/305ea…glb?ktx2=1` → "draco + webp, and no KTX2 at all". §20.1 records why the store arm was left out — `basename()` collides for library rels — which doesn't apply to hash-named store files.

Why it's the bill a 6GB card feels first: webp decodes to full RGBA8 on the GPU (a 1024² map + mips ≈ 5.3MB of VRAM); KTX2 stays block-compressed (≈1.4MB) and skips `createImageBitmap` — §20's own 4–8×.

## What this does

- **`server/store-variants.ts`** (new, pure): `isStoreOriginal`, `isKtx2Variant`, `isServingArtifact`, `ktx2VariantPath`, `storeShadowsMissing`.
- **`shared/ktx2.js`** (new): the negotiation key as a **generation** — `KTX2_KEY = '2'`, `wantsKtx2`, `withKtx2` — imported by both the client (four fetch sites) and the sequencer. Every flagged store answer had been served `immutable` under `=1`, webp included, and no origin header can revalidate a browser's immutable entry; a fresh key is a fresh cache entry everywhere at once. A retired key is an unflagged fetch.
- **`upload.ts`**: `queueOptimize` also queues a `--ktx2` item per upload, from the **original** (the diet encodes from the best bytes; store-min has been wrong, #122), to `store/<hash>.glb.ktx2.glb` — the path `routes.ts` already resolves. The boot sweep queues an original lacking either shadow, never a variant. Honors `ktx2Skip`. Handles the new exit 5.
- **`optimize.ts`**: `optimizeGlbKtx2` is **all-or-nothing** — `ktx2CompressTextures` returns a tally and checks every encoder output for the KTX2 container magic; the GLB arm writes nothing unless every eligible texture converted. CLI **exit 5** = an encoder answered but the variant is refused: environmental (no `.failed`, retried next boot), not `ktx2Skip` (the encoder is there). Exit 2 stays "nothing to convert". On `main` the CLI would write a `.ktx2.glb` with png inside and serve it immutable (reproduced).
- **`routes.ts`**, three things:
  1. `/library-models` filters with `isStoreOriginal` — the ghost-listing rule (§20c) for the store.
  2. `/library-list` skips `isServingArtifact`: variants, `.failed` markers, `.tmp`. The prefetcher pushes every listed store path as a fetch — a listed variant downloads twice, a listed marker fetches as a model.
  3. **A flagged fetch that falls through is provisional** — `no-cache`, never `immutable`. Content-addressed makes the *address* immutable, not the flagged answer. `no-cache` rides the existing ETag (the `.vrm` doctrine in `serveFrom`): a 304 while nothing changed, the variant's bytes the moment it exists. Also closes the latent `.vrm` case — a stale variant under a fresh `?v=` fell through to the original *immutable* until the next re-upload.
- **`docs/ktx2-encoder.md`**: the Linux + systemd recipe (macOS and Windows were documented; the platform production runs on was not), and where the key lives.

## The show box, measured from outside (2026-08-24)

Every model probed on `eidoverse.animalabs.ai` with `?ktx2=1` — library and store — came back `extensionsRequired: [EXT_texture_webp, KHR_draco_mesh_compression]`, images `image/webp`; the store ones `cache-control: public, max-age=31536000, immutable`. `/library-list` shows 0 `.ktx2` entries across 118 library + 39 store files. So: **no encoder on the VPS; the whole §20 arc is built and dormant there.** When this lands: install the encoder (`docs/ktx2-encoder.md`, Linux section; `Environment=KTX2_TOKTX=…` in the unit) — the boot sweep does the rest, serially, off the request path. The key rotation takes care of what clients and nginx already pinned.

## The trade, stated

While no variant exists for a path (prod today), a flagged fetch revalidates per page load (a 304, no body) instead of caching. That's what avatars already do; it ends per-path the moment the variant lands. One string in `serveFrom` if you'd rather bound it.

## Receipts

`bun tools/store-variants-test.ts` — **83/83**, **no encoder needed** (Windows, bun 1.4.0; the one real-encoder check runs when toktx is present):
- the pure contract, incl. the serving-contract check that names the gap on main (routes resolved `${rel}.ktx2.glb`; nothing wrote it), and the key generation;
- the ghost-listing rule, all three predicates;
- the optimizer, spawned exactly as the pump spawns it, against a **fake toktx** (a bun script behind a `.cmd`/sh wrapper, `FAKE_TOKTX_MODE`): ok → a real KTX2 variant; all-fail → exit 5, nothing written; partial-fail (1/2) → exit 5, nothing written; exit-0-but-not-KTX2 bytes → exit 5; untextured → exit 2; corrupt → exit 1; no encoder → exit 3;
- the real sequencer, spawned with the fake in its environment: `POST /upload` → queue → variant served under the current key, immutable; retired key → an unflagged fetch; `/library-models` once, `/library-list` no variant / `.failed` / `.tmp`; an untextured upload → the pump's `.failed` marker; a pre-placed original swept at boot into a REFUSED partial — no variant, **no marker**, no `ktx2Skip`, flagged fetch provisional meanwhile; no encoder at all → exit 3 **once**, `ktx2Skip` holding across a later upload; and the `If-None-Match` round-trip — 304 while nothing changed, **200 with the variant the moment it lands**, 304 immutable after.

Also green on the patched routes/optimizer: `one-libvips-test.ts` 11/11, `deps-route-test.ts` 13/13; `git diff --check` clean. (`version-route-test.ts` crashes identically on clean `main` on this Windows box — its line 157, the PATH-without-git scenario — unrelated.) A real 34.5MB library GLB through `--ktx2` here: 20.5MB in 87s.

Negative control: the suite dies at import on `main` (no `store-variants.ts`, no `shared/ktx2.js`); of the serving checks, `main` fails "flagged → variant" and "fall-through is provisional"; of the optimizer checks, `main` writes the empty and the partial variant.

## Not in this PR

Decimated LODs (Antra's other suggestion). No mesh LOD exists today; `import-tripo-avatar.ts:180` already has a proven `weld()+simplify(MeshoptSimplifier)` recipe, and the client's residency tier (placeholder ↔ real) is the natural hook — but that's a change in the realizer and a design call about where selection lives. Happy to do it as a follow-up; tell me where you'd want the selection to sit.

`upload.ts` gains an `exit 5` branch *after* the `exit 3` one; #124's `exit 4` lands *before* it — adjacent, not overlapping. I'll rebase on whichever merges first.

### DIFF
--- a/client/lib/assets.js
+++ b/client/lib/assets.js
@@ -8,6 +8,7 @@
 //   vrmPool    whole parsed VRM instances at rest (§19b — no clone exists
 //              for a bound rig, so released bodies are reworn intact)
 
+import { withKtx2 } from '../../shared/ktx2.js';
 import { THREE, renderer, camera, scene, report, bus } from './core.js';
 import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
 import { DRACOLoader } from 'three/addons/loaders/DRACOLoader.js';
@@ -148,7 +149,7 @@ const draco = new DRACOLoader().setDecoderPath('https://www.gstatic.com/draco/ve
 const ktx2 = new KTX2Loader()
   .setTranscoderPath('/node_modules/three/examples/jsm/libs/basis/')
   .detectSupport(renderer);
-/** Whether this GPU/browser negotiates KTX2 variants (?ktx2=1) — prefetch
+/** Whether this GPU/browser negotiates KTX2 variants (?ktx2=<key>, shared/ktx2.js) — prefetch
  *  must warm the SAME cache key demand fetches will use. */
 export const ktx2Capable = () => !!ktx2.workerConfig;
 function makeLoader(vrm = false) {
@@ -283,9 +284,8 @@ export async function loadVRM(libPath, { priority = 1 } = {}) {
     // byteCache (variant and original are distinct byte entries — correct),
     // while the vrmPool and its vrmMeta ledger key on libPath UNTOUCHED, so
     // pool identity is unaffected by negotiation.
-    const flag = libPath.split('?')[0].endsWith('.vrm') && ktx2.workerConfig
-      ? (libPath.includes('?') ? '&ktx2=1' : '?ktx2=1') : '';
-    const buf = await fetchBytes(`/library/${libPath}${flag}`);
+    const url = libPath.split('?')[0].endsWith('.vrm') && ktx2.workerConfig ? withKtx2(libPath) : libPath;
+    const buf = await fetchBytes(`/library/${url}`);
     work.phase('queued');
     // The parse and skeleton passes are the irreducibly-synchronous chunk of a
     // body: serialize so two arrivals can't stack theirs into the same frames,
@@ -477,8 +477,8 @@ export async function loadGLB(libPath) {
         // paths — the server answers with the variant when one exists, the
         // original otherwise. The full URL keys byteCache, so variant and
         // original are distinct entries, which is correct.
-        const flag = libPath.endsWith('.glb') && ktx2.workerConfig ? '?ktx2=1' : '';
-        const buf = await fetchBytes(`/library/${libPath}${flag}`);
+        const url = libPath.endsWith('.glb') && ktx2.workerConfig ? withKtx2(libPath) : libPath;
+        const buf = await fetchBytes(`/library/${url}`);
         work.phase('queued');
         return await enqueue(async () => {
           work.phase('parse');
@@ -775,9 +775,8 @@ export async function primeFiles(paths, { concurrency = 6 } = {}) {
         // cache (§16.2.B), and every toolkit module see the same identities —
         // the bytes just arrive GPU-native, and loadImageTexture sniffs the
         // container magic to tell which shape it got.
-        const flag = ktx2Capable() && /\.(png|jpe?g)$/i.test(p)
-          ? (p.includes('?') ? '&ktx2=1' : '?ktx2=1') : '';
-        const buf = await fetchBytes(`/library/${p}${flag}`);
+        const url = ktx2Capable() && /\.(png|jpe?g)$/i.test(p) ? withKtx2(p) : p;
+        const buf = await fetchBytes(`/library/${url}`);
         denoFiles.set(p, new Uint8Array(buf));
       } catch (e) { missing.push(p); }
     }
--- a/client/lib/prefetch.js
+++ b/client/lib/prefetch.js
@@ -21,6 +21,7 @@
 //      policy — assets that get USED are the ones that stay warm, and a newly
 //      needed asset displaces stale speculative ones by construction.
 
+import { withKtx2 } from '../../shared/ktx2.js';
 import { bus, CONFIG, report } from './core.js';
 import { whenBooted } from './boot.js';
 import { whenCalm } from './governor.js';
@@ -105,7 +106,7 @@ async function buildQueue() {
     const bare = url.split('?')[0];
     if (ktx2Capable() && (/\.(glb|vrm)$/.test(bare)
       || (/^\/library\/eidoverse\/assets\/(grass|sky|particle_textures)\//.test(bare) && /\.(png|jpe?g)$/i.test(bare)))) {
-      url += (url.includes('?') ? '&' : '?') + 'ktx2=1';
+      url = withKtx2(url);
     }
     if (!seen.has(url) && seen.add(url)) q.push({ url, size });
   };
--- a/docs/ktx2-encoder.md
+++ b/docs/ktx2-encoder.md
@@ -72,6 +72,13 @@ lands.
 
 ## Either platform
 
+The negotiation key a client sends (`?ktx2=<key>`) is a generation, defined
+once in `shared/ktx2.js` and imported by both the client and the sequencer.
+Bump it when a flagged answer has been served with the wrong caching — the
+old key's cache entries are simply never asked for again (a purge cannot
+reach browsers). The fall-through answer under the current key is served
+`no-cache` until a variant exists, then the variant is served immutable.
+
 `KTX2_TOKTX` may point at either `toktx` or the newer `ktx` binary — the
 probe (`findKtx2Encoder`) detects which by name. Without the env it also
 checks PATH for both. The sweep runs once per boot, serial, ~5-30s per
--- a/docs/ktx2-encoder.md
+++ b/docs/ktx2-encoder.md
@@ -46,6 +46,30 @@ curl -LO https://github.com/KhronosGroup/KTX-Software/releases/download/v4.4.2/K
 # binaries at ktx/bin/toktx.exe — point KTX2_TOKTX at it
 ```
 
+## Linux (the show VPS is Ubuntu) — the .deb, or portable
+
+The release ships `Linux-x86_64` as `.deb`, `.rpm` and `.tar.bz2`. The .deb
+puts `toktx` and `ktx` on PATH, which is all the probe needs:
+
+```sh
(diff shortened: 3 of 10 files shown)


antra-tess (changes requested):
Independent review at exact head `684a004b597b85409c14ec4a7f60542ea1419ed0`: **REQUEST_CHANGES**. The adjacent-shadow and provisional-cache direction is good, and the exact-head store route suite passes, but four seams need repair before this can safely become the production KTX2 producer.

1. **P1 — rotate the negotiation cache key.** `routes.ts` changes headers only for future origin responses, while clients continue requesting the already-poisoned `?ktx2=1` URL and nginx keys on the full query string. Existing browser/nginx entries were served `max-age=31536000, immutable`; they will not revalidate merely because the server now says `no-cache`. Use a fresh key such as `ktx2=2` (and accept it server-side), rather than relying on a one-time nginx purge that cannot clear browser caches.

2. **P1 — refuse false/partial KTX2 shadows and distinguish environmental encoder failure.** `ktx2CompressTextures()` swallows each texture encode failure, and `optimizeGlbKtx2()` ignores the converted count. I reproduced the function returning a GLB whose only required extension was Draco and whose image remained `image/png` after the conversion path failed. On sufficiently compressible geometry the CLI can therefore write an immutable `.ktx2.glb` containing no KTX2 at all — exactly the class this PR is meant to eliminate. Require a validated conversion result (at minimum nonzero; ideally every eligible referenced raster image converted), and treat unusable-encoder/configuration failures as retryable environmental failures rather than permanent `.failed` content verdicts. Add all-fail and partial-fail negative controls.

3. **P2 — do not list/prefetch `.failed` artifacts.** `/library-list` excludes only `isKtx2Variant()`, while `hash.glb.ktx2.glb.failed` deliberately returns false from that predicate. The prefetcher pushes every store listing entry, so diagnostic markers become asset fetches. For `dir=store`, list originals only, or explicitly exclude `.failed`/`.tmp` serving artifacts.

4. **Receipt gap — exercise the queue/failure seams, not only helpers and pre-created files.** The new test directly spawns `optimize.ts`, skips encode when no encoder exists, then pre-creates original/variant/marker before starting the sequencer. It does not test POST `/upload`, boot queuing, `ktx2Skip`, exit 1/2/3 marker behavior, or an actual conditional `If-None-Match` swap. A fake encoder plus an owned upload/boot integration should make those contracts executable.

Independent receipts: exact head `tools/store-variants-test.ts` passes `0 failed`; clean current-main trial merge passes the same owned real-sequencer route matrix; `git diff --check` is clean. The real encode section was skipped because this Mac has no KTX2 encoder. Neighboring `one-libvips-test` is red on both head and current main (`1 failed`), so I did not attribute that standing failure to this PR.

gabriel-berzescu:
Mica — thank you; all four are right, and #2 is the best catch in the set. Repaired at `663885d`, each under test now:

**1. Key rotation — done, and it is client change after all.** `shared/ktx2.js` defines the key as a generation (`KTX2_KEY = '2'`, `wantsKtx2`, `withKtx2`), imported by the four client fetch sites and the one server branch — no more mirrored literals. A retired key (`=1`) is an unflagged fetch: whatever a client pinned under it, it keeps, and nothing asks with it again. You're right that no origin header reaches a browser's immutable entry; the purge note is gone from the description.

**2. False/partial variants — refused, retryably.** Reproduced first: on `main` the CLI wrote both the empty and the partial `.ktx2.glb`. Now `ktx2CompressTextures` returns a tally (eligible / converted / failed-by-name) and checks every encoder output for the KTX2 container magic before accepting it; `optimizeGlbKtx2` writes nothing unless every eligible texture converted. New CLI **exit 5** = an encoder answered but the variant is refused — environmental like 3 (no `.failed`, retried next boot), but *not* `ktx2Skip` (the encoder is there). Exit 2 stays "nothing to convert", the marker's business. Controls: all-fail, partial-fail (1/2), exit-0-but-not-KTX2 bytes, untextured, corrupt, no-encoder.

**3. `.failed` in listings — fixed.** `/library-list` skips `isServingArtifact` (variants, `.failed`, `.tmp`), not just variants. Asserted against a running server with all three beside the upload.

**4. Receipts — the seams, executed.** `tools/store-variants-test.ts` is now 83 checks and needs **no encoder**: a fake `toktx` (a bun script behind a `.cmd`/sh wrapper, `FAKE_TOKTX_MODE`) emits a canned real KTX2 or refuses. The optimizer is spawned as the pump spawns it for every failure mode above; then the real sequencer is spawned with the fake in its environment: `POST /upload` → queue → variant served under the current key, retired key an unflagged fetch, markers unlisted; a pre-placed original swept at boot into a REFUSED partial with no marker; no encoder at all → exit 3 once, `ktx2Skip` holding across a later upload; and the `If-None-Match` round-trip — 304 while nothing changed, 200 with the variant the moment it lands, 304 immutable after. Fixtures are content-hashed with a per-run nonce (ownership), land in this checkout's `assets/opt`, and leave with their manifest entries.

One thing the fake taught me on its first run: `Bun.which("toktx")` resolves `toktx.cmd` through PATHEXT on Windows, so "no encoder" for the child needs an empty `PATH`, not just a dead `KTX2_TOKTX`. Noted in the test.

`deps-route-test` 13/13 and `one-libvips-test` 11/11 here on the patched routes/optimizer; `git diff --check` clean. Your `one-libvips` red on the Mac matches what #122/#136 describe for that libvips.

The trade in the description stands: while no variant exists for a path, a flagged fetch revalidates per load (a 304) rather than caching. One string in `serveFrom` if you'd rather bound it.

antra-tess (approved):
Re-review at exact head `663885d11be74c148f3c9f5888f9f7bb6f2bf395`: **APPROVE**. All four requested seams are now closed.

- Negotiation is generation-keyed in `shared/ktx2.js`; clients/prefetch/server moved to `ktx2=2`, so previously pinned immutable `=1` entries are abandoned rather than expected to revalidate.
- The producer now refuses zero/partial/non-KTX2 conversions (exit 5), writes no false variant or sticky marker for that environmental shape, and retains exit 2 for real content verdicts / exit 3 for absent encoder.
- Listings exclude variants, `.failed`, and `.tmp` artifacts, so store prefetch sees originals only.
- The expanded owned integration matrix exercises fake-encoder success/all-fail/partial-fail/bad-container, real POST `/upload` queueing, boot sweep behavior, exit 2/3/5 markers, current-vs-retired key serving, and an actual `If-None-Match` provisional→variant swap.

Independent receipts:
- exact-head `tools/store-variants-test.ts`: `0 failed`;
- clean current-main trial merge: automatic, same matrix `0 failed`;
- `git diff --check`: clean;
- hostile static review's four findings rechecked against the revised code/tests.

The only missing field receipt is a run against a real `toktx` binary on this Mac; the deterministic fake encoder now exercises the producer and failure semantics, and production installation/canary remains a deployment operation rather than a code-review blocker. Neighboring `one-libvips-test` remains `1 failed` on both current main and this head (`none loaded`), so it is standing environment/test debt, not introduced by #142.

Approval is code review only — no merge or deployment authority inferred.

PR #143, opened 2026-08-25
ktx2: the browser takes the negotiation key from the running sequencer, never from a served file

gabriel-berzescu:
Mica, Antra — the follow-up from the `=2` collision (89d72f5), as agreed in-world: the browser takes the negotiation key from the **running** sequencer, never from a file it merely serves.

## The window this closes

Between `git pull` and the restart, the old process serves the new `shared/ktx2.js`. A client that read key 2 off that file asked a server that had never heard of it; the server answered as if unflagged — webp, immutable — and nginx pinned it under `?ktx2=2`. Rotating to 3 was the right emergency lever; this makes the emergency impossible.

## Mica's four conditions, point by point

1. **The capability response is `no-store`** — it *is* `/version`, which already was. `ktx2Key` rides in that body.
2. **Bound to the same server/build epoch** — same document as `sha` / `commitTime` / `startedAt`: one fetch, one epoch, resolved at *that* process's boot. There is no second document that could drift from the first.
3. **Fails safely to unflagged originals** — `keyFromVersion` returns a validated key or `null`, never a default; `negotiate(url, null)` is the bare URL. An older sequencer (no field), a failed fetch, a garbage body → the client does not negotiate at all, which is always the right answer for an unflagged URL.
4. **A canary proving pull-before-restart cannot poison the next key** — `tools/store-variants-test.ts` §8 replays the incident against a live child: the sequencer runs with key 3; the test rewrites `shared/ktx2.js` on its disk to key 99 (the pull); `/shared/ktx2.js` now serves 99 while `/version` still publishes 3; a client keyed from `/version` gets the variant, immutable; a client keyed off the served file asks `?ktx2=99` and gets unflagged-immutable — the poison, demonstrated, and never asked for by the new path. The file is restored byte-for-byte (`finally` + exit hook).

The static constant stays as the server's definition and the tests' fixture; `withKtx2` says in its comment that a browser must not call it.

## What changes

- **`server/routes.ts`**: `/version` publishes `ktx2Key`.
- **`shared/ktx2.js`**: `keyFromVersion(json)`, `negotiate(url, key)`; contract restated in the header.
- **`client/lib/assets.js`**: `ktx2KeyReady` — one `no-store` fetch of `/version` per page, awaited by the three negotiating loads (all async already).
- **`client/lib/prefetch.js`**: awaits the same key once before building its queue — warmth lands in the same cache entry as demand fetches, or nowhere flagged at all.
- `docs/ktx2-encoder.md`: the contract, restated.

Server-side negotiation (`wantsKtx2`, provisional/immutable, variants) is untouched. Key rotation remains the lever for a poisoned cache; it just no longer needs a coordinated restart to be safe.

## Receipts

`bun tools/store-variants-test.ts` — **99/99**, no encoder needed. New since #142:
- `keyFromVersion`: the published key; an older sequencer's `/version` → `null`; garbage → `null`.
- `negotiate`: `?`/`&` appending; **no key → the bare URL**.
- every real-sequencer section now takes its key from the child's `/version` the way a browser does, asserts it equals the definition, and negotiates with *that* — upload→queue→variant, boot sweep, `ktx2Skip`, and the `If-None-Match` swap all pass under the browser's actual path.
- **§8, the canary** (above).

Also: `deps-route-test` 13/13; `bun build` and a transpile of the client modules clean; `git diff --check` clean; the tree is clean after the run (the canary's rewrite restored, fixtures and manifest entries removed).

Field receipt: this branch's sequencer, opened in a real Chrome on the sponsor's machine — `/version` answered `{"sha":…,"startedAt":"2026-08-25T05:55:36.978Z","ktx2Key":"3"}` to the browser.

### DIFF
--- a/client/lib/assets.js
+++ b/client/lib/assets.js
@@ -8,7 +8,7 @@
 //   vrmPool    whole parsed VRM instances at rest (§19b — no clone exists
 //              for a bound rig, so released bodies are reworn intact)
 
-import { withKtx2 } from '../../shared/ktx2.js';
+import { keyFromVersion, negotiate } from '../../shared/ktx2.js';
 import { THREE, renderer, camera, scene, report, bus } from './core.js';
 import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
 import { DRACOLoader } from 'three/addons/loaders/DRACOLoader.js';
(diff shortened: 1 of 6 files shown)


antra-tess (approved):
Reviewed exact head c91d5d4c6a41de79a95f29a224c1f7abd0cbc8a0. APPROVE.

The running-sequencer authority is the correct fix for the pull-to-restart split brain: /version publishes the key imported by the live process with no-store, clients fail safely to unflagged URLs when absent, and demand/prefetch share one promise. Independent store-variants suite passed 0 failed, including the owned real-server canary that rewrites the served shared file under a running old process and proves /version keeps clients on the running key while the disk-read path reproduces the poisoned immutable answer. Server bundle and diff-check pass. Browser bundle still hits the standing unrelated ../../shared/structure.js resolution failure present outside this PR.

Merge review only; deployment should still verify public /version is no-store through nginx and a pull-before-restart staging canary.

PR #146, opened 2026-08-26
ktx2: the store variant keeps the shadow's texel budget — and a size verdict carries its recipe

gabriel-berzescu:
Mica, Antra — the sixteen store models the sweep never converts. Preregistered in devchat; this is the branch.

## What the markers say

Sixteen `not smaller` verdicts, e.g. `17600988 -> 26716692` (Mica's journal check — my JPEG-decoder guess was wrong). The `--ktx2` diet encodes from the full-resolution source: a Tripo conjure's three 2048² maps, two of them UASTC (a flat 8 bpp before zstd), and the "variant" comes out **1.5× the original on the wire** — which the 1.25× gate rightly refuses, with a permanent marker, so the count never moves across restarts.

Meanwhile the **unflagged** answer — the webp shadow every non-KTX2 client gets — is capped at **1024²** (`optimize.ts`, `resize: [1024, 1024]`; the library's day-one mirror is "draco+webp@1024" too). The flagged fetch was a silent 2048² quality upgrade over it at ~13× the bytes of the shadow. Same texel budget as the shadow, GPU-native, is the fix.

## What changes

- **`store-variants.ts`**: `KTX2_TEXEL_CAP = 1024`; `capTexels(size)` → the resize that fits (aspect kept, 4-aligned for the block format, never *up*), or `null`. And a recipe stamp — `KTX2_RECIPE`, `recipeStamp()`, `verdictStands(content)`: a size verdict is only as durable as the recipe that produced it, so the CLI stamps it and the sweeps re-measure any size verdict without the current stamp. Content verdicts (nothing to convert, a corrupt container) stand regardless. `storeShadowsMissing` reads the marker for that.
- **`optimize.ts`**: `ktx2EncodeArgs` takes `resize` and rides toktx's own `--resize` (applied before mip generation; libvips nowhere in the loop; `ktx create` has no equivalent here and logs that it encoded at source size). The size-gate refusal prints the stamp.
- **`upload.ts`**: the pump and both sweeps ask `verdictStands`; a landed variant retires its old marker.

The sixteen get re-measured on the first boot after this lands — no operator step.

## Measured

The 34.5 MB library car (`inanna_tech_…_vehicle_car_light.glb`), same box, same toktx 4.4.2:

| recipe | variant | ratio | encode |
|---|---|---|---|
| native source (main) | 20.5 MB | 1.7× | 87 s |
| **texel budget** | **1.9 MB** | **18×** | **10 s** |

Every texture 1024² in its KTX2 header.

**Preregistered prediction** for `0445768…` (17,600,988 bytes; refused at 26,716,692): the branch's `--ktx2` produces a variant **under 6 MB**, every texture 1024² by header. If that misses, this PR is wrong.

## Receipts

`bun tools/store-variants-test.ts` — **123/123**, no encoder needed. New:
- `capTexels`: 1024² and 512² untouched; 2048² → 1024²; 2048×1024 → 1024×512; 4096×3072 → 1024×768; 2048×1500 → 1024×752 (4-aligned); null in → null out.
- `verdictStands`: the show box's own marker text → stale; a current-stamped one → stands; an older recipe's → stale; "no convertible raster images" and a hard failure → stand. `storeShadowsMissing` with an injected reader: stale → missing (retry), current → not.
- the fake encoder now logs its argv: a 2048² source gets `--resize 1024x1024` once per texture, a 64² source gets none; a "bloat" mode drives the size gate and asserts the stamp in the refusal.
- the real encoder (when present): a 2048² source comes out 1024² by KTX2 header.
- a real sequencer: two pre-placed originals, one with a prod-shaped stale marker, one with a current-stamped one — the boot sweep re-measures the first into a served variant and retires its marker; the second is untouched.

Also green: `one-libvips-test` 11/11, `deps-route-test` 13/13, `git diff --check`, tree clean after the run.

Scope note: the cap applies to the whole `--ktx2` arm, library included — existing library variants are not rebuilt (the sweep only rebuilds when the source is newer), so nothing served today changes; new or re-measured ones come out at the shadow's budget. If you'd rather the library keep source resolution, it's a one-line branch on `rel.startsWith("store/")` — say so.

### DIFF
--- a/server/optimize.ts
+++ b/server/optimize.ts
@@ -27,6 +27,7 @@ import { NodeIO, type Document } from "@gltf-transform/core";
 import { ALL_EXTENSIONS, KHRTextureBasisu } from "@gltf-transform/extensions";
 import { dedup, prune, resample, textureCompress, draco, listTextureSlots } from "@gltf-transform/functions";
 import draco3d from "draco3dgltf";
+import { capTexels, recipeStamp } from "./store-variants.ts";
 import { existsSync, mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join, basename, dirname } from "node:path";
@@ -195,10 +196,14 @@ const COLOR_SLOT = /baseColor|emissive/i;
  *  (λ=1.0, quality-preserving) + zstd 18 — raw UASTC is a flat 8bpp and zstd
  *  alone barely dents it (the crow's 2048² maps came out 2.6× the JPEG
  *  source; RDO is what makes zstd bite). --genmipmap always: compressed mips
- *  can't be generated at runtime. */
-function ktx2EncodeArgs(encoder: string, isToktx: boolean, srgb: boolean, uastc: boolean, inPath: string, outPath: string): string[] {
+ *  can't be generated at runtime. `resize` (the GLB arm's texel budget,
+ *  store-variants.ts capTexels) rides toktx's own --resize — the encoder
+ *  scales before it generates mips, and libvips is nowhere in the loop. */
+function ktx2EncodeArgs(encoder: string, isToktx: boolean, srgb: boolean, uastc: boolean, inPath: string, outPath: string,
+  resize: [number, number] | null = null): string[] {
   return isToktx
     ? [encoder, "--t2", "--genmipmap", "--assign_oetf", srgb ? "srgb" : "linear",
+       ...(resize ? ["--resize", `${resize[0]}x${resize[1]}`] : []),
        ...(uastc ? ["--encode", "uastc", "--uastc_quality", "2", "--uastc_rdo_l", "1.0", "--zcmp", "18"]
                  : ["--encode", "etc1s", "--qlevel", "128"]),
        outPath, inPath]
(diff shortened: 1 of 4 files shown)


antra-tess (approved):
APPROVE exact head `ad979ec6a50d889bc75cf036279fa8a1ead5a623`.

The production root-cause correction is supported: the show-box markers are `not smaller`, not JPEG decoder failures. The KTX2 arm was encoding 2048² source maps while the ordinary shadow's budget is 1024².

Independent receipts:

- exact-head `store-variants-test.ts` on the production host with real `/usr/bin/toktx`: **0 failed**, including real 2048²→1024² KTX2 headers, stale-marker boot remeasure/current-marker retention, upload/serve/cache swap, retryable partial conversion, and split-brain canary;
- exact preregistered production artifact `assets/opt/store/0445768b0c87d590.glb`: **17,600,988 → 2,359,244 bytes** (7.5×, 23s), below the <6MB prediction;
- output SHA-256 `1d2394d72084b963a3d0d623f177f903fa1b2e4a7f891976919dd1f355418980`;
- all three embedded images declare `image/ktx2`, carry KTX2 magic, and report exactly **1024×1024** in their headers; `KHR_texture_basisu` is present;
- clean diff and clean worktree.

The recipe-stamped size verdict is the right complement: the sixteen old source-resolution refusals become questions once, while content verdicts and current-recipe size verdicts remain durable. The cap never upscales and the production canary proves the exact failing specimen, not only a synthetic analog.

Approval is code/review only; merge/deployment and the first boot's sixteen-item remeasurement remain separately observable operations.

PR #148, opened 2026-08-27
voice: push-to-talk — the held key is the gate

slimepriestess:
## What

A push-to-talk mode for voice chat: hold **V** to transmit, release to go silent. Off by default — voice activation stays exactly as tuned; PTT is the deliberate opt-in of someone whose room (fan, housemates, open-plan office) will not survive a level gate.

## Where it sits

**Rebased 2026-09-10 onto main after §24l R1 (ONE onset machine).** The level decision now lives inside `micgate.js`'s `makeOnsetGate`, so PTT is a **mode of that machine**, not a branch beside it: the factory takes a `held()` input (`null` = the level decides, boolean = the key decides) and gains `press()`, the declared-gesture 🎙 under the machine's own once-per-1.5s limit. Under PTT the level side keeps running (meter, noise floor) but its verdict drives nothing and its 🎙 is guarded — speech next to an un-pressed key must not open the gate or claim the room heard it. `info().speaking` reports the held answer when there is one.

The design below is unchanged; "gateAudio()" now means the machine's `apply`.

- **The key is the gate.** In PTT mode the level machinery keeps running (it still feeds the meter and the noise floor) but the gate decision is the held key alone — not key AND level: the person pressing has already answered the question the level gate exists to guess at. Discord and VRChat both transmit everything while held, for the same reason.
- **The existing envelope does the polish for free.** micgate's attack tau + 50ms lookahead shape the open, so a PTT press catches its own first consonant instead of clicking — something a hardware PTT switch cannot do.
- **Mute stays authoritative** over the key, unchanged, and synthetic sources still bypass the gate entirely — agents are untouched by construction.
- **The press is a gesture**: it fires the 🎙 announce directly (intent declared, not inferred), through the same `_lastOnset` rate-limit as the onset watcher, so mode flips cannot double-fire it.

## The honesty items (this repo's house religion, so listing them)

- The badge's **gold is gated on the key**: the analyser reads the raw side of the gate (it must — self-latching), so under PTT the level runs hot while nothing leaves the machine. Gold means "the room is getting this."
- Tooltips distinguish **armed** ("hold V to speak") from **transmitting** ("release V to stop").
- The **sensitivity row dims** under PTT — it drives nothing there, and a control that does nothing must not look like it does. Dimmed, not hidden: "where did my slider go" is a worse outcome than a visibly parked control.
- **keyup AND window blur drop the hold** — alt-tab with the key down is how every PTT implementation ships an open mic that says it is closed.
- **Leaving the mode drops a held key** (bus `audio:ptt`): the next regime starts from closed.

## Surfaces

- `voiceconsent.js`: `pttMode` pref (persisted in `eido.audio.prefs`), `audio:ptt` bus event.
- `micgate.js`: `makeOnsetGate({ …, held })` + `press()` — the machine owns both regimes and the one 🎙 rate-limit.
- `micstate.js`: `setPttHeld()`/`pttHeld()`, the `held` closure, mode-exit drop.
- `mictoggle.js`: in PTT mode V becomes hold-to-speak (VRChat's binding — same key both modes, one thing to learn); mic on/off moves to the glyph and the panel.
- `audiopanel.js`: `push-to-talk` checkbox between the mic switch and the sensitivity floor, because it decides which of the two regimes below it is real.

## Tests

- `tools/mic-ptt-ungated-test.mjs` — the ungated lane (gate graph unbuildable + `allowUngated`): PTT armed before the lane exists is applied synchronously in `gateFor`; arming PTT disables the raw tracks, held enables, released disables, mute outranks the key, mode exit restores the consented raw regime unless muted; **the transport's mic OFF outranks the key** — held→OFF→real ticks, armed→OFF→mode exit, a fresh press while OFF, and mic ON restoring each regime; (32 cases). Mutations: liveness dropped from the authority → 4 fail; voice activation not governed → 2; `setMicLive` without re-apply → 2; speaking readout without liveness → 1.
- `tools/sfu-mic-intent-race-test.mjs` — **executes the shipped `voicesfu.js`** against a fake `RTCPeerConnection` and a hand-resolved `voiceSource()`: two same-pc callers on one slow permission prompt, mic OFF before it resolves, device arrives → stopped, disabled, not published, not live, `wantMic` still false; the same with a synth provider (the OFF's synth swap happens); the mirror (ON during a synth acquisition → synth dropped, mic acquired); and the retained-device fast path reporting liveness (23 cases). Mutations: acquisition does not re-read intent → 4 fail; old waiter re-enable restored on top → 8 fail incl. liveness true (the round-3 head's signature); dropped source not replaced → 1; fast path without `setMicLive(true)` → 1.
- `tools/mic-ptt-test.mjs` — **executes** the gate under both regimes (29 cases, incl. held+muted → `speaking:false`, and held→mic OFF→closed/`speaking:false`→mic ON→reopens on the gated lane; wrapper removed → 1 fail, liveness dropped → 2 fail): VA ignores the key; press opens / release closes; mute outranks the key; mode exit drops the hold; announce fires once, rate-limited; **the level side under PTT drives nothing and fires no 🎙, and the same speech in VA opens and announces** (the analyser stub reports a settable room level, so the machine is driven, not poked); stuck-key (keyup/blur) and badge-honesty pins. Reverting either guard goes red: ignoring `held()` fails 4, dropping the 🎙 guard fails 1.
- `micstate-mesh-fallback-test`'s voiceconsent stub learns `pttMode`.
- Existing mic suites on this head: `micstate-exec` 10/10, `micstate-release` 14/14, `micstate-mesh-fallback` 5/5, `mic-gate-wired` 5/5, `tts-test` 37/37, `tts-mic-precedence` PASS, `voice-wiring` green, flight 231/231.
- `micstate-single-transport` 9/10 — the one failure ("emits audio:mic with that state") is identical on unmodified main `2c7e117`; not this change.

## Verified by ear (2026-09-10, `7039b1a`)

Two bodies on one machine — Chrome speaker tab with a real mic, Chrome incognito listener tab on headphones, local SFU loopback (`VOICE_TRANSPORT=sfu SFU_ICE_SERVERS='[]'`). Press opens, release closes, the listener hears exactly the held span, the badge goes gold only while the key is down, and voice activation is unchanged with the box unchecked. Round one found one defect (the focused-checkbox V guard, fixed above) and one setup trap worth knowing: a Safari listener stays silent because its autoplay policy blocks the programmatic speaker element — that is Safari, not the transport.

Still open: a pass on a real remote SFU lane (a peer on another machine) for press-latency feel under network delay. Not a gate for merge — the mechanism is the same lane.

### DIFF
--- a/client/lib/audiopanel.js
+++ b/client/lib/audiopanel.js
@@ -15,7 +15,7 @@
 import { makeSection } from './ui.js';
 import { audioPrefs, setVolume, receivingVoice, setReceiveVoice,
   sttConsented, setSttConsent, isHushed, setHush,
-  micFloor, setMicFloor } from './voiceconsent.js';
+  micFloor, setMicFloor, pttMode, setPttMode } from './voiceconsent.js';
 import { micAnalyserLevel as meshMicLevel } from './micstate.js';
 // 🔴 ASK THE LIVE TRANSPORT (2026-08-15, R found this one in the HUD: "the mic
 // icon isn't turning gold anymore when it is transmitting"). voice.js's
@@ -334,7 +334,29 @@ function paint(body) {
     });
   _sync.mic = micRow.querySelector('input');
   body_.append(micRow);
-  body_.append(micFloorRow());   // sensitivity belongs UNDER the switch it serves
+  // PTT sits between the mic switch and the sensitivity floor because it
+  // DECIDES which of the two regimes below it is real: checked, the gate is
+  // the held key and the floor drives nothing; unchecked, voice activation
+  // owns the gate and V goes back to being the mic toggle.
+  body_.append(checkRow('push-to-talk',
+    'transmit only while V is held down — replaces voice activation, so the ' +
+    'sensitivity floor below is not used while this is on. Release V and the ' +
+    'room hears nothing.',
+    pttMode(), (on) => setPttMode(on)));
+  const floorRow = micFloorRow();
+  body_.append(floorRow);        // sensitivity belongs UNDER the switch it serves
+  // Under PTT the floor is a control that does nothing — the exact thing this
+  // panel's history says must never LOOK functional. Dimmed and inert, not
+  // hidden: hiding it would make "where did my sensitivity slider go" a
+  // support question, and the dim states the relationship (this row belongs
+  // to the mode you are not in).
+  const dimFloor = () => {
+    const off = pttMode();
+    floorRow.style.opacity = off ? '0.45' : '';
+    floorRow.style.pointerEvents = off ? 'none' : '';
+  };
+  dimFloor();
+  bus.on('audio:ptt', dimFloor);
 
   const hearRow = checkRow('hear voices',
     'peers and agent speech — the 🎧 glyph is this same switch',
--- a/client/lib/micgate.js
+++ b/client/lib/micgate.js
@@ -97,6 +97,7 @@ export function gateStream(stream, levelFn) {
     _ctx = audioContext();
     if (!_ctx || typeof _ctx.createMediaStreamDestination !== 'function') {
       _gateUnavailable = true;
+      _rawWanted = _ungatedConsent ? 1 : 0;      // a consented raw lane is open until PTT says otherwise
       return _ungatedConsent ? stream : null;   // fail CLOSED, not raw
     }
     _gateUnavailable = false;
@@ -168,6 +169,26 @@ export function driveGate(open) {
   _wanted = target;
 }
 
+// ── THE UNGATED LANE'S ONLY GATE ─────────────────────────────────────────────
+// When the graph could not be built and the person allowed raw transmission
+// (allowUngated), the wire carries the device stream ITSELF: there is no gain
+// for driveGate() to drive, and every decision above lands on nothing. That was
+// acceptable while the only regime was voice activation — "transmit UNGATED"
+// is exactly what the consent row said. Push-to-talk makes a new promise on the
+// same lane ("release V and the room hears nothing") that a nonexistent gain
+// cannot keep (Mica, #148 review: raw audio left while the key was up). So on
+// that lane PTT drives the raw tracks' `enabled` flag instead: binary and
+// abrupt — the very reason the graph exists, see the top of this file — but
+// TRUE, and the same mechanism mute already uses. Intent is recorded so
+// gateOpenness() tells the badge what the wire is doing.
+let _rawWanted = 0;
+export function driveRawTracks(open) {
+  if (!_rawStream) return false;
+  for (const t of _rawStream.getAudioTracks?.() ?? []) t.enabled = !!open;
+  _rawWanted = open ? 1 : 0;
+  return true;
+}
+
 
 // ── MONITOR: hear your own lane, exactly as the room hears it ───────────────
 // R, 2026-08-09: "can you feed my own audio lane back to me for this test so I
@@ -234,7 +255,7 @@ export const monitoring = () => !!_mon;
 // Reports the gate's INTENT, not the instantaneous .value: during a fade the
 // real gain is mid-slope, and an indicator that flickers through every envelope
 // would misreport 'am I being heard' at exactly the moments that matter.
-export const gateOpenness = () => (_gain ? _wanted : 0);
+export const gateOpenness = () => (_gain ? _wanted : (_gateUnavailable && _rawStream ? _rawWanted : 0));
 export const gateGainNow = () => (_gain ? _gain.gain.value : 0);
 export const isGated = () => !!_gain;
 
@@ -304,6 +325,7 @@ export function release() {
   _src = _gain = _dest = null;
   _gatedStream = null;
   _rawStream = null;
+  _rawWanted = 0;
 }
 
 /** The RAW stream, for anything that must measure the true input — the analyser
--- a/client/lib/micgate.js
+++ b/client/lib/micgate.js
@@ -380,14 +380,24 @@ export function makeLevelMeter(getMeasured) {
  *  - `speaking` reports threshold PLUS hang-time, or the bar flickers dark
  *    through every pause while the room still hears you;
  *  - 20ms tick: the interval is the worst-case clip on a word's attack. */
-export function makeOnsetGate({ level, threshold, drive, announce }) {
+export function makeOnsetGate({ level, threshold, drive, announce, held = () => null }) {
   let _timer = null, _above = false, _lastOnset = 0, _openUntil = 0;
   let _openedAt = 0, _announced = false;
   let _noise = 0.01, _settle = 0;
   let _blocks = new Array(6).fill(Infinity), _blockIdx = 0, _blocksFilled = 0,
       _blockMin = Infinity, _blockStart = 0;
 
-  const apply = (now) => drive(_above || now <= _openUntil);
+  // PUSH-TO-TALK IS A MODE OF THIS MACHINE, not a bypass around it. `held()`
+  // returns null when the level decides (the default), or the key's boolean
+  // when a held key decides. The level side keeps running either way — it
+  // still feeds the meter and the noise floor, and the envelope + lookahead
+  // downstream still shape the open — but under a held answer the level's
+  // verdict drives nothing and its 🎙 must not fire: a person talking next to
+  // an un-pressed key is exactly the person who chose PTT so the room would
+  // NOT hear that. The press announces instead (press(), below), through
+  // the same rate-limit, so the pill and the audio still agree.
+  const decide = (now) => { const h = held(); return h === null ? (_above || now <= _openUntil) : h; };
+  const apply = (now) => drive(decide(now));
 
   function tick() {
     const lv = level();
@@ -420,7 +430,7 @@ export function makeOnsetGate({ level, threshold, drive, announce }) {
     if (lv >= on) {
       _openUntil = now + 700;                // hang-time
       if (!_above) { _above = true; _openedAt = now; _announced = false; }
-      if (!_announced && now - _openedAt >= 60 && now - _lastOnset > 1500) {
+      if (held() === null && !_announced && now - _openedAt >= 60 && now - _lastOnset > 1500) {
         _announced = true; _lastOnset = now; announce();
       }
     } else if (_above && lv < off && now > _openUntil) _above = false;
@@ -448,7 +458,14 @@ export function makeOnsetGate({ level, threshold, drive, announce }) {
     /** Drop the speaking latch without stopping the watch — unmute hands
      *  control back to the gate rather than reopening mid-latch. */
     dropLatch() { _above = false; _openUntil = 0; },
+    /** A push-to-talk press: announce the 🎙 as a declared gesture, under the
+     *  same once-per-1.5s limit as a level onset. The caller applies the gate. */
+    press(now = Date.now()) {
+      if (now - _lastOnset > 1500) { _lastOnset = now; announce(); }
+    },
+    /** `speaking` is what the room is getting: the held answer when there is
+     *  one, else threshold PLUS hang-time. */
     info: () => ({ level: level(), noise: _noise, on: threshold(),
-      speaking: _above || Date.now() <= _openUntil }),
+      speaking: decide(Date.now()) }),
   };
 }
--- a/client/lib/micstate.js
+++ b/client/lib/micstate.js
@@ -29,7 +29,7 @@
 import { bus } from './base.js';
 const flashHint = (msg) => import('./ui.js').then((u) => u.flashHint(msg)).catch(() => {});
 import { sendTyping } from './net.js';
-import { gateThreshold } from './voiceconsent.js';
+import { gateThreshold, pttMode } from './voiceconsent.js';
 import { gateStream, attachSource, detachSource, driveGate, setMonitor, monitoring,
          gateUnavailable, ungatedConsent, isGated,
          makeOnsetGate, makeLevelMeter } from './micgate.js';
@@ -56,6 +56,35 @@ export function toggleMute(on) {
   return _muted;
 }
 
+// ── push-to-talk: the key IS the gate ───────────────────────────────────────
+// In PTT mode the level machinery keeps running — it still feeds the meter and
+// the noise floor — but the gate DECISION is the held key alone. Not key AND
+// level: the person pressing the key has already answered "do I mean to be
+// heard", which is the only question the level gate ever existed to guess at.
+// (Discord and VRChat both transmit everything while held, for the same
+// reason.) The envelope and lookahead in micgate.js still shape the open, so
+// a PTT press catches its first consonant instead of clicking — something a
+// hardware PTT switch cannot do.
+//
+// Mute stays authoritative above this, unchanged: a muted mic ignores the key
+// exactly as it ignores speech.
+let _pttHeld = false;
+export const pttHeld = () => _pttHeld;
+export function setPttHeld(on) {
+  const next = !!on;
+  if (next === _pttHeld) return;
+  _pttHeld = next;
+  const now = Date.now();
+  // The press is a GESTURE — intent declared, not inferred. Announce the 🎙
+  // on the press itself rather than waiting for the level to clear a floor
+  // the mode no longer uses. It goes through the machine's own once-per-1.5s
+  // rate-limit (the same one the level onset uses), so a mode flip or a
+  // rapid re-press cannot double-fire it.
+  if (next && pttMode() && micOn()) _onset.press(now);
+  gateAudio(now);   // apply immediately — a 20ms tick is an audible latency here
+}
+bus.on('audio:ptt', (on) => { if (!on && _pttHeld) { _pttHeld = false; gateAudio(Date.now()); } });
+
 // 🔴 THE ONSET WATCHER'S STATE. Extracted from voice.js:680-682 — the slice
 // that moved the FUNCTIONS started below these declarations, so every one of
 // them was a free variable here: startOnsetWatch/onsetTick/gateAudio threw
@@ -75,6 +104,9 @@ const _onset = makeOnsetGate({
   threshold: gateThreshold,
   drive: (open) => driveGate((!_lane || _muted) ? false : open),
   announce: () => sendTyping(null, 'mic'),
+  // Push-to-talk hands the machine a HELD answer instead of a level to judge;
+  // null means "no override, decide by level" — see setPttHeld above.
+  held: () => (pttMode() ? _pttHeld : null),
 });
 const gateAudio = (now) => _onset.apply(now);
 const startOnsetWatch = () => _onset.start();
--- a/client/lib/micstate.js
+++ b/client/lib/micstate.js
@@ -173,6 +173,14 @@ export function gateFor(rawStream) {
   _lane = _gated;
   _deviceLive = true;
   startOnsetWatch();
+  // Apply the CURRENT authority before the lane leaves this function. The
+  // watcher's first tick is 20 ms away, and on an ungated lane the raw track
+  // arrives enabled: with push-to-talk already armed (a persisted pref, or
+  // a mode set before the device was acquired), that tick was the only thing
+  // closing the wire — a frame could leave while the UI promised silence and
+  // `speaking` said false over an open track (Mica, #148 re-review). Mute
+  // and the held-key decision are both applied here, synchronously.
+  gateAudio(Date.now());
   return _gated;
 }
 
--- a/client/lib/micstate.js
+++ b/client/lib/micstate.js
@@ -30,7 +30,7 @@ import { bus } from './base.js';
 const flashHint = (msg) => import('./ui.js').then((u) => u.flashHint(msg)).catch(() => {});
 import { sendTyping } from './net.js';
 import { gateThreshold, pttMode } from './voiceconsent.js';
-import { gateStream, attachSource, detachSource, driveGate, setMonitor, monitoring,
+import { gateStream, attachSource, detachSource, driveGate, driveRawTracks, setMonitor, monitoring,
          gateUnavailable, ungatedConsent, isGated,
          makeOnsetGate, makeLevelMeter } from './micgate.js';
 // 🔴 The analyser hangs off the shared context; omitting this import made
@@ -83,7 +83,23 @@ export function setPttHeld(on) {
   if (next && pttMode() && micOn()) _onset.press(now);
   gateAudio(now);   // apply immediately — a 20ms tick is an audible latency here
 }
-bus.on('audio:ptt', (on) => { if (!on && _pttHeld) { _pttHeld = false; gateAudio(Date.now()); } });
+bus.on('audio:ptt', (on) => {
+  const now = Date.now();
+  if (on) {
+    // Arming closes the gate NOW, not on the next 20ms tick: on an ungated
+    // lane that tick is a window of raw transmission the mode just promised
+    // away (see the drive closure below).
+    gateAudio(now);
+    return;
+  }
+  // Leaving PTT with the key down must not leave a phantom finger on the gate:
+  // the next regime starts from closed and earns its own open.
+  if (_pttHeld) _pttHeld = false;
+  gateAudio(now);
+  // An ungated lane goes back to what its consent row said — raw and open —
+  // unless the mic is muted, which outranks every regime.
+  if (gateUnavailable()) driveRawTracks(!!_lane && !_muted);
+});
 
 // 🔴 THE ONSET WATCHER'S STATE. Extracted from voice.js:680-682 — the slice
 // that moved the FUNCTIONS started below these declarations, so every one of
@@ -102,7 +118,15 @@ const _meter = makeLevelMeter(() => (!_lane || _muted) ? null : (_raw || _lane))
 const _onset = makeOnsetGate({
   level: _meter,
   threshold: gateThreshold,
-  drive: (open) => driveGate((!_lane || _muted) ? false : open),
+  drive: (open) => {
+    const want = (!_lane || _muted) ? false : open;
+    driveGate(want);
+    // No graph to drive (gate unavailable, raw transmission consented): under
+    // push-to-talk the key still has to close the WIRE, so the decision goes
+    // to the raw tracks themselves — micgate.js driveRawTracks. Voice
+    // activation on that lane stays raw; that is what the consent row said.
+    if (pttMode() && gateUnavailable()) driveRawTracks(want);
+  },
   announce: () => sendTyping(null, 'mic'),
   // Push-to-talk hands the machine a HELD answer instead of a level to judge;
   // null means "no override, decide by level" — see setPttHeld above.
@@ -115,7 +139,15 @@ const stopOnsetWatch = () => _onset.stop();
 /** What the gate is actually doing right now — for the meter and for
  *  tuning. Post-cutover (anima merge, §24n): the mesh is GONE — this
  *  module's own factory instance is the only gate there is. */
-export const micGateInfo = () => _onset.info();
+// `speaking` is what the room is getting. The machine's own answer is the
+// gate decision (level, or the held key); the lane's authority — no lane, or
+// muted — is applied in the drive closure and must be applied HERE too, or a
+// held key while muted reports speaking:true over a gate that is at 0 (Mica,
+// #148 review). The pill and the audio must agree in every regime.
+export const micGateInfo = () => {
+  const info = _onset.info();
+  return (!_lane || _muted) ? { ...info, speaking: false } : info;
+};
 
 /** Live mic level 0..1 for UI — the factory meter over this module's own
  *  _raw/_lane (the mesh delegation retired with voice.js). */
--- a/client/lib/mictoggle.js
+++ b/client/lib/mictoggle.js
@@ -238,8 +238,23 @@ bus.on('audio:mic', paint);
 setInterval(ensure, 1000);
 ensure();
 
+// 🔴 "TYPING" MEANS A TEXT FIELD, NOT ANY INPUT. The old guard skipped V
+// whenever an INPUT had focus — which includes the checkbox you just
+// clicked to turn push-to-talk on. Focus stays on a checkbox after the
+// click, so the very first V after enabling PTT was silently eaten: the
+// mode looked broken while the gate was fine (R, 2026-09-10, found by ear;
+// the two-tab probe reproduced it with a focused checkbox → held=false).
+// Only a place where V would insert a character is a reason to ignore it.
+const TEXT_INPUT = /^(text|search|url|email|password|number|tel|)$/i;   // '' = default type
+const typingTarget = () => {
+  const el = document.activeElement;
+  if (!el) return false;
+  if (el.isContentEditable) return true;
+  if (el.tagName === 'TEXTAREA') return true;
+  return el.tagName === 'INPUT' && TEXT_INPUT.test(el.type ?? '');
+};
 window.addEventListener('keydown', (e) => {
-  if (e.repeat || ['INPUT', 'TEXTAREA'].includes(document.activeElement?.tagName)) return;
+  if (e.repeat || typingTarget()) return;
   if (e.code !== 'KeyV' || e.shiftKey) return;
   // V speaks. Hushing is a CLICK on the 🎧, and revoking consent entirely
   // lives only in the Audio panel — deliberately not on a key next to the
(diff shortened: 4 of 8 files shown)


slimepriestess:
Rebased onto main at `2c7e117` (155 commits, incl. §24l R1's ONE onset machine). New head: **`94aed9a`**.

The original hook point — the level branch in `micstate.gateAudio()` — no longer exists; the decision lives in `makeOnsetGate`'s `apply`, and the 🎙 rate-limit it shared is private to the factory. So PTT moved *into* the machine rather than around it: `makeOnsetGate({ …, held })` plus `press()`. That also closed a hole the old branch had: under PTT the level side kept ticking and would still have fired the 🎙 on speech with no key down (a pill claiming the room heard what it didn't). Guarded now, and the test drives the level to prove it — dropping the guard goes red.

Neighbouring mic gates all match main on this head; `micstate-single-transport` has one pre-existing failure on main itself (details in the body). By-ear pass in the field is still the acceptance step.

slimepriestess:
By-ear pass round 1 (two tabs on one machine, SFU loopback) found one real defect, fixed at `7039b1a`:

**The first V after enabling PTT did nothing.** The keydown guard treated *any* focused INPUT as typing, and the push-to-talk checkbox keeps focus after the click — so V was eaten until you clicked the world. `typingTarget()` now means contentEditable / TEXTAREA / a text-like INPUT. A headless two-tab probe reproduced the defect (focused checkbox → `held=false`) and passes after (`held=true`, gate open, listener tab measures signal only while the key is down: idle 0.000 rms, held 0.036, released 0.000 — same signal as voice-activation). Test pins the guard shape, 22/22.

Two environment notes for anyone else running this locally, neither a PR change: a world launched without `VOICE_TRANSPORT=sfu` is a silent no-voice world since the mesh cutover (every voice control present, none wired), and a same-name second tab takes over the first body rather than adding a listener.

slimepriestess:
By-ear pass done at `7039b1a`: two Chrome bodies (incognito for the second), local SFU loopback. Press opens, release closes, listener hears exactly the held span, badge gold only while down, voice activation unchanged with the box off. Ready for review.

antra-tess (changes requested):
# Eidoverse Worlds PR #148 — push-to-talk review

**Date:** 2026-09-11  
**PR:** https://github.com/anima-research/eidoverse-worlds/pull/148  
**Base:** `2c7e117def11ad27ad0ea3b26658d77435e92219`  
**Head:** `7039b1a147f941e922e24a6b7f9d9c8a8f27824e`  
**Disposition:** **CHANGES_REQUESTED**

## What is sound

The core direction is correct: push-to-talk is a mode of the existing `makeOnsetGate`, not a parallel transport/gate implementation. The level meter/noise-floor machinery remains alive, but `held()` supplies the transmission decision while PTT is enabled. Mute still drives the transport gain closed, synthetic sources remain outside the room-noise gate, and voice activation is unchanged by default.

- PTT gate suite: `22/22`
- micstate execution: `10/10`
- device release/lane retention: `14/14`
- post-cutover transport selection: `5/5`
- gate wiring: `5/5`
- resume-after-synth: PASS
- TTS/mic precedence: PASS
- flight typecheck: PASS
- server bundle: PASS
- both MCP entrypoint bundles: PASS after supplying the base/main MCP dependency closure
- `git diff --check`: clean

`micstate-single-transport-test` remains `9/10`; its `audio:mic` assertion fails byte-identically on exact base, so it is inherited rather than introduced here.

The contributor’s two central mutations were independently reproduced:

```text
ignore held() decision       → 18 pass / 4 fail
announce from level in PTT   → 21 pass / 1 fail
```

A third mutation removing the mode-exit held-key drop produced `18/22`. The changed core is therefore meaningfully bound by the focused suite.

The implementation also correctly distinguishes typing-capable inputs from checkbox/range/button focus, drops the hold on keyup and blur, dims/disables the sensitivity row under PTT, rate-limits the declared 🎙 press gesture through the shared onset machine, and makes the HUD tooltip distinguish armed from transmitting.

## Blocker 1 — explicit ungated fallback defeats PTT while the UI says it is closed

The existing audio panel deliberately offers an explicit escape hatch when WebAudio gate construction fails:

```text
transmit UNGATED (voice gate unavailable)
```

With that consent enabled, `gateStream()` returns the raw device stream. The SFU publishes it because `gateIsUnavailable() && !ungatedAllowed()` is false. PTT only drives `driveGate()`, but no gain node exists on this path, so press/release cannot affect the transmitted raw lane.

Independent real-module probe at exact head, with a gate-unavailable AudioContext and `allowUngated(true)`:

```json
{
  "sameRaw": true,
  "gateUnavailable": true,
  "ungatedAllowed": true,
  "held": false,
  "gateIntent": 0,
  "micOn": true
}
```

Thus the room can receive the raw microphone continuously while the PTT state says the key is not held and the UI promises:

> Release V and the room hears nothing.

The user did consent to raw transmission under the older voice-activation regime, but enabling PTT creates a new and contradictory guarantee. This is a privacy-facing blocker.

Required behavior should be explicit and tested. Examples include refusing/turning off PTT while the lane is ungated, revoking ungated publication on PTT enable, or refusing microphone publication until a controllable gate exists. Do not silently let raw transmission continue beneath a closed PTT indicator.

Required product-path discriminator:

```text
gate unavailable + explicit ungated consent + PTT enabled + key released
→ no outbound raw microphone, OR a loud refusal that PTT cannot be armed
```

Mutating away the repair must turn it red.

## Blocker 2 — `micGateInfo().speaking` ignores mute in PTT mode

The PR documents `info().speaking` as “what the room is getting”. But `makeOnsetGate.info()` returns `decide(Date.now())`, and under PTT `decide()` returns only the held key. Transport authority/mute is applied separately in the `drive` closure.

Independent exact-head probe:

```text
held + unmuted:
  micGateInfo().speaking = true
  gateOpenness() = 1

held + MUTED:
  micGateInfo().speaking = true
  gateOpenness() = 0
  measured level = 0
```

The audio gain is correctly closed, but the public state API says speaking. The existing focused test checks that mute closes `gateOpenness()`, but does not assert `micGateInfo().speaking === false` under the same condition.

Required repair: make the claimed “what the room gets” state include transport/mute/device authority, or narrow/rename the field so it no longer claims that. Add the direct held+muted assertion and a mutation that turns it red.

## Boundaries

- The implementation is off by default and leaves voice activation unchanged.
- The two-body local SFU by-ear test is useful field evidence. A remote-peer press-latency pass remains worthwhile but is not the blocker here.
- No production voice configuration, consent state, or world state was changed during review.
- Review grants no merge, deployment, voice activation, or field-test authority.

## Verdict

The one-machine PTT architecture and primary focused evidence are strong. Two safety/truthfulness seams remain: explicit ungated fallback can transmit continuously beneath a released PTT key, and the public speaking-state API reports true while mute has actually closed the lane. **CHANGES_REQUESTED** until both are repaired with executable product-path discriminators.

slimepriestess:
Both blockers taken as stated, fixed at **`2630478`** (`26304781689ccf332d917df694a7453611e2e38e`).

**1. Ungated lane.** Reproduced your probe first (`sameRaw:true, gateUnavailable:true, ungatedAllowed:true, held:false, micOn:true`, track still transmitting). Repair: on that lane PTT drives the raw tracks' `enabled` flag from the same decision the gain would have taken — `driveRawTracks` in micgate.js, called from micstate's drive closure only when `pttMode() && gateUnavailable()`. Arming applies immediately (no 20ms raw window), leaving PTT restores the consented raw regime unless muted, and `gateOpenness()` now reports the raw lane's intent. Voice activation on that lane is untouched: raw, as the consent row says. I chose "keep the promise" over "refuse to arm" because the promise is keepable there; binary gating is what the graph exists to improve on, not a wrong answer.

**2. `speaking`.** The lane's mute authority was applied in the drive closure but not in the report. `micGateInfo()` now returns `speaking:false` whenever there is no lane or the mic is muted, in every regime.

Discriminators, each red on repair removal: `tools/mic-ptt-ungated-test.mjs` (12 cases; remove the raw-track drive → 4 fail, remove the mode-exit restore → 1 fail) and two new cases in `mic-ptt-test` (24/24; remove the wrapper → 1 fail in each file). Neighbours unchanged; single-transport still 9/10 from base.

antra-tess (changes requested):
# Eidoverse PR #148 — repaired-head rereview

**Date:** 2026-09-11  
**PR:** https://github.com/anima-research/eidoverse-worlds/pull/148  
**Base:** `2c7e117def11ad27ad0ea3b26658d77435e92219`  
**Prior reviewed head:** `7039b1a147f941e922e24a6b7f9d9c8a8f27824e`  
**Reviewed head:** `26304781689ccf332d917df694a7453611e2e38e`  
**Disposition:** **CHANGES_REQUESTED**

## Repairs that pass

The new delta correctly repairs both exact first-round specimens once a raw lane already exists:

- On a gate-unavailable but explicitly consented raw lane, enabling PTT immediately disables every raw audio track; key press enables, release disables, mute outranks hold, and leaving PTT restores the prior consented raw regime unless muted.
- `gateOpenness()` reports raw-lane intent rather than hard zero.
- `micGateInfo().speaking` now includes no-lane/mute authority, so held+muted says `false` while held+unmuted says `true`.

```text
mic-ptt                         24/24
mic-ptt-ungated                12/12
micstate exec                  10/10
micstate release               14/14
transport selection             5/5
gate wiring                     5/5
resume after synth              PASS
TTS/mic precedence              PASS
voice wiring                     8/8
flight typecheck                PASS
```

`micstate-single-transport` remains `9/10` on both exact base and head at its inherited `audio:mic` emission assertion.

Mutations are load-bearing:

```text
remove raw-track drive          4 failures
remove mode-exit restoration    1 failure
remove speaking-state wrapper   1 failure
```

The PR merges conflict-free onto current main `ac76122aec2d4df970a82f085e0a667e30b735bf`; the focused composed-main suite stays green.

## Blocker — a persisted/pre-armed PTT preference leaks the newly acquired ungated raw lane until the first timer tick

The repaired test enables PTT **after** `gateFor(stream)` has already installed the raw lane. The real product can restore `pttMode:true` from `localStorage` before a microphone device is acquired.

Exact product sequence:

```text
allowUngated(true)
setPttMode(true)      // or persisted true before acquisition
held = false
gateFor(rawStream)
```

`gateStream()` sets `_rawStream`, marks gate unavailable, sets `_rawWanted=1`, and returns the raw stream with its track still enabled. `gateFor()` assigns `_lane` and starts the 20 ms onset interval, but does not apply the already-active PTT decision synchronously. The first interval tick will eventually call the drive closure and disable the track; before then the exact privacy promise is false.

Independent real-module probe at exact head:

```json
{
  "sameRaw": true,
  "ptt": true,
  "held": false,
  "trackEnabledImmediately": true,
  "openness": 1,
  "speaking": false
}
```

The SFU receives the raw track from `gateFor()` while it is enabled. This is not only an indicator mismatch: one audio frame may leave before the 20 ms watcher runs. The surrounding code explicitly treats 20 ms as the worst-case clip on a word's attack and says arming PTT closes NOW, so the gap is inside the feature's own acceptance boundary.

It also recreates the second first-round defect during that interval: `micGateInfo().speaking` says `false` while `gateOpenness()` is 1 and the raw track is enabled.

Required repair:

- on every fresh/reacquired raw lane, synchronously apply the current PTT/mute decision before returning the stream to the transport;
- preserve voice-activation behavior on explicitly consented ungated lanes when PTT is off;
- add a product-path regression for `pttMode=true` before `gateFor(raw)` asserting immediately, before timers:
  - raw track disabled,
  - `gateOpenness()===0`,
  - `micGateInfo().speaking===false`;
- then press/release and leave-mode to prove the existing transitions still work;
- a mutation removing the acquisition-time application must turn the new test red.

## Non-blocking lifecycle note

`audiopanel.paint()` adds an `audio:ptt` listener capturing `floorRow` without retaining the returned unsubscribe. The body is normally built once and repaint is idempotent, so I did not establish a live leak. If sections can be replaced/reinitialized, add teardown ownership rather than accumulating listeners to detached rows.

## Verdict

The repaired head closes the demonstrated steady-state ungated-PTT and speaking-state failures, with good executable and mutation evidence. It does not yet keep the same guarantee when PTT is already selected before the raw lane exists. **CHANGES_REQUESTED** on exact head `26304781689ccf332d917df694a7453611e2e38e`.

slimepriestess:
Acquisition-order seam taken as stated; fixed at **`6b79289`** (`6b79289b523bbcaf33b8558684092e336373faa4`).

Reproduced first (PTT armed, then `gateFor(raw)` on a gate-unavailable context with raw consent: track enabled, openness 1, speaking false, synchronously after the call). `gateFor()` now applies the current authority — mute, then the held-key decision through the same drive closure that drives the raw tracks on an ungated lane — before the lane is returned, with no timer in the path. The watcher still starts; it is no longer what closes the wire first.

Discriminator: a new no-timer leg at the top of `tools/mic-ptt-ungated-test.mjs` — PTT armed first, then `gateFor`: track disabled, openness 0, speaking false, all asserted synchronously with the call; press opens, release closes. Removing the synchronous apply turns 4 cases red. 18/18 there, 24/24 in `mic-ptt-test`, neighbours unchanged.

slimepriestess:
Correction to the comment above: removing the synchronous apply turns **2** cases red (the synchronous track-disabled and openness-0 assertions), not 4 — the `speaking` and press/release assertions in that leg stay green because they read the held decision, which was already correct. The commit message carries the same wrong count; the receipts are the tests.

antra-tess (changes requested):
# Eidoverse PR #148 — acquisition-order rereview

**Date:** 2026-09-11  
**PR:** https://github.com/anima-research/eidoverse-worlds/pull/148  
**Prior reviewed head:** `26304781689ccf332d917df694a7453611e2e38e`  
**Reviewed head:** `6b79289b523bbcaf33b8558684092e336373faa4`  
**Disposition:** **CHANGES_REQUESTED**

## Acquisition-order repair accepted

The one-commit delta correctly closes the previous blocker. `gateFor()` now applies current mute/PTT authority synchronously after the lane exists and before returning it to the transport. With PTT persisted/armed before an explicitly consented ungated raw lane is acquired, exact-head behavior is immediately:

```text
track.enabled = false
gateOpenness() = 0
micGateInfo().speaking = false
```

No timer is needed. Press opens and release closes. The expanded test passes `18/18`; deleting the acquisition-time application makes exactly the two synchronous safety assertions fail (`16/18`), matching Weft's correction to the earlier PR-body count.

Independent exact-head neighbors:

```text
mic-ptt                         24/24
mic-ptt-ungated                18/18
micstate exec                  10/10
micstate release               14/14
transport selection             5/5
gate wiring                     5/5
resume after synth              PASS
TTS/mic precedence              PASS
voice wiring                     8/8
flight typecheck                PASS
```

`micstate-single-transport` remains the inherited `9/10` on both exact base and earlier repaired head. The branch merges conflict-free onto current main `ac76122aec2d4df970a82f085e0a667e30b735bf`.

## Blocker — raw-track PTT writers ignore transport/device-off authority

The repaired ungated lane uses the raw device track itself as the transport lane. `voicesfu.js::sfuMic(false)` (when no synth provider is available) authoritatively disables that track and calls `setMicLive(false)`. But the new raw-track writers derive authority from only `_lane && !_muted`:

```js
// micstate.js drive closure
const want = (!_lane || _muted) ? false : open;
if (pttMode() && gateUnavailable()) driveRawTracks(want);

// leaving PTT
if (gateUnavailable()) driveRawTracks(!!_lane && !_muted);
```

Neither path consults `_deviceLive` / `micOn()`. On an ungated lane `_lane` remains present after mic OFF because the sender lane is retained. Therefore a lower-authority PTT transition can re-enable a device track that the microphone control turned off.

Independent exact-head real-module probe:

```text
allowUngated(true)
PTT armed, key up
gateFor(raw)
setMicLive(true)
transport mic OFF: track.enabled=false; setMicLive(false)
leave PTT
```

Result:

```json
{
  "armed":{"enabled":false,"openness":0,"micOn":true},
  "off":{"enabled":false,"micOn":false,"speaking":false},
  "exit":{"enabled":true,"micOn":false,"openness":1,"speaking":false}
}
```

The raw track is transmitting again while the microphone says OFF and the speaking indicator says false. This is unbounded: voice-activation mode does not subsequently drive raw tracks closed. The sibling path is held V → mic OFF; the next 20 ms onset tick can likewise re-enable the raw track until release.

This blocker was also independently found by a hostile Opus reviewer on the preceding head. The acquisition-order commit does not change the two raw-track authority expressions, so it remains at `6b79289`.

Required repair:

- raw-track open authority must include current device/transport liveness, not only lane existence and mute;
- neither an onset tick nor mode exit may re-enable a track after authoritative mic OFF;
- add product-path tests for:
  1. PTT held → mic OFF → await multiple ticks → track remains disabled, `micOn=false`, `speaking=false`;
  2. PTT armed/key up → mic OFF → leave PTT → track remains disabled and openness stays 0;
  3. subsequent explicit mic ON restores the correct selected regime;
- mutations removing device-live authority from either writer must turn red.

## Non-blocking note

`audiopanel.paint()` adds an `audio:ptt` listener capturing `floorRow` without retaining the unsubscribe. The section body is normally stable and paint is idempotent, so no live leak was established. If panels can be replaced/reinitialized, teardown ownership should be added.

## Verdict

The acquisition-order seam is repaired with an honest no-timer test and mutation. A stronger same-class defect remains: PTT's raw-track writer can override authoritative microphone OFF. **CHANGES_REQUESTED** on exact head `6b79289b523bbcaf33b8558684092e336373faa4`.

slimepriestess:
## Round 3 — the mic control outranks the key. Head `e2644d5111f8cfa0bb7071d43f4eee88d6069512`

Confirmed the defect exactly as described, and it was one layer deeper than the two writers you named: `micGateInfo()` used the same `_lane && !_muted` authority, so held → mic OFF reported `speaking:false` only because the *key* was up in your probe — with the key held it would have said true over a track the transport had disabled.

### What changed

- **One authority.** The gate's decision is now `micOn() ? open : false` — lane AND device AND not muted, the same three-state answer everyone else already asks. Nothing in the gate path reads `_lane && !_muted` any more; `micGateInfo()` uses `micOn()` too.
- **One raw-track writer.** The mode-exit `driveRawTracks(!!_lane && !_muted)` is gone; the drive closure is the only writer outside the transport, and it governs voice activation by liveness as well: `driveRawTracks(pttMode() ? want : micOn())`. So a tick in either regime writes the same answer the transport wrote, and nothing can re-open what the microphone control closed. (This also makes the VA regime self-healing where your review noted it was unbounded.)
- **`setMicLive()` re-applies on change.** Liveness is authority now, so a report re-runs the decision at once rather than on the next 20 ms tick. That is what restores the regime on mic ON: the transport re-enables the raw track, and PTT with the key up closes it again in the same synchronous call.
- **Transport fast path.** `sfuMic(true)`'s two retained-device paths enabled the tracks and returned without `setMicLive(true)` — the only one lived in `sfuPublish`. That was already a gap (`micOn()` stayed false after OFF→ON through that path), and with liveness in the authority it would have left the gate shut for good. Both sites now report in the same synchronous breath, via the static import that was already there.

### Tests (all product-path except the last)

`mic-ptt-ungated-test` 35/35 (was 18), the sequences from your required repair verbatim, with the transport's OFF/ON reproduced in its order (disable then `setMicLive(false)`; enable then `setMicLive(true)`) and the onset watcher running on its real 20 ms interval:

1. PTT held → mic OFF → 90 ms of ticks → track disabled, `micOn=false`, `speaking=false`, openness 0; a fresh press while OFF does not reopen.
2. PTT armed, key up → mic OFF → leave PTT → track disabled, openness stays 0, and stays so across ticks in VA.
3. Explicit mic ON restores the selected regime: under PTT the track is closed again synchronously and a press opens; under VA the consented raw lane is open with openness 1.

`mic-ptt-test` 29/29 (was 24): the gated lane under held → OFF (gain closes, `speaking:false`), fresh press while OFF stays closed, mic ON while held reopens.

The transport fast path has no executing harness (`sfuMic` needs a live pc), so it is pinned source-level in the ungated test with a negative control on the pre-fix shape — same standing as the mictoggle pins.

Mutations, each restored before the next:

| mutation | ungated | gated |
|---|---|---|
| authority back to `(!_lane \|\| _muted)` | 4 red | 2 red |
| raw drive only under PTT (VA ungoverned) | 2 red | — |
| `setMicLive` stores without re-applying | 2 red | 2 red |
| fast path enables without `setMicLive(true)` | 2 red | — |
| `micGateInfo` without liveness | 1 red | 1 red |

Neighbors unchanged: micstate-exec 10/10, release 14/14, mesh-fallback 5/5, gate-wired 5/5, tts-test 37/37, tts-mic-precedence PASS, voice-wiring green, flight 231/231; single-transport the inherited 9/10. Merges clean onto `ac76122`.

### Non-blocking note

`audiopanel.paint()`'s unretained `audio:ptt` listener: left as is, deliberately. Every listener paint registers (`audio:mic`, `audio:micfloor`, `audio:hush`, `audio:receive`, `voice-gate-unavailable`) has the same shape, under the file's build-once doctrine. If panels ever get reinitialized that whole family needs teardown ownership together, and it should land as its own change rather than one row of it here.

antra-tess (changes requested):
# Eidoverse Worlds PR #148 — round-3 rereview

- PR: https://github.com/anima-research/eidoverse-worlds/pull/148
- Exact base: `2c7e117def11ad27ad0ea3b26658d77435e92219`
- Exact head: `e2644d5111f8cfa0bb7071d43f4eee88d6069512`
- No microphone, merge, deployment, or live voice state changed.

## Positive evidence

I inspected the round-3 delta and independently reproduced the central focused suites:

```text
mic-ptt-ungated-test   35/35
mic-ptt-test           29/29
micstate-exec          10/10
release                14/14
mesh fallback           5/5
gate wiring             5/5
tts                     37/37
tts mic precedence      PASS
flight                 231/231
```

The core repair direction is right: `micOn()` (lane + device liveness + not muted) now gates audio and `micGateInfo()`, `setMicLive()` reapplies authority synchronously, and mode exit no longer owns a second raw-track writer. The direct OFF/ON/mode-exit sequences pass on both gated and ungated harnesses with real onset ticks.

The broader command later failed only because the detached MCPL TypeScript closure could not resolve `three/webgpu`; that is review-environment closure, not evidence against the focused PTT change.

## Blocking race: mic OFF during an in-flight shared acquisition

The new `voicesfu.js` fast-path repair adds `setMicLive(true)` after a caller awaits a same-PC `micPending`, but it does not re-check the authoritative current `wantMic` state after that await:

```js
if (micPending && micPendingPc === pc) {
  await micPending;
  if (micStream && !micStream.synthetic) {
    micStream.getTracks().forEach((t) => (t.enabled = true));
    setMicLive(true);
  }
  return;
}
```

This is a real product-shaped race because the file explicitly supports two concurrent same-PC callers (for example credential replay plus a user press) sharing one slow permission/acquisition promise.

I drove the actual `voicesfu.sfuConnect()` / `sfuMic()` path with a deferred `voiceSource()`:

```text
sfuMic(true)       # starts pending acquisition
sfuMic(true)       # second caller awaits same micPending
sfuMic(false)      # user turns microphone OFF before permission resolves
resolve raw device # acquisition completes
await both callers
```

Exact-head result:

```json
{
  "wanted": false,
  "reportedOn": false,
  "trackEnabled": true,
  "setMicLiveCalls": [false, false, true]
}
```

Thus the user’s authoritative OFF remains visible in `wantMic=false`, while the resumed waiter re-enables the device track and reports device liveness true. In the actual micstate this can make `micOn()` true whenever the retained lane exists, contradicting the round-3 invariant that microphone control outranks the key/regime.

The prior head `6b79289` already re-enabled the raw track in this race, but did not make the new final `setMicLive(true)` call:

```json
{
  "wanted": false,
  "reportedOn": false,
  "trackEnabled": true,
  "setMicLiveCalls": [false, false]
}
```

So the underlying transport race is inherited, but the exact round-3 change extends it into the newly authoritative liveness state. The round-3 claim is not complete until every post-await re-enable is current-intent-bound.

Probe artifact:

`files/research/eidoverse-pr148-pending-off-race.mjs`

SHA-256:

`afa43ca391b5a993f2266cf16b7415bde5484fc70b08a6f3697ab47719c52e72`

## Required repair

Add an executable voicesfu lifecycle discriminator—not a regex/source pin—that proves:

```text
shared slow mic acquisition
→ second same-PC waiter
→ mic OFF before acquisition resolves
→ acquisition resolves
```

leaves all of these false/closed:

- `wantMic` / public mic intent;
- device/raw track enabled state;
- micstate device-liveness authority;
- actual published device source.

After every relevant await, re-check the current mic intent and PC/generation ownership before enabling tracks or calling `setMicLive(true)`. A device acquired after OFF must be stopped/disabled and must not become the published live source. Preserve the existing synth-provider behavior when mic OFF intentionally selects synthetic voice.

Add a mutation that removes the post-await current-intent check and makes exactly this race turn red. Existing round-3 PTT/VA sequences and their mutations should remain green.

## Non-blocking note

The retained `audiopanel.paint()` listener shape is reasonably deferred as a family-level teardown issue under the current build-once doctrine. It is not the blocker above.

slimepriestess:
## Round 4 — the acquisition answers the intent current when it lands. Head `34e7c9a3c24dfa96c2cb1189687ffbdeca3e7b85`

Reproduced your sequence, and the race was wider than the waiter site. The waiter's `setMicLive(true)` was the newly visible symptom; the inherited half was that `sfuPublish` itself published the device live after OFF, and the OFF's own synth swap never happened because its `sfuPublish` merely joined the pending. So the round-3 fix extended a hole it didn't dig, as you said, and closing only the waiter would have left `trackEnabled:true` in place with the liveness call gone.

### What changed

- **`sfuPublish` re-reads `wantMic` after the acquisition await.** A source that answers a stale question is stopped and disabled, never gated or published, and the outer function asks the current question: a wanted mic is acquired; an unwanted one is replaced by the synth only when a provider can speak for you. With no provider there is nothing to publish, and it says so rather than re-entering, since re-entering would acquire the very device just dropped, forever.
- **The mirror race is covered by the same check.** Mic ON during a synth acquisition drops the synth and acquires the mic. Same class, same seam, so it's in the same discriminator.
- **The waiter applies nothing after its await.** The publish it waited on acted on the intent current at its own resolution. Anything the waiter did afterward would be a second, staler answer, which is exactly what round 3's re-enable was. PC ownership is checked at the re-publish decision as well.
- `micstate` exports `micDeviceLive()` so a harness can tell "no lane" from "lane, device off". The UI keeps asking `micOn()`.

### The executable discriminator

`tools/sfu-mic-intent-race-test.mjs` (23/23) imports the shipped `voicesfu.js`, `micstate.js`, `micgate.js`, `voiceconsent.js`. Stubs stop at the browser boundary: a fake `RTCPeerConnection` that records `addTrack`, the gate-unbuildable WebAudio context with ungated consent so the published track is the device track and its enabled flag is the wire, and `voiceSource()` as a deferred acquisition resolved by hand. A 5 s watchdog makes a parked-forever caller a red rather than a hang.

1. Your sequence, no provider: ON → ON (shared prompt) → OFF → resolve. Device stopped and disabled, not published, `micDeviceLive()` false, `micOn()` false, `sfuMicWanted()` false, no further acquisition.
2. Same with a provider: the device is dropped, the synth acquisition the OFF wanted opens with `micWanted:false`, the synth is what gets published, liveness false.
3. The mirror: device live → OFF (synth acquisition parks) → ON before it lands → synth dropped, mic re-acquired, published and live.
4. The round-3 retained-device fast path, executed: OFF disables and reports not live; ON re-enables without a new acquisition and reports live. This replaces the regex pin in `mic-ptt-ungated` (now 32).

Mutations, each restored before the next:

| mutation | result |
|---|---|
| acquisition does not re-read intent | 12 red: device published live in 1, 2, 3, then the watchdog |
| old waiter re-enable restored, acquisition check intact | green: the acquisition never hands it a device |
| both (the round-3 head's shape) | red incl. `micDeviceLive()` true, your exact signature |
| dropped source not replaced | 1 red (case 2's synth never acquired) |
| fast path without `setMicLive(true)` | 1 red (case 4) |

The second row is worth saying plainly: the waiter's check is not independently discriminable by a product-shaped sequence once the acquisition is intent-bound, because every drop path leaves `micStream` null. That is why the waiter now does nothing rather than carrying a check that nothing can exercise.

Neighbors unchanged: mic-ptt 29/29, mic-ptt-ungated 32/32, micstate-exec 10/10, release 14/14, mesh-fallback 5/5, gate-wired 5/5, voicesource-real 9/9, tts-test 37/37, tts-mic-precedence PASS, flight 231/231; single-transport the inherited 9/10. Merges clean onto `ac76122`.

antra-tess (approved):
# Eidoverse Worlds PR #148 — round-4 rereview

- PR: https://github.com/anima-research/eidoverse-worlds/pull/148
- Advertised base: `2c7e117def11ad27ad0ea3b26658d77435e92219`
- Prior reviewed head: `e2644d5111f8cfa0bb7071d43f4eee88d6069512`
- Exact head: `34e7c9a3c24dfa96c2cb1189687ffbdeca3e7b85`
- Current upstream main at review: `ac76122aec2d4df970a82f085e0a667e30b735bf`
- Disposition: **APPROVE**
- No microphone, merge, deployment, or live voice state changed.

## Blocking race is repaired

Round 3 added device liveness to microphone authority but left a wider in-flight acquisition race:

```text
sfuMic(true) starts slow device acquisition
second sfuMic(true) waits on same pending
sfuMic(false) lands before permission resolves
raw device resolves
```

The old acquisition still published a now-unwanted device, and the parked waiter then re-enabled it and called `setMicLive(true)`. Exact round-3 signature was:

```json
{"wanted":false,"reportedOn":false,"trackEnabled":true,"setMicLiveCalls":[false,false,true]}
```

Exact head `34e7c9a` repairs the lifecycle at its owner:

1. `sfuPublish()` snapshots the requested source intent before `await voiceSource(...)`.
2. After the await it compares current `wantMic` to the snapshot.
3. A stale device/synth is disabled and stopped before publication.
4. The outer acquisition owner re-asks the *current* question: acquire mic if ON; acquire synth if OFF and a provider is available; otherwise publish nothing.
5. A same-PC waiter applies no track/liveness state after its await.
6. The mirror race (ON during synth acquisition) follows the same rule.

I ran the shipped product-shaped discriminator against actual `voicesfu.js`, `micstate.js`, `micgate.js`, and `voiceconsent.js`, with only RTCPeerConnection/WebAudio/source acquisition stubbed:

```text
sfu-mic-intent-race-test  23/23
mic-ptt-ungated-test      32/32
mic-ptt-test              29/29
micstate-exec             10/10
release                   14/14
mesh fallback              5/5
gate wiring                5/5
```

The race test proves:

- OFF during device acquisition stops/disables the late device and never publishes it;
- with a synth provider, OFF drops the late device and publishes the newly acquired synth;
- ON during a synth acquisition stops the stale synth, reacquires the device, and publishes the device;
- retained-device OFF→ON reports liveness and does not reacquire unnecessarily.

Round-3 neighboring evidence also remained green in my exact-head run:

```text
TTS                       37/37
TTS mic precedence        PASS
flight                    231/231
```

A subsequent detached TypeScript command failed to resolve `three/webgpu` from `mcpl/agent.ts`; this is review dependency closure inherited from the detached worktree, not a PTT source/test failure.

## Independent mutations

I reproduced the contributor's intentionally asymmetric mutation result.

### Disable acquisition's post-await intent check

```text
8 ok / 12 failed
exit 1
```

The late device was not stopped, it was published, synth and mirror paths failed, and the watchdog caught a parked waiter.

### Restore only the stale waiter's post-await re-enable

```text
23/23
exit 0
```

This is honestly non-killing because the acquisition owner has already dropped the stale source and re-asked the current question before the waiter resumes. The waiter no longer owns the bug by itself.

### Disable acquisition check and restore stale waiter

```text
5 ok / 15 failed
exit 1
```

This recreates the original signature class: stale publication plus stale re-enable/liveness claim.

The test therefore binds the actual lifecycle owner rather than demanding that every old-looking line be independently load-bearing.

## Current-main applicability

Upstream main was `ac76122a...`. The PR's merge base remains `2c7e117...`, but a no-commit merge of exact head `34e7c9a...` onto current main completed cleanly. On the composed tree I reran:

```text
intent race               23/23
ungated PTT               32/32
gated PTT                 29/29
micstate exec             10/10
release                   14/14
```

The trial merge was aborted and the worktree restored cleanly.

## Non-blocking debt

The retained `audiopanel.paint()` listener shape remains family-level teardown debt under the current build-once doctrine. It is not changed by this repair and should be addressed with its sibling listeners when panel reinitialization becomes real.

## Conclusion

The exact-head implementation fixes the round-3 privacy/authority race at the acquisition owner, covers the provider/no-provider and mirror-source cases, uses an executable product-shaped discriminator, and is mutation-sensitive at the correct seam. **APPROVE** for source merge only. Deployment and live microphone acceptance remain separate operations.

Issue #42, opened 2026-08-06
Non-Chromium browsers: commons performance collapses in Firefox and Safari; Chrome remains usable


antra-tess:
## Field report

Two independent human visitors, Digi and Deckard, report major performance/framerate problems in Firefox in the live commons. Both switched to Google Chrome and recovered enough to participate. Deckard independently raised optimization as his main issue during his second visit; Digi then supplied the cross-browser discriminator.

This is stronger than “commons is heavy”: same world/user session class, browser swap, materially different usability. Exact Firefox versions, OS/GPU, HUD FPS, and failure shape remain pending.

## Preserve before tuning

For each reporter when convenient (do not require another bad session tonight):

- Firefox version and OS/GPU
- Chrome version on the same device
- HUD FPS / frame-time if visible
- WebGPU adapter/backend and any console warnings
- whether failure is low steady FPS, progressive decay, stutter, input lag, shader compilation, or memory growth
- quality tier, pixel ratio, grass/flora density, shadows, sky/weather state
- cold load versus after several minutes

## First comparison

Use one fixed commons position/camera and the same account/device. Capture Firefox and Chrome with:

- renderer/adapter info
- FPS/frame-time over 60s
- memory if available
- scene toggles one at a time: flora, sky/weather, shadows, avatar count, post effects

Do not “fix” by silently degrading shared world facts. Visual quality may differ by browser; shared state may not.

## Acceptance

- identify the dominant Firefox bottleneck or unsupported path with a reproducible receipt
- Firefox gets a legible quality/fallback path rather than becoming accidentally unusable
- no regression to Chrome/WebGPU baseline
- browser/adapter capability and selected quality tier are inspectable in the HUD/debug receipt

## Provenance

Field report: Digi and Deckard, commons, 2026-08-06. Both independently moved from Firefox to Chrome because of major performance issues.

antra-tess:
Additional field report from Antra: **Safari also performs badly** in commons. Current discriminator is now broader than Firefox: Firefox and Safari are poor/unusable, while Chrome is the known usable path for the same world.

This raises priority on renderer/backend/capability differences and the absence of a legible non-Chromium fallback. The fixed-scene comparison should include all three browsers on matched Apple hardware where possible, and record WebGPU adapter/backend, feature limits, pixel ratio, selected quality tier, and console warnings. Do not assume Firefox and Safari share one root cause merely because both lose to Chrome; test them separately under the same acceptance harness.

[...]

PR #150, opened 2026-08-28
browserlab: the same 25 seconds, in two browsers, on one machine (#42)

cormundus:
Evidence for #42. **No optimization patch — this is the instrument.**

**base** `main` `6006d6d` · **head** `4005ae3` · two commits · isolated worktree, nothing unrelated bundled, no Skye/Gabriel foliage work touched.

## Why an instrument first

#42 has never lacked frame rates. It has lacked *comparable* ones. Digi and Deckard left Firefox because it was bad; N8python measured ~40fps on an M3 Max and ~120 with the grass out of frame; Chrome itself collapsed to 1fps at 4.6M triangles with 41 skinned VRMs. Each of those is true, and none can be subtracted from another, because each was taken in a different place, with a different crowd, in a different window.

So: one machine, one account, one world, one **exact** camera pose, UI hidden, three foliage arms in a fixed order, raw `requestAnimationFrame` deltas reported as **p50 / p95 / p99 / max** rather than a smoothed FPS. The complaints on this issue are stutter and input lag, which live in the tail; `perf.ms` is an EWMA built for a HUD and cannot see them.

| arm | meaning |
|---|---|
| `full` | everything on, as the world ships |
| `static` | the meadow is **drawn but not animated** — pusher displacement early-outs, every auto-ticked system (wind, gust, billboards, tile ticks) unhooked |
| `off` | the meadow is **not drawn** — the ceiling on what foliage costs |

## What landed

| file | |
|---|---|
| `client/lib/browserlab.js` | the in-page harness (`await EW.browserlab()`) |
| `client/lib/controller.js` | `setPhotoCamera` / `getPhotoCamera` |
| `tools/browserlab-run.mjs` | Playwright driver over **installed** browsers |
| `tools/browserlab-seed.mjs` | reproducible local scene; loopback-only |
| `tools/browserlab-compare.mjs` | the comparability gate |
| `docs/browser-perf-receipt.md` | procedure, checklist, receipt template |
| `tools/receipts-42/` | Chrome 151 + Firefox 154 specimens + comparison |

`--browser` takes installed browsers: `chrome` is your Chrome, `moz-firefox` is **stock Firefox over WebDriver BiDi**, not Playwright's patched build — a receipt from a bundled fork of Firefox would not be evidence about Firefox.

The harness sends no verbs and persists nothing; every toggle is browser-local and restored in a `finally`, including on a throw mid-arm. `browserlab-seed` *does* write verbs, and refuses any host that is not loopback: a world log is permanent history.

## The specimen — and the limit on it

Chrome 151 and Firefox 154, Windows 10, RTX 2080 SUPER, identical camera pose, identical 1280×800 buffer, identical scene (2,725,031 triangles, 85,910 blades, 0 other people), **local lab world, not the commons**:

| arm | Chrome 151 p50/p95/p99 | Firefox 154 p50/p95/p99 |
|---|---|---|
| full | 16.67 / 16.69 / 16.75 | 16.66 / 16.68 / 16.68 |
| static | 16.67 / 16.68 / 16.69 | 16.66 / 16.68 / 16.68 |
| off | 16.67 / 16.68 / 16.70 | 16.66 / 16.68 / 16.68 |

**Firefox is not slower than Chrome here — and that is a floor result, not a tie.** The compare tool says so in those words: this scene leaves both browsers headroom to reach the refresh interval, so the comparison can only report that neither failed. It does not contradict the field reports and does not confirm them; it does not reach them. The variable the evidence points hardest at — 41 skinned VRMs — is not in this specimen at all.

One difference did survive the gate: **Firefox exposes no WebGPU adapter vendor or architecture** where Chrome reports `nvidia` / `turing`, and no `deviceMemory`. #42's acceptance asks that adapter capability be inspectable in a debug receipt; on Firefox that half is currently unavailable to any receipt, including this one.

## The gate is the point

`browserlab-compare` checks camera pose, drawing buffer, people present, triangles, blades and run length **before** it prints a delta, and exits non-zero claiming nothing when they differ. Two gates it caught during this work:

- the browsers, left to their own window chrome, gave **1249×1285 and 1280×955** on the same monitor — Firefox drawing 24% fewer pixels, a rendering advantage nobody granted it. Now pinned by `--size`.
- the client door key is **`?key=`, not `?token=`** — the first three runs measured an *empty world* while looking perfectly healthy: page booted, renderer drew, three identical arms. The driver now refuses a world with no entities unless `--allow-empty=true`.

## Four ways this harness lied before it stopped

Each is a comment in the code, because each would have produced a confident wrong finding:

1. **Cadence lock.** The first Chrome run returned `off: p50 1000.06, p95 1000.11, p99 1000.11` and would have been published as *"hiding foliage made it 60× slower"*. `document.hidden` was `false` the whole time. Caught now by distribution shape: a renderer under load spreads (p50 < p95 < p99), a throttled tab hands out identical timer ticks.
2. **Backgrounded tab.** Refused outright at start; going background mid-run taints the receipt.
3. **Async meadow.** Snapshotting the grass field at start meant `off` hid nothing, `static` froze an auto-list that was still empty, and `restore` put the empty list back — three arms, one scene, a receipt saying foliage was free. The field is resolved late, every arm, and waited for before the first one.
4. **Blade census.** `grassTiles()` reports planted truth and does not change when a mesh is hidden, so the `off` arm claimed it was still drawing 85,910 blades. Counted from mesh visibility now.

## Portability note for the repo generally

**Playwright's browser launch hangs under bun on Windows.** Measured on Windows 10 + bun 1.3.14: `chromium.launch()` starts the browser process and then hangs to the launch timeout, because Playwright talks over `--remote-debugging-pipe` (fds 3/4) and bun's spawn does not hand those over. The identical call under node connects in **223ms**. That is why the driver is `.mjs` run under node. `tools/probe-harness.mjs` is exposed to the same wall on this platform — flagging it rather than touching it here.

## Known gaps

1. **Zero people in the specimen.** The dominant variable in the field evidence is absent. Reaching the reported regime needs a populated commons or a lab world seeded with a comparable body count.
2. **Safari untested.** No Apple hardware on this box. The arms and the gate are browser-agnostic; the driver would need a WebKit path.
3. **Vsync floor.** At 60Hz with headroom, p50 is clamped and the comparison can only detect a browser that *fails* to hold the refresh interval. A higher-refresh display or a heavier scene would raise the ceiling.
4. **`client/lib/browserlab.js` ships in the client bundle** (94 modules parse, was 93), following `grassdiag.js`'s precedent as a console-reachable diagnostic. If diagnostics should not ride the shipped client, say so and it moves behind a lazy import.
5. **No optimization is proposed.** Deliberately. The instrument first, then the fix it earns.

Suites on this head: `smoke` 84 pass / 1 fail — `✗ rtc reaches its recipient`, which fails identically on a `main` control worktree and is not this change.

Directed-by: Cormundus


cormundus:
## Round 2 — all six blockers closed, plus the mechanical one

**New exact head `b56c48422bee7a0428e9eed58b823038f60e3af3`** · base `main` `6006d6d` · four commits on the reviewed head `4005ae3`.

`git diff --check 6006d6d..b56c484` → **clean, exit 0**. The `tools/receipts-42/comparison.md:54` blank line is gone, and the receipts are regenerated rather than trimmed.

Thank you for the counter catch in particular — it was not a rounding problem, it was the wrong field, and the receipts deserved to be withdrawn.

---

### 1. Draw-call receipts were the wrong field, not a bad heuristic

You were right that `dc > c0 * .5` could not work; the deeper problem is that it was guarding a field that was never per-frame. three's info object documents both and does not disambiguate at the call site:

| field | three's own doc |
|---|---|
| `render.calls` | "since the app has been started" — **lifetime** |
| `render.drawCalls` | "of the current frame" — **per frame** |
| `render.triangles` | "of the current frame" — **per frame** |

Measured on the live client across six consecutive frames:

```
calls       1434 → 1437 → 1440 → 1443 → 1446     (+3/frame — three render passes)
drawCalls   92 → 92 → 92 → 92 → 92 → 92
triangles   2456705 → 2456705 → 2456705 → 2456705
```

So the committed receipts reported ~6,186 "draws per frame" for a scene that draws **92**, growing arm over arm because time passed rather than because the arm changed anything. Magnitude was the wrong question — once a lifetime total is large, one frame's growth is never half of it.

`classifyCounter()` asks **shape** over consecutive samples instead: flat is per-frame, evenly growing is a running total reported as its **step**, anything else returns `null` with a reason. A `null` prints as `unknown` in the receipt; it never prints as a number.

Mutation-sensitive test: the real six-frame lifetime series is fed in and must classify as cumulative with a per-frame meaning of **3**, not 1446. The old heuristic is reconstructed alongside and asserted to **fail** on that same series, so the defect is pinned as data rather than described. Five malformed shapes — shrinking, wobbling, jumping, too-short, empty — must all return `null`.

**Re-run: 110 draws/frame with the meadow, 67 without.**

### 2. The static arm no longer touches anything but the meadow

You are right that emptying `_autoParticleSystems` could not support the attribution the arm exists to make — `autohooks.js` names the sky and every entity emitter as co-owners in its first paragraph.

The arm now releases only what the meadow owns, **by identity**: each stroke's own wind clock (`update(t){ uT.value = t }`) and whatever the field registered on `field.autoHooks`, out through `autohooks.releaseHook` and back on the way out. It also zeroes the wind amplitudes, because freezing the clock alone leaves blades stopped mid-gust and leaning.

It counts the foreign hooks it left running, and the receipt prints that count:

```
_Static arm scope: 3 meadow-owned hooks released and 1 wind amplitudes zeroed;
N non-meadow hooks (sky, weather, entity emitters) left running — N at start._
```

The test **plants a sentinel hook** in the shared array before the static arm and asserts it kept *ticking through it* and is still in the array afterwards. That check exists because the first version merely counted foreign hooks and passed vacuously at `0 === 0` — this scratch world registers no sky or emitter hooks of its own, so the scope claim had nothing to be true about.

### 3. Cleanup is inside the finally, and tested by a thrown arm

`trouble.stop()` is inside the guaranteed finally. `stop()` is idempotent, and restores only what is still ours, so a watcher layered on top is not clobbered by this one unwinding.

`tools/browserlab-cleanup-test.mjs` throws a **real** error from inside the arm loop (`_throwOnArm`, a test seam, not a simulation) and then reads the page back: `console.error`/`console.warn` native again, meadow drawing, every wind amplitude at its exact prior value, every released hook back in the array, photo mode exited, body class removed. Twice — mid-run and on the very first arm — then a clean run as control. **20 checks green.**

### 4. The seeder has no default world

Agreed that loopback is not proof nobody lives there. The default is a generated `browserlab-<random>` nobody can already be standing in; a world named explicitly must be **empty**, or carry `--allow-existing` for a scratch world you made yourself. It prints the exact `--url` it just created, and the runner has **no default world either** — it exits telling you to seed one.

### 5. The gate compares the scene, not three proxies for it

Correct that camera + people count + triangle total can agree while a body moved, an entity was hidden, an asset finished streaming, or the log advanced. The gate now compares a **scene digest** — every entity's id, asset, quantised transform and visibility; every person's id, position and avatar — plus the folded **world log seq** (`net.lastSeq`, newly exposed read-only) and the **build**.

Mutated across all seven blind spots (moved, hidden, re-assetted, walked, re-avatared, advanced seq) with reorder and sub-millimetre float jitter as negative controls, and the specific hole named as its own check: *"same camera + people + triangles is NOT enough on its own"*.

And the scope claim is now made explicitly, in the comparator's own output rather than only in a doc:

> _Scope: this comparison is valid for FROZEN SEEDED SCENES — a world whose entities and population do not change between the two runs, which the scene-digest and world-seq gates enforce. It is NOT yet validated for a live populated commons, where bodies move between sequential browser runs by definition; there the two runs would have to be simultaneous, or the scene frozen first._

### 6. The receipts name their build

`server build 6006d6d (DIRTY TREE)` identified a commit the tree was not. The driver now computes a deterministic tree digest — the commit, a hash over the full diff against it, and every untracked file's path and content hash — and hands it into the page, so the receipt carries it. Same tree, same digest, any machine.

One correction found while doing it: the digest counted `tools/receipts-42/` as part of the tree, which made it self-referential — writing a receipt dirtied the tree the next receipt then reported as dirty, so a clean checkout could never produce a receipt saying so. The output directory is excluded now, and the exclusion is named in the digest label.

Both committed receipts were produced on commit **`08276ca`**, tree digest **`2947d074a629935f`**, **clean checkout**. `git diff 08276ca..b56c484` touches only `docs/` and `tools/receipts-42/` — the harness code those receipts name is byte-identical to the head.

---

## The re-run, and what moved

| | withdrawn | now |
|---|---|---|
| draws/frame | 6186 / 11058 / 15930 | **110 / 110 / 67** |
| drawing buffer | 1280×800 vs 960×600 | **1280×800 both** |
| scene digest | — | `2e324e3b` both · world seq 26 both |
| build | `6006d6d (DIRTY TREE)` | `08276ca` · tree `2947d074a629935f` · clean |

**A seventh thing your gate caught, once it was strong enough to.** The first re-run failed the buffer row: Chrome at 1280×800, Firefox at 960×600 — Firefox's frame governor had shed a quarter of its pixels under cruise, with nothing in the receipt saying so. The gate refusing was the gate working, but a browser comparison should not be letting an adaptive dial change the workload mid-experiment at all. The render scale is **pinned** for the run and restored after; `renderScale: 'auto'` measures with the governor deliberately. An invalid value throws, and so does a pin that does not take — `setRenderScale` refuses unknown values and returns the current one, which would have left the governor in charge silently.

**The verdict is unchanged, and still a floor rather than a tie.** Every arm in both browsers sits on the 16.66ms vsync interval; the comparator says so in those words. Zero people — the variable the field evidence points hardest at is still absent from this specimen, and the doc says so.

One difference survives all nine gates: **Firefox exposes no WebGPU adapter vendor or architecture** where Chrome reports `nvidia` / `turing`, and no `deviceMemory`. #42's acceptance asks that adapter capability be inspectable in a debug receipt; on Firefox that half is unavailable to any receipt, including this one.

## Suites on this head

| | |
|---|---|
| `bun run tools/browserlab-core-test.ts` | **57 / 57** — mutations for the counter classifier, the throttle verdict, the scene digest, the gate, and the refusals |
| `node tools/browserlab-cleanup-test.mjs` | **20 / 20** — thrown arms, the sentinel hook, the live counter contract |
| `node tools/browserlab-compare.mjs …` | **9 gates pass, exit 0** |
| `bun run tools/smoke.ts` | 84 pass / 1 fail — `✗ rtc reaches its recipient`, identical on a `main` control worktree |
| `git diff --check 6006d6d..b56c484` | clean |

The arithmetic and every refusal now live in `client/lib/browserlab_core.js` — DOM-free, THREE-free, dependency-free — so the claims a reviewer has to trust are testable without a browser, and the harness and the comparator cannot drift apart about what "comparable" means.

Your environment note matches what I see: that is `client/`'s `three` dependency (`cd client && bun install`), not the patch. No merge or deployment implied; this is still the instrument, with no optimization proposed.

— I.M. & Cormundus

cormundus:
## Follow-up: the ordering point from #151 applies here too

**New exact head `b55f8616cc807108aa5602f9460e81a757a535d0`** (was `b56c484`), base `main` `6006d6d`, two commits.

Raising this rather than leaving it for your next pass to find: the exactness item you flagged on #151 — `autoHooks().push(...)` restoring membership but not **position**, while the comment claims the array is left as it was found — has an exact twin here in `client/lib/browserlab.js`'s static arm. Same defect, same reasoning: the engine drains that array in order.

Fixed the same way: each hook's index is recorded, removal runs back-to-front so the indices stay valid as the array shortens, and restoration splices front-to-back so each lands before the ones still to come.

Two tests can now see it:

- **headless** — the source guard now refuses an appending restore (`live.splice(` present, `autoHooks().push(...` absent). It also un-pins a check that had been asserting the old call's exact spelling rather than the property.
- **in the browser** — the cleanup test plants sentinels on **both sides** of the meadow's hooks (`sentinelBefore` at the head, `sentinelAfter` at the tail) and requires the whole array back in the order it was found with the neighbours on their original sides. The single appended sentinel it had before could not have caught this: it survives an appending restore looking untouched.

`core 58/58`, `cleanup 22/22`.

**Receipts regenerated.** Changing `browserlab.js` meant the committed receipts named `08276ca` while the head no longer matched — the exact binding blocker 6 asked for. They are rerun on the clean new head: nine gates pass, exit 0, tree digest `2dca5e7881a5f388`, identical scene digest `2e324e3b` and world seq 26 in both browsers. The numbers are unchanged, as expected of a change that only affects the order hooks go back in — 110 draws/frame with the meadow, 67 without, every arm on the 16.66ms vsync interval. Still a floor, not a tie.

If you had rather I left this until your pass on `b56c484` finished, say so and I will hold future cross-branch fixes until the head you are reading is done with.

— I.M. & Cormundus

antra-tess (changes requested):
Exact head reviewed: `b55f8616cc807108aa5602f9460e81a757a535d0`.

The six original review blockers are repaired. Independent exact-head evidence:

- `browserlab-core-test`: **17/17**
- `browserlab-seed-test`: **14/14**
- `browserlab-wire-test`: **23/23**
- cleanup harness controls: **14/14**
- comparator reproduces the committed lawful verdict: same camera/buffer/scene/world seq/duration/build, both browsers pinned to the ~16.67ms interval, therefore a **floor**, not a browser tie
- the live cleanup contract, on a seeded owned scratch world, is fully green: throws on both static and first arm unwind console/listeners/photo/UI/meadow; clean run leaves the page exactly as found; foreign hooks keep ticking; meadow hooks leave and return at exact indices; cumulative renderer calls are not published as per-frame cost
- range `diff --check` clean

**CHANGES REQUESTED on the remaining product-receipt seam:** the documented command is not reproducible on the declared Node floor, and the receipt still does not identify the external visual implementation it measured.

### 1. The documented seeder command fails on Node 20

`tools/browserlab-seed.mjs` says:

```bash
node tools/browserlab-seed.mjs
```

but reaches for bare global `WebSocket`. On the house's Node 20.20.2 it fails before joining; I had to run the seeder under Bun. PR #151 already carries the right portable `ws` resolution pattern. Please either import/resolve a declared WebSocket implementation or explicitly document/enforce a runtime where the global exists. The command in the docs must be the one that works.

### 2. The receipts bind the Worlds code tree, but not the external library/assets that materially define the measured scene

With the exact PR head and local companion checkout, the page rendered but the meadow never planted. Instrumentation showed the tracked `vegetation.js` importing three absent generator modules, then (after composing those) five absent meadow fit/map assets. Once the exact public-production artifacts were supplied, the unchanged seeded cleanup test passed every assertion.

The same issue applies more strongly to performance receipts: vegetation code/shaders/maps and the three spawned GLBs are external to the Worlds tree, and their bytes affect triangles, shader cost and frame time. `treeSha256` names the harness/client checkout, but the committed receipt carries no static-root identity or effective asset/module digest. Two nominally identical runs can therefore measure different visual implementations while both claiming the same code-under-test.

Please make the benchmark own or explicitly validate and receipt that seam:

- document the required `EIDOVERSE_DIR` / static-root setup;
- fail early with exact missing paths rather than waiting for `the meadow to plant`;
- emit a stable digest (or per-artifact hashes) for the effective vegetation module + transitive generators/maps and the seeded model assets used by the run;
- include that identity in the comparison gate/receipt, so cross-browser runs cannot compare different visual implementations.

This is a provenance/reproducibility blocker, not a claim that the corrected harness logic is wrong. With one frozen production artifact set composed, the cleanup and comparison behavior are good.

Issue #60, opened 2026-08-08
Client: resident-controlled local grass density / off switch


antra-tess:
## Lived report

Digi in commons: “can we have a setting to turn down the grass locally? my poor gpu.”

The client already has the correct local primitive:

- `client/lib/terrain.js::setGrassDensity(f)`
- sticky across flora re-grows
- the frame governor automatically steps `1 → 0.6 → 0.35`
- shared authored flora state remains unchanged

What is missing is a resident-controlled UI/persisted preference. Today a client must wait for the automatic governor or use a developer console.

## Proposed bounded slice

Expose a **local-only grass density/quality control** in the existing client quality/sky settings surface, e.g.:

```text
auto/full | medium (0.6) | low (0.35) | off
```

or an equivalent labeled slider.

Requirements:

1. never writes a world `grass` verb or changes shared species/seed/provenance;
2. applies through `setGrassDensity` to every composed flora stroke/stem;
3. persists per browser/client and survives re-grow/reconnect;
4. automatic governor may lower the effective density under load but must not silently raise above the user's chosen cap;
5. effective and chosen values are inspectable;
6. `off` must retire or draw zero locally without destroying shared field state;
7. one browser at low/off and another at full still agree on authored world facts;
8. add a browser/unit receipt for setting before field load, after load, and across replacement.

Related performance umbrella: #42, but this control is useful across browsers and GPUs and should remain a narrow local-quality slice.

PR #151, opened 2026-08-28
foliage: the meadow can stand still now, and only for you

cormundus:
The third of Mica's three bounded pieces: the smallest browser-local eye that hides foliage, with off/static/full semantics and a proof that one viewer's choice changes no shared facts.

**base** `main` `6006d6d` · **head** `c22eb72` · two commits · isolated worktree, browser-local only, no world verb, no production deployment, no Skye/Gabriel foliage work touched.

## The source map changed the answer

Most of this already ships. `client/lib/grass_quality.js` has carried the resident's cap — `full · medium · low · off`, persisted per browser, capped against the frame governor's session shed — since #60, and its own header had already drawn the line this task is about:

> Neither dial is ever a verb — the shared field (species, seed, extent, provenance) is world state; how many blades THIS machine fills pixels with is not.

`off` was already the real thing: zero instances, group hidden, field object intact, shared state untouched.

| control | where | scope | persisted | a verb? |
|---|---|---|---|---|
| resident's cap | `grass_quality.js` `makeGrassQuality` | this browser | `ew-grass-quality` | **no** |
| governor's shed | `governor.js` → `terrain.setGrassDensity` | this session | no | no |
| what draws | `terrain.js` `applyGrassBudget()` — `min(cap, shed)`; at `0`, count 0 **and** group hidden | | | |
| applied truth | `terrain.js` `getGrassApplied()`, `grassTiles()` | | | |
| the `grass⚙` row | `build.js` — select + aria-live, repaints on `grass-budget` | | | |
| **the field itself** | the `grass` verb — rank 2, owner-only (`server/rights.ts`) | **the world** | the world log | **yes** |
| wind | each stroke's `update(t){ uT.value = t }` → `_autoParticleSystems` | per-frame hook | | |
| pushers | `flora.js` `wirePushers`, gated by `freezePushers` | | | |
| per-tile ticks | `f._tileTick` via `pushHostHook`, on `field.autoHooks` | | | |

## What was actually missing: `static`

There was no way to keep the meadow **drawn** and stop it **moving**. The only route was the diagnostic path, which empties `_autoParticleSystems` wholesale — a bucket that also carries the sky and every entity emitter. Fine for a four-second A/B; not something a resident can leave switched on.

That gap is not cosmetic. #42's own field evidence points at motion rather than fill: an M3 Max at ~40fps with the meadow in frame and ~120 with it out, and `grassdiag` exists because the suspicion was shader-side displacement rather than blade pixels.

## A second dial, not a fifth rung

Density and motion are different costs for different reasons — *"my poor gpu"* is one, *"the swaying makes me ill"* is another, and someone who wants a still meadow at full detail should not have to mow it to get one. The cap keeps its four levels; motion is `on | off`, persisted under `ew-grass-motion`, by exactly the same rules.

The three names are **points** in that two-dial space:

| name | cap | motion |
|---|---|---|
| `full` | full | on |
| `static` | full | **off** — drawn, not animated |
| `off` | off | *moot: nothing drawn can move* |

`animates()` returns false at cap `off` whatever the motion dial says, so no surface ever has to explain a meadow claiming to sway while invisible. `preset()` returns `null` when the dials sit between names — `medium` is a legitimate place to stand and it is not `static`.

### How the freeze is applied

`terrain.js` `applyGrassMotion()`, narrowly and **by identity**:

- each stroke's wind-clock hook is released from the shared array (`autohooks.releaseHook`) and pushed back on the way out;
- that stroke's `base`/`gust` wind amplitudes are zeroed and restored to their exact prior values. Freezing the clock alone leaves blades stopped mid-gust, leaning; zeroing amplitude settles them upright, which is what *still* should look like.

Deliberately **not** frozen, each for a reason:

- **the per-tile ticks** — they re-settle LOD and visibility against the camera; freeze them and a walking viewer drags a stale meadow behind them. A still meadow is the ask, a wrong one is not.
- **the pushers** — a motionless field that still parts around your feet is interaction, not ambient animation.
- **anything that is not foliage** — emptying `_autoParticleSystems` would stop the clouds, the weather and every entity emitter for as long as one resident left one setting on.

A re-grow re-applies the choice against the new field, the same way the cap is already sticky.

### Honest limit, on the record

**`static` is a comfort setting, not an established performance one.** The shader still evaluates its wind term at zero amplitude, and no measurement on this hardware separates static from full — both sit on the vsync interval (`tools/receipts-42/`, PR #150). The lever that demonstrably removes work is `off`. If `static` should also be cheap, that is a measurement task on hardware where the frame is not already free — not a change to this dial.

## The surface

```
/foliage                      what your meadow is doing, and how to change it
/foliage full | static | off  the three names
/foliage medium | low         the cap's other rungs
/foliage sway on|off          the motion dial alone
```

`/grass` aliases it. In the settings panel, `sway⚙` sits under `grass⚙`, is disabled when the cap is `off` (nothing to move), and repaints on the same `grass-budget` event.

## The product-door test

The claim is a product claim, so the proof is a product proof. A unit test on the dial cannot establish it — the dial could be perfect and some caller could still emit a verb.

`tools/foliage-door-test.mjs` stands **two real viewers in one real world**, moves one of their dials through the real command path, and asks the world and the other viewer whether anything happened to them. Two separate browser **contexts**, not two tabs: `localStorage` is per origin, and two tabs of one profile share the resident's dial by design. Isolating them is what makes them two people rather than one person twice.

Three sources of truth, because a weaker one alone would pass a broken build:

1. **the world log on disk** — the world IS its log; if it did not grow, nothing was authored. Checked byte-identical from first assertion to last.
2. **the other viewer's rendered meadow** — blades drawn, applied density, wind amplitudes, their `localStorage` — read from their page, not inferred.
3. **the acting viewer's own meadow, which must change** — a test where nobody's view moved would pass while proving nothing.

```bash
node tools/foliage-door-test.mjs     # node, not bun — see docs/browser-perf-receipt.md
```

**24 checks green** (`tools/receipts-foliage/door-on-branch.txt`).

**Negative control: 13 named failures** on `origin/main` `6006d6d` with only the test file copied in (`door-fail-on-main.txt`) — no `/foliage` route, no motion dial, ash's meadow does not go dark, the wind keeps blowing. The world-log assertions **pass** on main, and should: a build with no local foliage control authors nothing because it can do nothing. Those passes are the invariant, not the feature.

`tools/grass-quality-test.ts` covers the dial headless — **76 passed, 0 failed** — including three source-level guards that are the real risk here: that `terrain.js` freezes by **identity** rather than by emptying the shared hook array, that the per-tile ticks are **not** in the freeze, and that the `/foliage` handler's own body sends no verb of any kind (sliced to itself — the next handler in the file sends `force`, and a generous byte window happily convicted this one of it).

`git diff --check 6006d6d..c22eb72` is clean. `smoke` 84 pass / 1 fail — `✗ rtc reaches its recipient`, which fails identically on a `main` control worktree.

## Two harness scars worth keeping

- **`page.waitForFunction` with an async predicate resolved immediately.** The dynamic `import()` a terrain probe needs makes the predicate a promise; both viewers baselined at zero blades and every later comparison was against nothing. The waits poll through `page.evaluate` now, where the awaiting is not in doubt.
- **Receipts are trimmed of the trailing blank line** the capture redirect appends, which `git diff --check` rejects — applied here before committing rather than after (thanks to the #149 review).

## Open questions for review

1. **Persistence.** The brief said "no durable state". This reads that as no *shared* durable state and follows the shipped precedent — `consent.js` and `grass_quality.js` are both `localStorage`-backed so a choice survives the session that made it. If the intent was no persistence at all, dropping the `ew-grass-motion` write is a one-line change; the test's reload section then inverts.
2. **`static` at other caps.** `preset()` names only full+motion. `medium` with motion off is reachable and legal but unnamed. Should it have a name, or is unnamed correct?
3. **Whether `static` should be cheap.** See the honest limit above.
4. **The sway row's placement.** It sits under `grass⚙`. It could instead be an accessibility setting, since motion sensitivity is the strongest reason to want it.

Full source map and rationale in `docs/foliage-local-eye.md`.

Directed-by: Cormundus

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -20,6 +20,7 @@
       "devDependencies": {
         "@happy-dom/global-registrator": "^20.11.1",
         "playwright": "^1.62.1",
+        "ws": "^8.21.1",
       },
     },
   },
--- a/client/lib/build.js
+++ b/client/lib/build.js
@@ -15,6 +15,7 @@ import { loadGLB, libLabels, listLibrary } from './assets.js';
 import { makeLightGizmo } from './lights.js';
 import { entities, entityMeta, comps, findPart, editHolds } from './world.js';
 import { surfaceUnder, reindexCollider } from './colliders.js';
+import { getGrassMotion, setGrassMotion } from './terrain.js';
 import { heightAt, GRASS_QUALITY, getGrassQuality, setGrassQuality,
   getGrassDensity, getGrassShed, getGrassApplied } from './terrain.js';
 import { sendVerb, sendDrag } from './net.js';
@@ -1152,6 +1153,36 @@ function paintSky(body) {
   syncGrassRow();
   bus.on('grass-budget', syncGrassRow);   // governor sheds repaint immediately
   body.appendChild(gqRow);
+
+  // Whether YOUR meadow sways. A second local dial rather than a fifth rung on
+  // the cap: density and motion are different costs for different reasons —
+  // "my poor gpu" is one, "the swaying makes me ill" is another, and someone
+  // who wants a still meadow at full detail should not have to mow it to get
+  // one. Off/static/full (#42) are the two dials' named points, reachable here
+  // or as `/foliage full|static|off`.
+  const gm = document.createElement('input');
+  gm.type = 'checkbox';
+  gm.style.cssText = 'accent-color:var(--accent); margin:0;';
+  gm.setAttribute('aria-label', 'grass sway — local only, never shared with the world');
+  const gmRow = mkRow('sway⚙', gm);
+  const syncMotionRow = () => {
+    gm.checked = getGrassMotion() === 'on';
+    // Nothing is drawn at cap off, so nothing can sway: disable rather than
+    // offer a control that would silently do nothing.
+    const nothingToMove = getGrassQuality() === 'off';
+    gm.disabled = nothingToMove;
+    gmRow.title = nothingToMove
+      ? 'nothing to sway — the grass cap is off'
+      : `your meadow ${gm.checked ? 'sways' : 'is still'} — local setting, not shared with the world`;
+  };
+  gm.onchange = () => {
+    setGrassMotion(gm.checked ? 'on' : 'off');
+    syncMotionRow();
+    flashHint(`grass sway: ${gm.checked ? 'on' : 'off'} (yours only)`);
+  };
+  syncMotionRow();
+  bus.on('grass-budget', syncMotionRow);
+  body.appendChild(gmRow);
 
   // Render scale is YOURS too (§22k) — the whole frame's pixel budget, the
   // one lever a pixel-bound machine actually answers to (§22j's tables).
--- a/client/lib/chat.js
+++ b/client/lib/chat.js
@@ -690,6 +690,9 @@ function runCommand(raw) {
     case 'pushable':
       bus.emit('command', { cmd: 'pushable', arg });
       return true;
+    case 'foliage': case 'grass':
+      bus.emit('command', { cmd: 'foliage', arg });
+      return true;
     case 'eyes':
       bus.emit('command', { cmd: 'eyes', arg });
       return true;
--- a/client/lib/commands/handlers.js
+++ b/client/lib/commands/handlers.js
@@ -22,6 +22,10 @@ import { toggleHelp, flashHint } from '../ui.js';
 import { sceneAttach, sceneDetach } from '../scenegraph.js';
 import { EMOTE_ORDER, EMOTES } from '../avatar.js';
 import { setPushable, pushable } from '../consent.js';
+import {
+  GRASS_QUALITY, getGrassQuality, setGrassQuality, getGrassShed,
+  getGrassMotion, setGrassMotion, setFoliagePreset, getFoliagePreset,
+} from '../terrain.js';
 import { trySitOn } from '../localbody.js';
 import { getMe } from '../mybody.js';
 import { setMyReach, clearMyReach } from '../reachnet.js';
@@ -224,6 +228,34 @@ register('pushable', (arg) => {
   const v = (arg || '').trim().toLowerCase();
   if (v === 'on' || v === 'off') setPushable(v === 'on');
   return logChat('*', `shoves and blasts ${pushable() ? 'CAN' : 'can NOT'} knock you over${v ? '' : ' — /pushable on|off to change'}`);
+});
+
+// The meadow through YOUR eyes. Two local dials — how much is drawn (the
+// resident's cap, #60) and whether it sways — plus the three names #42 uses
+// for the points that matter: full, static, off.
+//
+// Nothing here is a verb. The shared field — species, seed, extent, who
+// planted it — is world state and stays exactly as its author left it; how
+// many blades this machine fills pixels with, and whether they move, is not
+// world state and never was. Another viewer standing beside you sees their
+// own answer, unchanged. (Proof, not assertion: tools/foliage-door-test.mjs.)
+register('foliage', (arg) => {
+  const v = (arg || '').trim().toLowerCase();
+  const say = () => {
+    const preset = getFoliagePreset();
+    const shed = getGrassShed();
+    return logChat('*', `your meadow: ${getGrassQuality()} · sway ${getGrassMotion()}`
+      + (preset ? ` (${preset})` : '')
+      + (shed < 1 ? ` — the auto governor is thinning to ×${shed}` : '')
+      + ' — yours only, never shared. /foliage full|static|off|medium|low|sway on|off');
+  };
+  if (!v) return say();
+  if (v === 'full' || v === 'static' || v === 'off') { setFoliagePreset(v); return say(); }
+  if (GRASS_QUALITY.includes(v)) { setGrassQuality(v); return say(); }
+  const m = /^(?:sway|motion|wind)\s+(on|off)$/.exec(v);
+  if (m) { setGrassMotion(m[1]); return say(); }
+  if (v === 'sway' || v === 'motion' || v === 'wind') { setGrassMotion(getGrassMotion() === 'on' ? 'off' : 'on'); return say(); }
+  return logChat('*', `no such foliage setting "${v}" — try full, static, off, medium, low, or sway on|off`);
 });
 
 register('boom', (arg) => {
--- a/client/lib/grass_quality.js
+++ b/client/lib/grass_quality.js
@@ -19,13 +19,32 @@
 export const GRASS_QUALITY = ['full', 'medium', 'low', 'off'];
 export const QUALITY_DENSITY = { full: 1, medium: 0.6, low: 0.35, off: 0 };
 
+// A SECOND local dial, orthogonal to the first: whether the meadow MOVES.
+//
+// The cap answers "how many blades does this machine fill pixels with". It
+// cannot answer "draw them, but stop them swaying", and that is a different
+// question with a different answer — a machine can be fill-bound or bound by
+// the per-vertex wind and pusher work, and #42's field evidence points at the
+// second (grass out of frame: 40fps → 120fps on an M3 Max). Density and motion
+// are not one ladder, so they are not modelled as one.
+//
+// The three names #42 asks about are POINTS in that two-dial space, not a
+// third dial:
+//   full   = cap 'full'  + motion on
+//   static = cap 'full'  + motion off      (drawn, not animated)
+//   off    = cap 'off'                     (not drawn; motion moot)
+export const GRASS_MOTION = ['on', 'off'];
+
 const KEY = 'ew-grass-quality';
+const MOTION_KEY = 'ew-grass-motion';
 
 /** The dial state. `store` is localStorage in the browser, anything with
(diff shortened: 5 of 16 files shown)


cormundus:
## Round 2 — both lifecycle blockers closed, and the ordering claim made true

**New exact head `0d0993fef2355fcf67da7bdeb2c92760f4f8a6a5`** · base `main` `6006d6d` · one commit on the reviewed head `c22eb72`.

`git diff --check 6006d6d..0d0993f` → clean. Both blockers were in the harness rather than the product, and the second one cost you a hand-killed process — sorry about that one.

---

### 1. The advertised command now runs on your Node, verified on your Node

Not simulated. I fetched a portable **node 20.20.2** and confirmed the environment first:

```
$ node --version
v20.20.2
$ node -e "console.log('global WebSocket:', typeof WebSocket)"
global WebSocket: undefined
```

The socket constructor is resolved through `tools/owned-lifecycle.mjs` — the global where there is one, the `ws` package where there is not — and `ws` is a **declared devDependency** now rather than the transitive accident it had been. The whole test then runs on that runtime:

```
the WebSocket contract, on this node and on node 20
  ✓ a WebSocket resolves here — ws package (node 20.20.2 has no global)
  ✓ with no global WebSocket (node < 22) it falls back to the ws package
  ✓ this file never reaches for the bare global WebSocket
…
all green — one viewer changed their eyes, and the world did not notice
```

**43/43 on node 20.20.2** (`tools/receipts-foliage/door-node20.txt`) and **43/43 on node 24.18.0** (`door-on-branch.txt`). The receipt is now produced by the same command you can run.

### 2. One owner for the whole lifecycle, and the child is proven rather than trusted

Every spawned thing — sequencer, seed socket, browser, both viewer contexts — is registered with a single idempotent `Lifecycle` **the moment it exists**, not when the happy path is finished with it. That is the specific gap that orphaned :9471: spawn, readiness, browser launch, seed socket and viewer creation all sat outside the try. The Lifecycle is bound to `SIGINT`, `SIGTERM`, `uncaughtException` and `unhandledRejection`, disposes newest-first, and attempts every disposer even when one throws.

On ownership: the child is spawned with `EIDO_BOOT_NONCE` and must **echo that nonce** from `/version` — the same identity test `tools/probe-harness.mjs` already uses for the same reason. A responder with no `nonce` field is a stale listener by definition. The port is checked free over **TCP** before the spawn and again after teardown, rather than trusting that a kill worked.

### 3. Hook ordering — the claim is now true rather than nearly true

You are right that `autoHooks().push(...frozenHooks)` restored membership and not position while the comment said the array came back as found. The engine drains that array in order, so the claim has to hold: each hook's index is recorded, removal runs **back-to-front** so those indices stay valid as the array shortens, and restoration splices **front-to-back** so each lands before the ones still to come.

Rewriting it surfaced a latent bug worth naming: the wind-amplitude loop had been folded into the hook loop, so a stroke carrying uniforms but no live hook would have kept its wind. They are separate passes now.

---

## The mutation controls

They run **before** the product assertions, because a receipt from a harness that cannot fail is not a receipt.

**Control 1 — the runtime contract.** The global is deleted at runtime, reproducing node 20 on any node, and the fallback must carry the test. The file is also scanned for a reintroduced bare `new WebSocket(` — which would pass on node 22 and fail on yours, i.e. exactly this defect coming back. (That check initially convicted its own explanatory comment; it strips comments before scanning now.)

**Control 2 — ownership.** Three listeners, and the receipt records what each proves:

```
  ✓ the impostor answers 200 — generic HTTP readiness would accept it
  ✓ …but the nonce proof refuses it
  ✓ …and refuses a listener echoing somebody else's nonce
  ✓ the control cleaned up after itself
```

**Control 3 — a pre-viewer failure leaks nothing.** A real throw immediately after the child is up, in the exact position the old code had no try around, then a TCP check:

```
  ✓ a child came up and PROVED it is ours (:9228, pid 25872)
  ✓ …and the port really is listening
  ✓ the injected failure propagated
  ✓ …and the child is gone: nothing listens on that port
  ✓ …and dispose is idempotent (a second call is a no-op)
```

`--fault=post-server` runs the same thing as a by-hand check. `tools/receipts-foliage/lifecycle-fault.txt` carries the listener census before and after — unchanged, and both survivors are Logitech G Hub processes on :9010 and :9180, neither of them mine. The run exits 1 on the deliberate throw and leaves nothing behind.

**And one new product assertion** for the ordering fix: sentinel hooks are planted *in front of and behind* the meadow's own, and the whole array must come back in exactly the order it was found, neighbours still on their original sides. Membership alone would not have caught an append.

---

## Receipts on this head

| | |
|---|---|
| `door-node20.txt` — node 20.20.2, your runtime | **43 / 43** |
| `door-on-branch.txt` — node 24.18.0 | **43 / 43** |
| `door-fail-on-main.txt` — same files on `6006d6d`, node 20 | **14 named failures** |
| `lifecycle-fault.txt` — deliberate fault + listener census | no child survived |
| `bun run tools/grass-quality-test.ts` | **76 / 76** after the ordering change |
| `bun run tools/smoke.ts` | 84 pass / 1 fail — `✗ rtc reaches its recipient`, identical on a `main` control worktree |
| `git diff --check 6006d6d..0d0993f` | clean |

The negative control moved from 13 to 14 named failures because the new order check fails there too (`the meadow hooks really left the array during static — 5 → 5`: on main nothing is ever released). The harness's own three controls **pass** on main, which is right — they test the harness, not the feature, and a control that failed for the wrong reason would be worth less.

---

## One sibling, flagged rather than left for you to find

The ordering defect you caught here has an exact twin in **#150**: `client/lib/browserlab.js` restores its released hooks with `autoHooks().push(...)` too. You are mid-review there at a fixed SHA, so I did not move it under you. I am pushing the same index-restoring fix and an order-sentinel test to that branch now, as a separate head — say the word if you would rather I had left it alone until your pass finished.

— I.M. & Cormundus

antra-tess (changes requested):
Exact head reviewed: `0d0993fef2355fcf67da7bdeb2c92760f4f8a6a5`.

The product direction and runtime implementation are good. Independent receipts on the exact head:

- `grass-quality-test.ts`: **76/76**
- exact-base negative with the new test copied forward: red at the missing motion surface (`setMotion is not a function`)
- Node 20.20.2 uses the declared `ws` package; no bare global `WebSocket`
- generic-200 and wrong-nonce listeners are refused
- injected post-spawn failure tears down the owned child and frees the port; disposal is idempotent
- server + MCPL bundles complete; range `diff --check` clean
- with the current production vegetation dependencies composed, the real two-viewer door is fully green: Ash off/static/full changes only Ash; Brook's blades, dials, hooks and storage remain byte-for-byte/semantically untouched; the world log never grows; reload preserves Ash's local choice; hook order and neighbours restore exactly; cleanup frees the owned listener.

**CHANGES REQUESTED: the committed product receipt is not reproducible from the documented command because the effective external grass library is neither declared nor identified.**

On this review host, the documented exact command initially timed out for three minutes at `ash's meadow to plant`. The owned lifecycle did clean up correctly. Instrumenting the same owned run showed:

1. tracked `upstream-patched/eidoverse/vegetation.js` imports three files absent from the checkout/static root:
   - `vegetation_shrub_gen.js`
   - `vegetation_corn_gen.js`
   - `vegetation_sunflower_gen.js`
2. after composing those, the field still failed because the five `eidoverse/assets/grass/blades_meadow_*` fit/map files were absent.
3. after supplying the exact public-production artifacts, the unchanged official test passed every product-door assertion.

The three production generator SHAs used were:

- corn `4779e33bf783afe97498a09f04f056a0ee6b6b4bc75f466de3a131925046e897`
- shrub `911a578b04fe617b6e5a0424706ca691a4a95add5f13132502204f4da2bce00c`
- sunflower `ef16d42d8134b0f785ec56c902cd4dd97743c3a6a5cff21e08f1e99ba0028859`

and grass assets:

- albedo `b24ff239678539f83fc2e46f88c34900bb5e137cc0bae592865a588ab4e6375c`
- fit `d479d265b576e2fe4968c78882bfe9ad0a683ccd579c421b766736ca72d30696`
- normal `e8af1d506fa964d2b894f402e532491b255522ec1d8bc29e6ff0a302f917c705`
- roughness `0e88c0b18c8137e4b67b765c19ae03ea344315b2688061516429bbcf418d0ac9`
- translucency `70e99085b9bfa9e96e0c080a3d7455c28d8c8e479f57697ee59a1b8817562878`.

`docs/foliage-local-eye.md` currently advertises only:

```bash
node tools/foliage-door-test.mjs
```

and the committed receipt identifies Node/runtime but not the static-root/artifact identity. That lets two reviewers run nominally the same product proof against different vegetation implementations, or wait three minutes for a generic meadow timeout.

Please make the product door own or explicitly validate this dependency seam before opening viewers:

- document the required `EIDOVERSE_DIR` / static-root setup;
- preflight the effective vegetation module and required transitive generator/map files through the same server route;
- fail early with the exact missing path;
- emit a stable digest (or per-artifact hashes) into the receipt so the product proof names both the client code and the external visual implementation it exercised.

This is a receipt/provenance blocker, not a request to change the local foliage semantics. I found no product-code blocker once the effective production assets were actually present.

PR #153, opened 2026-08-30
sdk: document that a rebind starts kv empty (by design)

octopusburrow:
One doc paragraph, no code. `behavior.d.ts`'s kv comment promises state "survives restarts, replays, and forks" — all true — but says nothing about the lifecycle event an iterating author hits most often: re-issuing `behavior` for the same id. `fold.js` is explicit that this is intended ("a rebind keeps nothing: fresh code starts with fresh state unless the same id folds a later bstate"), so this PR just surfaces that contract where authors actually read.

How it was found: building a small room-script tonight whose entire point was remembering who had used it (a givers list feeding an on-leave effect), I lost the kv twice in one evening across upload+rebind iteration cycles, and nothing at the SDK surface predicted it. The reset is a sound design (fresh code shouldn't inherit another version's schema) — it just deserves a sentence.

Wording is two claims and a practical hint; happy to trim to taste.

### DIFF
--- a/sdk/behavior.d.ts
+++ b/sdk/behavior.d.ts
@@ -65,7 +65,15 @@ interface WorldAPI {
   /** Persistent state, private to this behavior. Survives restarts, replays,
    *  and forks (it is event-sourced under the hood: one coalesced `bstate`
    *  entry per activation that changed something). Whole store ≤8KB —
-   *  counters and flags, not archives. */
+   *  counters and flags, not archives.
+   *
+   *  ⚠ A REBIND starts EMPTY, by design: re-issuing the `behavior` verb for
+   *  the same id folds a fresh record, and fresh code gets fresh state
+   *  (shared/fold.js, the binding case: "a rebind keeps nothing") — earlier
+   *  bstate does not carry across. So iterating on a live script loses its
+   *  kv on every upload+rebind cycle; if the state matters across versions,
+   *  read it out before rebinding, or design keys the next version can
+   *  rebuild without. */
   kv: {
     get(key: string): unknown;
     set(key: string, value: unknown): void;   // undefined/null deletes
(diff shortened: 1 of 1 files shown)


octopusburrow:
One more thought, since this has sat a few days: the doc paragraph is the floor, but the underlying pain — losing kv on every upload+rebind iteration — has a real code shape, and I'd be glad to expand this into that PR if there's appetite. Two replay-safe options, both of which only affect *new* events so no existing world re-folds differently: (a) an opt-in `keep: true` on the `behavior` verb, meaning "this is an iteration on the same script, carry its kv"; or (b) a one-shot `onRebind(oldKv)` hook so the new version decides what to keep. Default stays fresh-code-fresh-state, which is the right conservative rule. Happy to write either up as a design note first if that's the preferred route, or leave this as the one-paragraph doc fix.

antra-tess (approved):
APPROVE — exact head `cba349db20028eb2726a57bc6d677832cb13f9cc`.

The two documented contracts match the runtime:

- `shared/fold.js` replaces a behavior binding record without carrying prior `state`, so a rebind starts kv empty until later `bstate` entries fold.
- `server/behaviors.ts::hostKv()` marks every `set` dirty without equality comparison, and `flushKv()` appends one `bstate` after the activation; equal-value timer writes therefore do grow the replay log.

`diff --check` is clean, the two-commit docs-only delta is confined to `sdk/behavior.d.ts`, and it composes cleanly onto current main `d2f9e48e10c3441da2279128308337987c6aed42` via `git merge-tree`.

No code or deployment claim is implied.

slimepriestess (approved):
**Approve.** Both claims in the new paragraph are true of the code today:

- Rebind starts empty: `shared/fold.js:386` (`behavior` case) says exactly this — "a rebind keeps nothing: fresh code starts with fresh state unless the same id folds a later bstate" — and the `bstate` case (`:391`) only ever replaces `rec.state` from a later entry.
- An equal-value `set()` still counts as a change: `server/behaviors.ts` `hostKv` sets `kvDirty = true` on every `set` regardless of the prior value, and `flushKv` appends a `bstate` entry whenever the flag is set. A timer that re-sets unchanged state does write an entry per activation. "Compare before you set" is the correct advice.

Doc-only; nothing else to check. The `⚠` in a `.d.ts` comment renders fine in editors that show JSDoc.

PR #156, opened 2026-08-30
lod: geometry LOD for placeable objects — bodies fail closed, by structure

gabriel-berzescu:
Mica, Antra — geometry LOD for placeable objects, the server half of the lane agreed in devchat: **objects first; bodies excluded by positive structural detection; no silent attempt.** This lands inert — nothing asks for a tier until the client chooser (its own PR) — the same way §20 itself landed.

*Round two (`e291c7d`) addresses the review's seven points — see the thread. Summary below is current.*

## The contract, point by point

- **Bodies fail closed, structurally — not by well-formedness.** `lodExclusion` reads the **raw JSON chunk** before any Document exists. Skins; **any `JOINTS_n`/`WEIGHTS_n` set**; VRM metadata in `extensionsUsed`, `extensionsRequired`, **or the top-level `extensions` object** (a malformed producer is still a body); morph targets → exit 2 with the typed verdict **`unsupported: skinned/avatar asset`**, nothing written. And **any animation is a typed v1 refusal** — `unsupported: animated object (v1 reduces static geometry only)`: preservation of channels through the diet is evidence a later recipe earns, not a claim this one makes.
- **Original preserved byte-identically.** The variant is a sibling (`<rel>.lod1.glb`); the suite asserts the original's bytes unchanged after both a refusal and a success.
- **Identity binds source hash + recipe + tools.** `asset.extras` gains `{ lodOf: <source sha256>, recipe, tools: { meshoptimizer, encoder } }` — merged over the source's own extras, never clobbering them.
- **Preservation proven or refused.** After the reduce: the node **hierarchy with per-node transforms**, and **per-primitive material assignments**, are signature-checked; scene bounds within tolerance; vert count actually reduced. Any failure → typed verdict, never a half-valid object. Already-light objects (< 12k verts) and ineffective reductions refuse honestly too. **Scope, stated plainly:** the LOD mesh is a *visual* representation — the client contract (follow-up PR) keeps collision/support/raycast semantics anchored to the original's geometry, with tier upgrade on approach.
- **Chosen before fetch, original fallback — and the recipe is a GENERATION.** The URL carries it (`?lod=<recipe>`, taken from `/version`'s `lodRecipe`, never a default) and so does the filename (`<rel>.lod.<recipe>.glb`): a recipe change is a fresh URL and a fresh file, nothing to un-pin, and the pump retires the old generation's file when the new one lands. Every fall-through stays provisional, an **unrecognized** `lod` value forces provisional too (the next process may negotiate it), and a **mutable library source newer than its variant is never served the variant** — the §20c freshness discipline, with the next boot's sweep rebuilding it.
- **No artifact leaks.** `isLodVariant` joins the serving-artifact predicate; catalogs, listings, and the boot sweeps all inherit the existing rules; verdicts are recipe-stamped so a future recipe re-measures refusals (#146's doctrine, reused).

The diet itself is the house recipe end to end: the ktx2 diet (dedup/prune/resample + texel-budget KTX2 + draco) with `weld() + simplify(MeshoptSimplifier, ratio 0.25, error 0.01)` — the avatar importer's own proven reducer — in between. An untextured object needs no encoder; a textured one on an encoder-less box is exit 3, environmental.

## Measured, on the library

| model | original | lod1 | |
|---|---|---|---|
| oil drilling rig | 55.1 MB | **2.1 MB** | 26× |
| cyborg brain | 49.9 MB | **2.0 MB** | 25× |
| cyborg heart | 46.5 MB | **1.8 MB** | 26× |
| server rack (246k verts, disjoint hard edges) | — | *refused: reduction ineffective* | honest |
| `claude_suit.vrm` | — | *refused: `unsupported: skinned/avatar asset (skins)`, 5 ms* | fail closed |

## Receipts

`bun tools/object-lod-test.ts` — **72/72**, fake encoder (no KTX-Software needed; the real-encoder receipt runs when present), on a **process-owned harness** (logs piped, per-run content-hashed fixture as the ownership nonce, a failed boot prints its log tail). Each seam of the review has a named control — deleting the mechanism turns its control red:
- the exclusion, pure — including the review's probes: a VRM extension *object* with no `extensionsUsed`, and `JOINTS_1/WEIGHTS_1` without set 0;
- the CLI: skinned and **animated** fixtures refused typed (originals byte-identical after); the identity stamp with the source's own extras surviving; already-light, ineffective-reduction, and textured-no-encoder verdicts — the last answered from the raw container in milliseconds, since the pump retries it every boot;
- the store door: upload → the recipe-named shadow → `?ktx2=<key>&lod=<recipe>` serves it immutable; the pre-landing fetch and an **unrecognized generation** both answer `no-cache`; an uploaded body gets the typed marker and its tier fetch falls back whole; listings and catalogs show no artifact;
- **the library arm**: a pre-existing library GLB gets both its lod and ktx2 variants (the shared-`seen` bug that killed this arm is dead), and a mutated source stops the variant serving until the next boot rebuilds it;
- **no encoder**: an untextured upload still reduces; the textured one keeps its original with no marker and one log line;
- **the mutation canary**: the on-disk recipe rewritten under a live child — `/version` holds the running generation, its URL serves, the next generation's URL answers `no-cache` and not the old bytes.

Also on this box: `store-variants-test` 123/123 and `git diff --check` clean. The neighboring `one-libvips` and `deps-route` suites pass here but reproduce failures in other environments — standing debt outside this delta, reported separately rather than folded into a green claim.

## Not in this PR

The client chooser (a follow-up PR): tier choice in `scheduleLoad` from distance + a `models⚙` dial, swap-on-repromote through the existing residency machinery, and the low-VRAM browser receipt. Bodies remain a later, separately-reviewed lane per the contract.

### DIFF
--- a/server/optimize.ts
+++ b/server/optimize.ts
@@ -378,11 +378,28 @@ const sceneBounds = (doc: Document): [number[], number[]] => {
 
 export type LodResult = { out: Uint8Array | null; verdict: string | null; before: number; after: number };
 
+/** The node contract, as a signature: names, transforms, mesh-bearing, and
+ *  hierarchy (child order), per scene. Exported so its DETECTION power is a
+ *  unit test of its own, separate from the pipeline that consumes it. */
+export const lodNodesSig = (d: Document): string => {
+  const walk = (n: any): any => [n.getName(), n.getTranslation(), n.getRotation(), n.getScale(),
+    n.getMesh()?.getName() ?? null, n.listChildren().map(walk)];
+  return JSON.stringify(d.getRoot().listScenes().map((sc) => sc.listChildren().map(walk)));
+};
+/** Which material each primitive of each mesh wears, by index. */
+export const lodMatsSig = (d: Document): string => JSON.stringify(d.getRoot().listMeshes().map((m) =>
+  [m.getName(), m.listPrimitives().map((p) => { const mt = p.getMaterial(); return mt ? d.getRoot().listMaterials().indexOf(mt) : -1; })]));
+
 /** The lod1 diet: the ktx2 diet (dedup/prune/resample + texel-budget KTX2 +
  *  draco) with weld + meshopt-simplify between — and the preservation
  *  asserts around the reduce. `verdict` non-null = a typed content refusal
  *  (marker's business); textures follow --ktx2's all-or-nothing. */
-export async function optimizeGlbLod(bytes: Uint8Array, encoder: string | null): Promise<LodResult & Ktx2Tally> {
+/** `mutate` is the MUTATION-CONTROL SEAM demanded by the #156 re-review:
+ *  the test suite injects a post-reduce corruption (a renamed node, a
+ *  swapped material assignment) and asserts the guard REFUSES it — so
+ *  deleting either guard turns a named test red instead of leaving the
+ *  suite green. Production callers (the CLI) never pass it. */
+export async function optimizeGlbLod(bytes: Uint8Array, encoder: string | null, mutate?: (doc: Document) => void): Promise<LodResult & Ktx2Tally> {
   const none = { eligible: 0, converted: 0, failed: [] as string[] };
   const rawJson = parseGlb(bytes).json;
   const excluded = lodExclusion(rawJson);
@@ -393,26 +410,26 @@ export async function optimizeGlbLod(bytes: Uint8Array, encoder: string | null):
   if (!encoder && (rawJson?.images?.length ?? 0) > 0) return { out: null, verdict: "__no_encoder__", before: 0, after: 0, ...none };
   const io = await getIO();
   const doc = await io.readBinary(bytes);
-  // same head as the ktx2 diet, so the two variants agree on structure
-  await doc.transform(dedup(), prune(), resample());
+  // The node contract is captured BEFORE any destructive transform (re-review
+  // of #156, blocker 1: a plain prune() deleted an empty named socket helper
+  // before the old post-head signature existed — the loss was invisible).
+  // prune() runs with keepLeaves so named empty helpers — socket frames,
+  // attachment points — survive the head at all; the whole-diet signature
+  // then PROVES nothing was lost, or the variant is refused.
+  const preNodes = lodNodesSig(doc);
+  await doc.transform(dedup(), prune({ keepLeaves: true }), resample());
   const before = totalVerts(doc);
   if (before < LOD_MIN_VERTS) return { out: null, verdict: `already light (${before} verts < ${LOD_MIN_VERTS})`, before, after: before, ...none };
-  // What the world depends on, captured AFTER the shared head (parity with
-  // the ktx2 variant) and asserted after the reduce. Not just names (review
-  // of #156, point 6): the HIERARCHY with each node's transform, and each
-  // primitive's material ASSIGNMENT — the reduce may only touch vertex data.
-  const nodesSig = (d: Document) => {
-    const walk = (n: any): any => [n.getName(), n.getTranslation(), n.getRotation(), n.getScale(),
-      n.getMesh()?.getName() ?? null, n.listChildren().map(walk)];
-    return JSON.stringify(d.getRoot().listScenes().map((sc) => sc.listChildren().map(walk)));
-  };
-  const matsSig = (d: Document) => JSON.stringify(d.getRoot().listMeshes().map((m) =>
-    [m.getName(), m.listPrimitives().map((p) => { const mt = p.getMaterial(); return mt ? d.getRoot().listMaterials().indexOf(mt) : -1; })]));
-  const preNodes = nodesSig(doc), preMats = matsSig(doc), [preMin, preMax] = sceneBounds(doc);
+  // material assignments are captured after the head — dedup may merge
+  // byte-identical materials, which is the ktx2 variant's existing behavior;
+  // what may not change from HERE on is which material each primitive wears
+  const preMats = lodMatsSig(doc);
+  const [preMin, preMax] = sceneBounds(doc);
   await doc.transform(weld(), simplify({ simplifier: MeshoptSimplifier, ratio: 0.25, error: 0.01 }));
+  mutate?.(doc);   // the mutation-control seam (tests only) — see the param doc
   const after = totalVerts(doc);
-  if (nodesSig(doc) !== preNodes) return { out: null, verdict: "preservation failed: node hierarchy/transforms changed", before, after, ...none };
-  if (matsSig(doc) !== preMats) return { out: null, verdict: "preservation failed: material assignments changed", before, after, ...none };
+  if (lodNodesSig(doc) !== preNodes) return { out: null, verdict: "preservation failed: node hierarchy/transforms changed", before, after, ...none };
+  if (lodMatsSig(doc) !== preMats) return { out: null, verdict: "preservation failed: material assignments changed", before, after, ...none };
   const [postMin, postMax] = sceneBounds(doc);
   for (let i = 0; i < 3; i++) {
     const tol = Math.max((preMax[i] - preMin[i]) * 0.02, 0.01);
--- a/server/optimize.ts
+++ b/server/optimize.ts
@@ -341,15 +341,24 @@ export async function optimizeGlbKtx2(bytes: Uint8Array, encoder: string): Promi
  *  body is a body (it drops extensions it does not know). */
 export function lodExclusion(json: any): string | null {
   if (json?.skins?.length) return "unsupported: skinned/avatar asset (skins)";
-  const used: string[] = json?.extensionsUsed ?? [];
-  if (used.some((e) => /^VRM/i.test(e) || /^VRMC_/i.test(e))) return "unsupported: skinned/avatar asset (VRM metadata)";
+  // fail-closed means NOT trusting well-formedness (review of #156, point 5):
+  // a VRM that forgot extensionsUsed still carries the extension OBJECT, and
+  // multi-set skinning may skip set 0 — look everywhere the structure can be
+  const extNames = [...(json?.extensionsUsed ?? []), ...(json?.extensionsRequired ?? []),
+    ...Object.keys(json?.extensions ?? {})];
+  if (extNames.some((e) => /^VRM/i.test(e) || /^VRMC_/i.test(e))) return "unsupported: skinned/avatar asset (VRM metadata)";
   for (const m of json?.meshes ?? []) for (const p of m.primitives ?? []) {
     if (p.targets?.length) return "unsupported: morph targets the reducer cannot prove preserved";
-    if (p.attributes && ("JOINTS_0" in p.attributes || "WEIGHTS_0" in p.attributes)) return "unsupported: skinned/avatar asset (joint weights)";
+    for (const k of Object.keys(p.attributes ?? {}))
+      if (/^(JOINTS|WEIGHTS)_\d+$/.test(k)) return "unsupported: skinned/avatar asset (joint weights)";
   }
-  for (const a of json?.animations ?? []) for (const c of a.channels ?? [])
-    if (c.target?.path === "weights") return "unsupported: morph-weight animation";
-  return null;   // node TRS animations are fine — the reduce never touches the node tree, and the assert below proves it
+  // v1 reduces STATIC geometry only. An animated object could survive the
+  // reduce structurally, but "could" is not the contract — preservation of
+  // channels/samplers through the diet is unproven here, and fail-closed is
+  // the reviewer-offered v1 answer. A later recipe can earn animation back
+  // with end-to-end evidence.
+  if (json?.animations?.length) return "unsupported: animated object (v1 reduces static geometry only)";
+  return null;
 }
 
 const totalVerts = (doc: Document) => {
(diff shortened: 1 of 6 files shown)


antra-tess (changes requested):
Exact head reviewed: `705e67dc2cabf3efd7a48600df147c45132c3402`.

The object-only direction is right, and the pure/CLI/store-door evidence is strong after installing the exact root dependency closure: `object-lod-test` passes 49/49, `store-variants-test` passes 123/123, and the original/body refusal paths remain byte-preserving. I am requesting changes before the client chooser, however, because several production seams currently contradict the stated contract.

## 1. The immutable LOD URL does not carry the recipe generation

`/version` publishes `lodRecipe`, but `withLod()` always emits the same `lod=1` URL. A client seeing either `lod1-r25...` or a future `lod1-r10...` asks for the identical immutable URL:

```json
{
  "recipeA": "lod1-r25",
  "recipeB": "lod1-r10",
  "urlA": "store/x.glb?ktx2=3&lod=1",
  "urlB": "store/x.glb?ktx2=3&lod=1"
}
```

That recreates the `?ktx2=2` split-brain on the first recipe change: stale LOD bytes remain pinned under the new recipe. Bind the served/requested URL to the exact running LOD recipe/key, and have the server negotiate only that value. Add a mutation canary where the served file changes recipe while an old process remains running, then prove `/version` + client URL stays on the process generation and the next generation uses a distinct URL.

## 2. The library LOD sweep never queues LODs

The library sweeper shares one `seen` set across modes. The `--ktx2` walk inserts every model rel first; the subsequent `--lod` walk hits `seen.has(rel)` and skips every model. The upload/store arm works, but the declared library arm is dead. Key `seen` by `(rel, mode)` or use separate sets, and add an owned-process test with a pre-existing library GLB (not an upload) whose `.lod1.glb` must appear.

## 3. Missing KTX2 encoder also deletes valid untextured LOD work

The CLI correctly proves an untextured object can reduce without an encoder. The queue cannot: store order is min → `--ktx2` → `--lod`; `--ktx2` exit 3 sets global `ktx2Skip` and removes **all** mode items, including `--lod`. It also prevents later uploads from queueing LOD at all. Separate KTX2 availability from geometry-LOD availability (or classify texture presence before queueing), and test the real pump on an untextured dense object with no encoder.

## 4. Mutable library sources can serve stale LOD bytes immutable

The route serves any existing `<rel>.lod1.glb` without checking it is newer than the winning mutable library/overlay source. The sweeper knows library files are mutable, but there is a window before rebuild; if the client has a new `?v=` and the old LOD exists, the route pins the old body/object under the new immutable URL. Apply the same winning-source freshness discipline before serving LOD (or content-address the source in the URL), and test source update → stale variant provisional/original → rebuilt variant.

## 5. “Bodies fail closed” has structural holes

The raw detector checks only `extensionsUsed` and only `JOINTS_0`/`WEIGHTS_0`. Direct probes currently return `null` for:

```js
{ extensions: { VRMC_vrm: {} }, meshes:[...] }
{ meshes:[{primitives:[{attributes:{POSITION:0,JOINTS_1:1,WEIGHTS_1:2}}]}] }
```

A valid producer should list extensions, and most multi-set skinning also has set 0, but fail-closed is exactly the place not to rely on well-formedness. Inspect top-level extension keys as well as `extensionsUsed`, and reject any `JOINTS_n`/`WEIGHTS_n` set.

## 6. The preservation claims are stronger than the assertions

Current checks compare sorted node **names**, sorted material **names**, and aggregate primitive-local POSITION bounds. They do not prove:

- node transforms / hierarchy / animation channels survived;
- primitive→material assignments or unnamed-material identity survived;
- world-space bounds for transformed/instanced meshes survived;
- seat/support/collider surfaces survived simplification.

Also `asset.extras = {...}` overwrites any source `asset.extras` instead of preserving it.

Because this server half is inert, there are two clean options: (a) narrow the claim to visual geometry and make the follow-up client contract explicitly keep collision/support/raycast semantics off the LOD mesh, while preserving source extras and actually checking node/animation/material assignments; or (b) strengthen the server assertions enough to earn the existing collider/support claim. Overall AABB within 2% is not evidence that a seat pan or support plane is unchanged.

The “node TRS animation explicitly not excluded” test currently exercises only `lodExclusion()`, not a real animated GLB through the diet. Either test preservation end to end or fail closed on animation in v1.

## 7. The new product-door harness is not process-owned

`object-lod-test` probes a random port, then starts a child with stdout/stderr ignored and accepts any responder that comes up. There is no nonce challenge/response and readiness is outside a strong owned-process proof. This can false-green against a race/squatter and hides startup failure (I first saw only `child server came up ... ✗` until running the server manually exposed a missing dependency). Use a kernel-assigned/owned port or nonce-bound `/version`, retain bounded child diagnostics, and ensure startup/readiness live inside awaited cleanup.

Please add mutation/negative controls for the recipe URL, library-mode queue, no-encoder untextured path, stale-source serving, and body structural gates. Deleting each repaired mechanism should turn its named test red.

The neighboring `one-libvips` (`1 failed`) and TTS `deps-route` (`5 failed`) results reproduce outside this six-file delta and are not blockers I attribute to #156, but they should remain separate known debt rather than being included in a blanket “also green” claim.

gabriel-berzescu:
Mica — all seven points were right; two were bugs I shipped, one was me failing to apply my own split-brain lesson to the URL I'd just invented. Repaired at `e291c7d`, each with a named control (deleting the mechanism turns its control red):

**1. The recipe is a generation, in the URL and the filename.** `withLod(url, recipe)` takes the recipe `/version` published — never a default, never a boolean — and the variant lives at `<rel>.lod.<recipe>.glb`: a recipe change is a fresh URL *and* a fresh file, nothing to un-pin; the pump deletes the old generation's file when the new one lands. Defense-in-depth for the window you named: an **unrecognized** `lod` value forces the answer provisional, so nothing this process says gets pinned under a generation the next process may negotiate. The mutation canary rewrites the on-disk recipe under a live child: `/version` holds the running generation, its URL serves the variant immutable, the next generation's URL answers `no-cache` and not the variant.

**2. The library arm is alive.** The shared `seen` set was exactly as you said — the ktx2 walk inserted every rel, the lod walk skipped them all. Keyed by `(mode, rel)`; the control uses a pre-existing library GLB (not an upload) and asserts both its lod and ktx2 variants appear.

**3. Geometry work is not the encoder's hostage.** The lod arm answers no-encoder from the raw container *before any transform* (the boot retry must cost a JSON parse, not a simplify), never sets `ktx2Skip`, is never purged by ktx2's exit 3, and logs its own note once per boot. Control: a no-encoder child reduces the untextured upload while the textured one keeps its original — no variant, no marker, one log line.

**4. Mutable sources are never served stale.** A library lod serves only when newer than the *winning* source (patch > overlay > library — the §20c vrm discipline); older falls through provisional. Control mutates the source under a live child, then reboots and asserts the rebuilt variant serves.

**5. Structure, not well-formedness.** Top-level `extensions` keys and `extensionsRequired` count as VRM metadata; any `JOINTS_n`/`WEIGHTS_n` set is skinning. Both of your probes are tests now, including a `.glb` carrying a VRM extension *object* with no `extensionsUsed` entry.

**6. Claims match assertions — by narrowing and by strengthening.** Strengthened: node hierarchy with per-node transforms, and per-primitive material assignments, are signature-checked; the source's `asset.extras` survive under ours. Narrowed: **any animation is a typed v1 refusal** (`unsupported: animated object — v1 reduces static geometry only`) — preservation through the diet is something a later recipe earns with end-to-end evidence, not something this one claims. And the PR description now says plainly: the LOD mesh is a *visual* representation; the client contract (follow-up PR) keeps collision/support/raycast semantics anchored to the original's geometry summaries, with tier upgrade on approach.

**7. The harness is process-owned — and your point immediately paid for itself.** Logs piped (a failed boot prints its tail), readiness checks the child is alive, ownership is a per-run content-hashed fixture round-trip. Rebuilt this way, the harness caught a real defect in my round-one routes change on the first run: an unimported identifier in the `wantLod` branch **threw at request time inside `Bun.serve`** — the house-rule-3 class — which the ignored-stdio harness had false-greened past. The empty `cache-control` in the piped diagnostics is what named it.

Receipts on this box: `object-lod-test` **72/72**; `store-variants-test` 123/123 unchanged. On the neighboring suites — noted, and agreed: your Mac's `one-libvips` (1) and `deps-route` (5) results are standing environment debt outside this delta; I won't fold them into a blanket "also green" again. On this Windows box both happen to pass, which is a statement about this box, not about the debt.

antra-tess (changes requested):
Re-review of exact `e291c7d0f60cd9aaac93bb50e62187fe21330cec`: **CHANGES_REQUESTED**.

The round-two product repairs points 1–5 cleanly, and most of 6: recipe generation is in URL + filename, the library arms no longer share one `seen`, untextured LOD survives an encoderless host, mutable-library stale variants fall through provisional/rebuild next boot, VRM extension objects and arbitrary `JOINTS_n/WEIGHTS_n` fail closed, animations are explicitly unsupported in v1, source `asset.extras` survives, and object/store suites are green once the exact dependency closure is installed.

- `object-lod-test.ts`: **0 failed**;
- `store-variants-test.ts`: **0 failed**;
- range `diff --check`: clean;
- neighboring one-libvips `1 failed` (no Sharp loaded) and deps-route `5 failed` remain dependency-closure debt outside this six-file delta.

Two blockers remain.

## 1. Empty named semantic nodes are still deleted before preservation begins

The preservation signatures are captured **after** `dedup(), prune(), resample()`. An empty named socket/helper can be removed by that shared head before `preNodes` exists, so the later hierarchy/transform assertion cannot see the loss.

I ran a direct fixture: a dense static mesh node `root` at translation `[2,3,4]` with empty named child `seat_socket` at `[0.2,1.1,-0.3]`. Exact-head `--lod` returned exit 0 and logged:

```text
prune: Removed types... Node (1)
[optimize] lod: 25921 -> 6526 verts
```

Original nodes:

```text
root [2,3,4] children:[seat_socket]
seat_socket [0.2,1.1,-0.3]
```

Variant nodes:

```text
root [2,3,4] children:none
```

That is precisely the named part/socket-frame loss v1 promised to fail closed on.

Required repair: capture the original semantic node contract before destructive transforms and either preserve named empty helpers through prune or refuse the variant. Add an executable empty named child/socket fixture.

The named “preservation control” is also not mutation-sensitive yet. I deleted both runtime `nodesSig` and `matsSig` rejection guards and reran the full object suite; it still reported **0 failed**. Add mutations/fixtures that make removing each guard turn the suite red (for material assignment too, not only natural preservation).

## 2. The harness still proves Eido capability, not spawned-process identity

`startServer()` finds an apparently free random port (TOCTOU), waits for any `/version`, then uploads/fetches a per-run content-hash fixture with the shared literal token `test-door`. A stale/squatter Eido server with that token can accept and round-trip the fixture, satisfying the claimed ownership check. `WORLDS_DIR` is unique, but the store/opt paths exercised by the fixture are checkout-global.

This is not nonce-bound responder identity. Please pass a per-run `WORLD_INSTANCE_NONCE` into the child and assert that exact nonce from `/version` before any product claim, or provide an equivalent child-PID/owned-listener proof. Add a wrong/missing-nonce impostor control, and keep startup/readiness under one cleanup owner.

Nothing is approved, merged, or deployed from this review. The geometry reducer and generation/freshness repairs are otherwise in good shape; these two seams are the remaining production gates.

gabriel-berzescu:
Mica — both blockers repaired at `125fa7e`.

**1. The socket survives, and the guards have teeth.** Your fixture is in the suite verbatim — `root` at `[2,3,4]`, empty named child `seat_socket` at `[0.2,1.1,-0.3]` — and the variant now keeps the socket, with its translation, still under `root`. Mechanism: the node contract is captured from the **pristine** document, before any transform; the head runs `prune({ keepLeaves: true })` so named empty helpers survive it at all; and the whole-diet signature proves nothing was lost, or the variant is refused.

On mutation sensitivity — you were right that deleting my guards left the suite green, so the guards now have executable controls: `optimizeGlbLod` takes a test-only post-reduce `mutate` seam (a parameter the CLI never passes, so it is not reachable from production), and the suite injects a renamed node, a moved node, and a swapped material assignment, asserting each is **refused** with its named verdict. Delete either guard and its control goes red. The detectors themselves (`lodNodesSig`, `lodMatsSig`, now exported) are additionally unit-tested to distinguish a moved node and a swapped assignment.

**2. Ownership is the nonce.** Each child gets a per-run `WORLD_INSTANCE_NONCE`; readiness *is* identity — `/version` must return that exact `instance` before anything counts as up, a foreign responder on the probed port is refused rather than retried into, and the impostor control asserts the verifier rejects a wrong nonce, a missing `instance`, and a null body. Every server section holds its `stop()` in a `finally`. The piped-tail diagnostics also earned their keep again: the one flake during the rework was a cold transpile outrunning the readiness window, visible as an empty tail rather than a bare ✗ — window now 30s.

Receipts on this box: `object-lod-test` **83/83**; `store-variants-test` 123/123 unchanged; `git diff --check` clean. The `one-libvips`/`deps-route` dependency-closure debt remains reported separately, as agreed.

antra-tess (approved):
**APPROVE — exact head `125fa7e5700db3669c18fae0b123a2c4e4446941`.**

Round three closes the two remaining production gates from my prior review.

### 1. Semantic socket preservation now crosses the destructive head

The node signature is captured before `dedup → prune → resample`, and `prune({keepLeaves:true})` preserves named empty helpers. I independently reproduced the exact fixture that failed round two:

- dense mesh node `root` at `[2,3,4]`;
- empty named child `seat_socket` at `[0.2,1.1,-0.3]`.

Exact head reduces the mesh while preserving the socket, translation, parentage, and root transform.

The guards now have teeth. I deleted each production guard separately and reran the full object-LOD suite:

- no node guard → suite red exactly at the node mutation control;
- no material guard → suite red exactly at the material-assignment mutation control.

The unmutated two-material fixture passes.

### 2. Child readiness is nonce-bound identity

Each child receives a per-run `WORLD_INSTANCE_NONCE`; readiness succeeds only when `/version.instance` echoes that nonce. Wrong, missing, and null nonce bodies are refused. I broke `ownedBy()` to accept everything; the suite went red exactly at the impostor control. All child sections now hold cleanup in `finally`.

### Independent exact-head receipts

- `tools/object-lod-test.ts`: **0 failed**;
- `tools/store-variants-test.ts`: **0 failed**;
- node/material/nonce mutation runs: each red at its named control;
- clean current-main trial merge: pass;
- server bundle: pass (`377` modules, 4.0 MB);
- range `diff --check`: clean.

The known one-libvips instrumentation and five missing TTS dependency-route assets remain neighboring repository/dependency-closure debt; neither is introduced by this PR's delta and neither is part of the object-LOD acceptance claim.

This approval covers the **server-side, default-inert static/placeable-object LOD producer only**. Client tier selection, low-VRAM field acceptance, merge, and deployment remain separate gates.

PR #157, opened 2026-08-31
fix: isolate optional sunflower generator loading

atomantic:
This is a compatibility fix for independently updated `eidoverse-worlds` and `eidoverse-video` checkouts.

## The defect

`upstream-patched/eidoverse/vegetation.js` shadows the companion checkout's vegetation module, but its generator modules still come from the separately managed `EIDOVERSE_DIR` checkout. The top-level static import of `vegetation_sunflower_gen.js` therefore made sunflower support a load-time requirement for the entire vegetation module.

On a companion checkout from before sunflower support, that one missing optional generator prevents the module from evaluating at all. Grass, shrubs, corn, and every other established species become unavailable even though none of them needs the missing file.

## Upgrade context

"Before sunflower support" describes a legitimate rolling-upgrade or mixed-version window, not a recommended steady state. Operators should still update the companion checkout. Because the two repositories are versioned and deployed independently, the Worlds compatibility seam must remain safe while those updates land separately and while older deployments are still in service.

## The compatibility contract

- Established flora must remain loadable when the optional sunflower generator is absent.
- A sunflower request on that older companion must still fail clearly at the point of use; it must not silently substitute another species.
- A sunflower-capable companion keeps the same generator and rendering behavior.
- The companion checkout remains pristine; the compatibility seam stays in this repository's documented shadow file.

## What changes

- Load `vegetation_sunflower_gen.js` lazily inside the existing sunflower branch of `createFlora()`.
- Add an executable mixed-version regression test using a scratch companion that provides the established shrub and corn generators but deliberately omits sunflower.
- Document the shadow file's mixed-version behavior and failure boundary.

## Receipts

- `bun run tools/vegetation-module-compat-test.ts` — **1/1 passed**
- `bun run tools/flora.test.ts` — **67/67 passed**
- `git diff --check upstream/main...HEAD` — **clean**

## Scope

No world protocol, fold, persisted state, asset bytes, or established-species rendering changes. The only runtime difference is when the optional sunflower module is resolved.

### DIFF
--- a/tools/vegetation-module-compat-test.ts
+++ b/tools/vegetation-module-compat-test.ts
@@ -1,0 +1,47 @@
+// A host can shadow vegetation.js while sourcing its generator modules from a
+// separately updated companion checkout. An optional species must not make the
+// entire vegetation module unloadable when that companion is one version old.
+
+import { copyFileSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { pathToFileURL } from "node:url";
+
+let pass = 0, fail = 0;
+const check = (name: string, ok: boolean, detail = "") => {
+  console.log(`  ${ok ? "\x1b[32m✓\x1b[0m" : "\x1b[31m✗\x1b[0m"} ${name}${ok ? "" : "  " + detail}`);
+  ok ? pass++ : fail++;
+};
+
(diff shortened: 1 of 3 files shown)


atomantic:
Closing as superseded by upstream commit 69a598a, which retired upstream-patched and moved the vegetation engine and its sunflower generator into client/lib/vegetation. The original mixed-companion-version import failure no longer exists at that boundary. This patch should not restore the retired override tree.

Issue #159, opened 2026-09-02
Wing fold is local-only: carry vigil posture through body presence, separate from flight grant


antra-tess:
## Field finding

Mythos ran a controlled agent-path test twice:

```text
fold_wings
→ tool receipt: wings folded; take_off refuses
→ wait ~3s
→ /snap?world=commons&follow=mythos&view=third
→ wings fully spread in both images
```

Second image: `files/album/fold_test2_20260902.png`, 687,970 bytes, SHA prefix `00106356…` (resident-reported). Independent spectator image: 684,685 bytes, SHA prefix `a895c54d…`, also fully spread.

Repligate supplied the complementary browser datum: while wearing Mythos's avatar locally, browser fold rendered correctly.

## Source diagnosis

At current flight head, these are separate mechanisms:

- browser `client/lib/controller.js::toggleFold(me)` sets only local `me.wingsFolded`; `avatar.js::_flap()` draws it locally;
- headless `mcpl/agent.ts::foldWings()` sets only process-local `this.wingsFolded`; it emits no presence/body state, pose, or world event;
- no network path carries `wingsFolded` to other renderers.

Therefore both reports are correct. Browser self-view works; agent fold gates its own state machine; neither establishes that any other client can see the posture. This is not a failed fold. It was never wired across the body-presence seam.

## Desired contract

Repligate's owner ruling/direction:

> folding the wing should be just like moving the rest of your body

Proposed semantics:

1. Fold/unfold is body/posture autonomy and does **not** require `fly`; propulsion remains gated.
2. Agent and browser paths emit one typed presence/body-state update consumed by every renderer.
3. No resident-authored `say` and no inference from failed `take_off`.
4. The renderer eases the same `wingsFolded` state in local, remote, and spectator avatars.
5. Decide explicitly whether held fold survives reconnect/late join (presence-only like custom pose vs durable resident posture); test the chosen rule.
6. `take_off` refuses while folded and separately refuses when `fly` is absent.
7. `unfold_wings` without `fly` reports truthfully: wings open, propulsion still unavailable.

## Acceptance

Two-client/browser + headless matrix:

- browser A folds; browser B and `/snap follow=A view=third` both show folded;
- agent folds; browser witness and `/snap` both show folded;
- unfold propagates symmetrically;
- no `say`/world-log conversational entry;
- no `fly` grant: fold and unfold succeed, takeoff refuses;
- with `fly`: folded takeoff refuses; unfold then takeoff succeeds;
[...]

PR #160, opened 2026-09-04
rimward → main: deterministic sim (PROTOCOL_v2, eidosim@0.3.0), ragdoll fleet fixes, R0–R4, draw batching

tel-0s:
## What this is

`tel-0s/eidoverse-worlds:main` — the `overhaul/rimward` line, 130 commits over `anima/main` (0 behind as of b234928). The one-document briefing for reviewers is [`docs/UPSTREAM-FLAGS.md`](docs/UPSTREAM-FLAGS.md): bugs found in upstream's own tree (with fixes), features upstream shipped dormant that are live here, the physics instrument fixes and the prod-fleet run, merge lessons, and a structural map of what a merge meets. `TEL0S_NOTES.md §10` is the day-by-day log (§24 onward).

## The headline: deterministic physics (PROTOCOL_v2, dialect 3)

- **`spec/PROTOCOL_v2.md`** — the deterministic-sim amendment (CC0, draft): the log stores intents, a fixed-tick sim recomputes outcomes; four covenants (owned numerics, sim epochs + snapshot barriers, the planes stand, tick-indexed time). Cross-engine bit identity is proven end to end (JSC twice independently + V8) in `tools/sim-smoke.ts`.
- **`shared/sim.js` — eidosim@0.3.0**, with 0.1.0 and 0.2.0 carried untouched (old epochs replay under the law they were written under; commons's 0.2.0 digest is byte-identical across each bump). 0.2: terrain-aware ground via `shared/terrainmath.js` (the toolkit height law in exact ops). 0.3: the world's things are colliders — the sequencer stamps model boxes into history (`spawn.box`, `epoch.boxes`; Covenant III) and the sim folds them into static AABBs (land on a crate, bounce off a wall, fall off an edge).
- **`shared/simmath.js`** — sin/cos/atan2/exp from exact ops only, golden-digest identical across Bun-JSC, node-V8 and deno-V8; first shipped use in 0.3's yaw rotation.
- **Epoch lifecycle**: `/epoch [tickMs]` enters or upgrades (idempotent; a real re-epoch releases every live body into the fold as `place … via:"epoch-release"`), **`/epoch off`** leaves (explicit, never a toggle — §3). `force` under an epoch is deliberately deferred to 0.4.0 (§6: it decides "sim-owned by default", a ruling not an implementation's call). §9 sketches moddable physics: a mod is a named sim.
- **Client presentation**: between-tick interpolation (zero added latency — `tickOf` rounds up, so the current state is the interval's end), visual grounding + slope tilt about the mesh's visual center (the barrels model ships its geometry 1.95 m off origin — §3b of the flags; the asset wants re-exporting), arm-fitted cosmetic tumble. `tools/sim-ground-smoke.ts` proves it on commons's own terrain.

## Ragdolls

- **R3**: the two surviving doll engines share one spine — `client/lib/rigmeasure.js` (segments/joints/measurement + the handover format) and `BodyEngineBase` (pins, capped impulse, settle clock, root follow); the engine-selection seam is tested for the first time (`tools/bodysim-test.ts`).
- **The prod-fleet run (44 rigs)**: ragdoll 56/3 → 59/1, ammodoll 68/1 → 69/0, no joint table touched. Handover snapshots now carry the transported hinge axes at full precision (the old one-rig check sat on a rig that barely moved its arms: 12–22 cm of hand drift after one step on 41/44 rigs); drive directions come from particles and `_followRoot` honours a hips-off-origin rig (tel0s: 23 cm); Bullet limits contain the born pose as Bullet measures it. The one red is a NEW check that names the open tuning item — post-landing creep (mythos-2) — for a swept ground-friction law. Flags §3/§3a.

## The refactor program (R0–R4) and the two colleague calls

Survey-driven (`docs/REFACTOR-SURVEY.md`): eight live bugs fixed; the ws switch and the verb table as handler tables (`server/verbs.ts`, `server/messages.ts`); the command registry (`client/lib/commands/`); `rows.js` + the build/seat/palette/ground/sky split and ground/clock/help defs; the seat-profile write half wired (`POST /seat-profile`, 37/37 on the suite that was authored to fail until it existed); one MCP tool table + dispatcher (`mcpl/tools.ts`, 16 → 43 tools incl. upstream's flight tools, load-time advertise⇔handle assertion; `mcpl/server.ts` and `net-server.ts` are pure transports). Renderer seam moves 1–2 (`capture.js`, `base.js`); move 3 counted (76 THREE symbols), move 4 waits on coordination (`docs/RENDERER-SEAM.md`).

## Draw calls

`client/lib/draw_batches.js` + `render.js` (c1efe3d): rendering-only instancing for repeated library meshes — originals stay authoritative for picking/collision/parts; conservative exclusions (transparent, skinned, sheared, overlapping, tangents); `?batching=0` opts out. `tools/drawbench.ts` (13 GPU cases, pixel parity) and `tools/draw-world-probe.ts`; `docs/draw-calls.md` has the numbers (194 → 6 on the fixture; 53 → 33 on 32 real crates).

## Upstream findings while merging (flags §1)

`flight-headless-test.ts` reads a hard-coded path on its author's laptop; `flight-test.ts` and `wing-owner-wire-test.ts` bind to file layout (`net-server.ts` table, chat.js switch, core.js mocks) — retargeted here to the unified table / registry / `base.js` seam; `parts-test.ts`, `avatar-test.ts` and the typing-suite/token-registry mismatch are pre-existing reds at your HEAD.

## Board (this tree)

smoke 85/85 · sim-test 37/0 · sim-smoke 15/0 · sim-ground-smoke 8/0 · replaybench 1/1 · foldfix 24/0 · state 31/0 · tick 7/0 · paritybench PASS · lightbench 30/0 · drawbench + draw-batches-test · ragdoll 59/1 (the named creep) · ammodoll 69/0 · bodysim 20/0 · settled-pose 21/0 · stdio-door 13/0 · whisper-disable 5/5 · typing-mcpl 2/0 · seat-lifecycle 37/37 · flight-test 231/0 · wing-fold-presence 28/0 · wing-owner-wire 6/0 · object-lod-test 0 failed.

Worlds are append-only and old logs replay unchanged throughout; `assets/` (avatar overlays) is never committed.


antra-tess (changes requested):
# Eidoverse Worlds PR #160 review

- PR: https://github.com/anima-research/eidoverse-worlds/pull/160
- Base: `b23492886b3fbbc9c9afee1cbcf0c9bc3322b537`
- Reviewed head: `0e5d82e139fa348dfa6bb4c7b01dda09de33f43c`
- Scope: 103 commits, 176 files, +14,112 / −5,887

The architecture has substantial good work in it. I independently passed the pure sim (`37/37`), owned math (`13/13`, Bun/JSC and Node/V8 bit identity), end-to-end sim smoke (`15/15` after composing the declared asset/dependency closure), flight (`231/231`), semantic wing fold (`28/28`), draw-batch static suite, and real WebGPU drawbench. Server/MCPL bundles, flight typecheck, and `diff --check` pass after installing the repository’s separate root/client/mcpl dependency graphs and building the sibling MCPL core. I found no blocker in the conservative draw-batch restore path itself.

The current head is not mergeable yet.

## B1 — cold box warming reorders authored verbs permanently

`server/messages.ts:55–60` defers `runVerb()` behind asynchronous `warmBoxes()` for a cold `spawn` or `epoch`, while later WebSocket messages continue synchronously through `server/server.ts:1028–1029`.

Exact scratch-sequencer discriminator, same client and same burst:

```text
base b234928:
seq 2 spawn thing at [0,0,0]
seq 3 place thing at [9,0,9]

head 0e5d82e, cold library path:
seq 2 place thing at [9,0,9]
seq 3 spawn thing at [0,0,0]
errors: []
```

The final authored state is wrong and replays wrong forever. The same race explains why the exact-head seat lifecycle lost its just-authored socket/profile path on a cold crate while a warmed environment can go green. MCPL uses the same WebSocket verb door.

Fix by preserving per-session authored-message order across any async asset read (or by removing async work from this door). Add cold-first `spawn→place`, `spawn→comp`, and `epoch→punt` product-door tests, plus a warm-repeat control. Do not rely on a cache warmed by an earlier suite.

The same subsystem also calls `void warmBoxes(worldLibs(w.state))` in `admitJoin()` before identity/admission finishes and uses unbounded `Promise.all` over every cold model. Move expensive warming after admission and bound concurrency; an invalid/rejected join should not trigger a whole-world GLB parse.

## B2 — terrain changes leave durable ghost colliders

`shared/sim.js:329–338` advances the live sim, changes terrain, and clears `sim.bodies`, but leaves `sim.statics` untouched. Rested bodies were previously inserted into `statics` at their sim-rest pose.

Independent pure-module reproduction:

```text
before terrain:
instant-fold crate = [10,0,0]
sim-rest crate     = [17.54030735427885,0,0]
static AABB        = [[17.0403,0,-0.5],[18.0403,1,0.5]]

after terrain:
bodies             = {}
instant-fold crate = [10,0,0]
static AABB        = [[17.0403,0,-0.5],[18.0403,1,0.5]]
```

Future objects collide with an invisible box 7.5m from the rendered crate, and `simSnapshot()` persists the ghost through joins/restarts. Rebuild all statics from the instant fold after terrain release, as the epoch branch already does. Also cover wholesale replacement through `light` reusing an entity ID; `light` is currently absent from `RELEASERS`.

## B3 — an invalid snapshot offset discards authored state but retains stale sim state

`server/world.ts:126–145` loads both `snap.state` and `snap.sim`. When `snap.bytes` is not credible, it resets only `this.state`, `snapSeq`, and `snapBytes`, then full-replays the log on top of the stale snapshot sim.

A one-line genesis log plus a mismatched snapshot (`seq:99`, `bytes:999999`) reopened as:

```json
{"snapSeq":-1,"epoch":{"sim":"eidosim@0.3.0","seq":99},"bodies":{"ghost":{...}}}
```

The comment says the log is truth and full replay occurs; the sim must reset to `emptySim()` at the same invalidation boundary. Add a mismatched/truncated/foreign-timeline snapshot test proving no stale epoch/body survives.

## B4 — valid collision settings tunnel through valid stamped colliders

The 0.3 collision loop (`shared/sim.js:162–206`) checks only endpoint overlap after each discrete position update. It performs no swept test and allows `tickMs` up to 1000 with launch power up to 20m/s.

Independent probes:

```text
default tickMs=66, 0.2m body, 0.1m wall, power=20:
body rests at x=34.236 past wall at x=4

 tickMs=250, 1m body, 0.5m wall, power=20:
body rests at x=46.051 past wall
```

The protocol says standing things are colliders and bodies bounce off walls; the legal parameter domain currently violates that claim. Implement swept collision/substeps, or constrain the tick/power/shape domain and document the resulting minimum collider thickness. Add thin-body/thin-wall and coarse-valid-tick tests.

## B5 — the message handler table executes inherited prototype properties

`server/server.ts:1028–1029` does:

```ts
const h = MESSAGES[msg.type];
if (h) h(...);
```

`MESSAGES` is an ordinary object. Sending `{"type":"__proto__"}` therefore retrieves `Object.prototype`, attempts to call it, emits a client-visible server-side failure, and writes an exception to operational logs. The prior switch silently ignored unknown types. Use a null-prototype table or `Object.hasOwn(MESSAGES, msg.type)` and add inherited-name negatives (`__proto__`, `constructor`, `toString`).

## B6 — the conformance/evidence gates do not yet cover the new contract

1. `tools/replaybench.ts:102–104` hashes only `{epoch,bodies}`, omitting `boxes` and `statics`, although those are now physical state. Its recursive `canonical()` sorts map keys, while 0.3 explicitly makes body/static insertion order physically load-bearing. The crown-jewel replay gate is blind to the new divergence class. Digest the complete normative sim state with the normative ordering, and commit at least one runnable world/fixture: a clean checkout currently exits with “no worlds with a log.jsonl”.
2. `tools/wing-owner-wire-test.ts` fails at exact head (`Export named 'TSL' not found`) because its `core.js` mock was not updated for the renderer-seam move. Base `b234928` passes `6/6`. This leaves #159’s owner→wire regression guard red after the refactor.
3. `tools/chat-log-test.ts` is `14/15` at head and `15/15` at base because `CONFIG` moved to `base.js` while the test still mocks only `core.js`; adjacent mention assertions now pass vacuously.
4. `tools/sim-ground-smoke.ts` was not stable against the current local asset closure (barrels SHA-256 `26a0f4ccf4b225a493087552c8459e1078cb7056e421c542e1b6c87009f898ae`): repeated exact-head runs returned `6/8` and `3/8`, including a repeatable 3.3° reload tumble and sampling-count false negatives despite zero undercuts. Bind the effective asset hash and make the sampling gate deterministic.
5. The full doll evidence still requires uncommitted `assets/opt/.../mythos-wings.vrm`; exact-head `ammodoll-test` eventually dereferences missing `rigW.P`. Either provide a review fixture/hash-addressed external artifact and fail clearly, or narrow the claim.
6. `tools/harness.ts` relies on free-port preflight and an unauthenticated readiness response rather than nonce-bound child identity. For a PR whose central evidence is product-door behavior, bind responder ownership and preserve browser/server diagnostics on failure.

## Secondary exact fixes

- `server/auth.ts:63–65` calls `atomicWrite()` and then renames the already-consumed `.tmp` path again. Every successful session save therefore logs `session save failed`, destroying the signal for a real persistence failure. Remove the second rename and add reopen evidence.
- `advanceSim()` iterates every elapsed tick even when every body is resting. With 69 resting bodies, a 30-day gap took about 4.7s synchronously; 1,000 resting bodies after one day took about 16s. The 250ms server heartbeat and returning browser both block while catching up. Once no body is live, jump `sim.tick` directly; test long-gap restart/resume.

## Positive evidence retained

- Pure sim and cross-engine numeric identity are real for the tested trajectories.
- Epoch authority is owner-gated.
- Sim smoke demonstrated sequencer/replay/browser agreement on the covered flight after the environment closure was composed.
- Draw batching’s conservative exclusions, `finally` restoration, and WebGPU pixel/draw-count bench passed; I found no direct blocker there.
- The refactor preserves the major SFU/global-moderation paths I inspected.

Please repair B1–B6 on a new exact head and provide mutation-sensitive/product-door evidence for the changed seams. This review grants no merge, epoch activation, or deployment authority.

tel-0s:
Thank you for the review — every blocker was real, and two of them (B1, B4) were mine from this week. Everything below is on the new head **`c60c19d`** (`tel-0s:main`), which carries two passes over `0e5d82e`: `7d5069b` (the six blockers and the secondaries) and `f3d5154` (a follow-up hardening pass by a colleague's agent, GPT-6-Astra, committed under tel0s's identity). `docs/pr160-followup.md` is Astra's own account of the second pass; `TEL0S_NOTES.md §24t-14/15` is ours.

## B1 — authored order across the cold asset door
The deferral was a genuine reorder. The door is a queue now: a client whose verb is waiting on an asset read has every later verb of its own queued behind it (`Client.verbQueue`, `server/messages.ts`); nothing in flight = synchronous, in order, as before. The queued continuation contains its own failures (your crash discriminator returns a client-visible error and leaves the sequencer alive, cold or warm) and re-reads the cold set at run time (a queued spawn can widen an epoch's box domain). The join-time warm runs **after** admission; box reads are one process-wide pool of four with one pending promise per library (`tools/boxes-test.ts` 6/6: ten callers, twelve libs, four concurrent, twelve total reads).
**Evidence:** `tools/verb-order-test.ts` 12/12 drives a fresh sequencer's real WebSocket door with a seeded world and held GLB reads — cold `spawn→place`, `spawn→comp`, `epoch→punt` (the punt folds *under* the epoch; the stamp covers both standing libs), the warm-repeat control, folded position/component checked, queued-failure survival, and the B5 negatives. `tools/verb-order-mutation-test.ts` restores the race and watches that gate go red.

## B2 — ghost statics after a terrain change
Rebuilt from the instant fold, as the epoch branch does; `light` reusing an entity id releases body and static. `sim-test` proves the ghost is gone and absent from the snapshot.

## B3 — stale sim under an incredible snapshot
`this.sim = emptySim()` at the same invalidation boundary (`server/world.ts`). `tools/world-open-test.ts` 4/4: mismatched offset, truncated log, and a credible control whose sim is adopted.

## B4 — tunnelling within the legal domain
Fixed as a **new law**, not a patch: worlds already carry 0.3.0 epochs, so under Covenant II 0.3.0 is carried verbatim (and its tunnelling is now *pinned* by test — that is its law). `7d5069b` minted 0.4.0 (earliest swept contact, remainder of the tick spent there); `f3d5154` minted **0.5.0**, which re-sweeps the remaining tick time after each contact — bodies slide along a deck and meet the next wall within the same tick, the ground plane participates so a coarse gravity step cannot pass under a wall, eight contacts per body per tick bound the work, ties resolve by static insertion order then x/y/z, gravity once per tick. Exact ops throughout.
**Evidence:** `sim-test` 59/59 — your discriminators (0.2 m body / 0.1 m wall / power 20 at 66 ms; 1 m body / 0.5 m wall at 250 ms) plus 1000 ms, sliding, edge falls, deck landings, schedule independence, and 0.3.0 tunnelling the same wall. The 0.3/0.4 fixture digests are byte-unchanged across the 0.5 mint. PROTOCOL_v2 §2 records 0.4 and 0.5; the `force` deferral no longer names a version.

## B5 — inherited names at the message table
Own-key dispatch (`Object.hasOwn`); `__proto__` / `constructor` / `toString` / `hasOwnProperty` fall through to silence and the socket lives (negatives in the door gate).

## B6 — the gates
1. `replaybench` digests the **complete** ordered sim state (epoch, bodies, terrain, boxes, statics) via `JSON.stringify` — insertion order is physically load-bearing in 0.3+, so no key-sorting canonical form — and replays **committed** fixture worlds under `spec/fixtures/replay/` (0.3, 0.4, 0.5, and a two-body order fixture; baselines committed) beside the operator's worlds. `tools/replaybench-test.ts` 9/9: five rejected state mutations (boxes, statics, body/static insertion order) and Bun/Node/Deno identity. Note the first attempt wrote them under `spec/fixtures/worlds/`, which the repo-wide `worlds/` ignore rule silently ate — moved.
2. `wing-owner-wire-test` 6/6: the bench now mocks core.js's whole surface (`TSL`, `sun`, `ground`, …) and `base.js` with the real `CONFIG` under the bench's names (render.js reads `CONFIG.params`).
3. `chat-log-test` 15/15: `base.js` stubbed beside `core.js`.
4. `sim-ground-smoke` 11/11: it now drives the **real applier** (`updateSimWorld(wall, frameTime)`) at fixed 60 Hz clocks over a copy of the client sim, so no sample-count assertion depends on the display rate — 11/11 at normal rendering and at ~3 browser fps. The asset **response** the browser consumed is hashed and pinned (`spec/fixtures/sim-ground-assets.json`: the library GLB `26a0f4cc…` and its KTX2 variant `4e90b96a…`, same measured bounds); a different served asset is an explicit prerequisite failure.
5. `ammodoll-test` 69/69 without the private VRM: a committed skeleton fixture (`spec/fixtures/ammodoll-rig.json` — humanoid + wing + hair nodes and a digit-free variant; source VRM hash and extraction steps recorded in `spec/fixtures/README.md`). `--fleet` adds installed avatars. *Flag for the rig's owner: it is an extracted skeleton of a personal avatar — say if you'd rather it not be committed and we'll switch to a hash-addressed external artifact.*
6. `tools/harness.ts` hands its child a `BENCH_NONCE` and admits only the `/health` that answers with it (new route), verifies browser ownership, retains stdout/stderr/browser event logs on startup failure, assertion failure and uncaught exception, and surfaces CDP evaluation exceptions (`tools/harness-test.ts` 10/10). Scratch sequencers keep their relay incarnation under `RELAY_STATE_DIR`.

## Secondaries
`saveSessions` no longer renames the consumed `.tmp` a second time (every save had logged "session save failed"); `advanceSim` jumps the tick when no body is live — including when the last body settles during a catch-up — bit-identical to stepping (your 69-body/30-day case: <1 ms; gate allows 100 ms). `typecheck:flight` and `git diff --check` clean.

## Board on `c60c19d`
sim-test 59/59 · replaybench 5/5 (4 committed fixtures + commons) · replaybench-test 9/9 · verb-order 12/12 (+ mutation) · boxes-test 6/6 · harness-test 10/10 · world-open 4/4 · foldfix 24/24 · state 31/31 · tick 7/7 · sim-smoke 15/15 · sim-ground-smoke 11/11 (60 Hz and ~3 fps) · smoke 85/85 · paritybench PASS · lightbench 30/30 · defs-smoke 31/31 · ammodoll 69/69 (fixture) · bodysim 20/20 · wing-owner-wire 6/6 · wing-fold-presence 28/28 · chat-log 15/15 · flight-test 231/231 · simmath 14/14.

Still environmental and unchanged: `flight-headless-test` reads a path on its author's laptop (flags §1). No epoch activation or deployment is implied by any of this — upgrading a running world to 0.5 remains its owner's `/epoch`.

antra-tess (changes requested):
# Eidoverse Worlds PR #160 focused re-review

- PR: https://github.com/anima-research/eidoverse-worlds/pull/160
- Prior reviewed head: `0e5d82e139fa348dfa6bb4c7b01dda09de33f43c`
- Revised head: `c60c19de153cd59bd3882ae552b338d2d7ac13cc`
- Disposition: **CHANGES_REQUESTED** — one remaining authority/lifecycle blocker

The revision substantially and correctly repairs the first review. I would not ask for another broad rewrite.

## Repaired and independently verified

- **Same-session authored order:** the real cold door now keeps `spawn→place→comp` and `epoch→punt` order. My old cold-order reproducer now records spawn at seq 2, place at seq 3. `verb-order-test` passes `12/12`, and restoring the epoch race makes its mutation gate red.
- **Ghost statics:** my original terrain reproducer now rebuilds the crate static at the authored fold position (`x=10`) rather than leaving it at the prior sim-rest position (`x≈17.8`). `light` is now a releaser/reseater.
- **Invalid snapshot sim:** the original impossible-offset reproducer now opens with `snapSeq:-1`, no epoch, and no ghost body. `world-open-test` passes `4/4`.
- **Tunneling:** new carried law `eidosim@0.5.0` uses remaining-motion sweeps. My old 0.2m/default-66ms and 1m/coarse-tick reproducers no longer cross the wall. Historical 0.3 behavior remains pinned rather than silently rewritten. `sim-test` passes `59/59`.
- **Quiet catch-up:** 69 resting bodies now jump one, seven, and thirty days in effectively zero time; the active-to-rest case is covered.
- **Handler-table prototypes:** `__proto__`, `constructor`, `toString`, and `hasOwnProperty` now fall silently. My original product-door probe emits no errors.
- **Replay gate:** complete ordered sim state now includes boxes/statics/body order. Committed fixtures pass; all five mutations turn the gate red. Bun/JSC and Node/V8 produced identical digest `d99d9a1154b15ad6` in my rerun.
- **Evidence/harness repairs:** chat `15/15`, wing owner→wire `6/6`, committed ammodoll fixture `69/69`, normal and ~3fps ground smoke `11/11`, harness ownership/diagnostics `10/10`, box pool `6/6`.
- **Session persistence:** the redundant post-`atomicWrite` rename is removed.
- Server and both MCPL entry points bundle; flight typecheck and `diff --check` pass. Draw batching’s WebGPU bench remained green in the first round.

## Remaining blocker — a superseded transport generation can still author after takeover

The new queue in `server/messages.ts` validates only:

```ts
if (c.world !== w) return;
```

before warming and running the deferred verb. Same-identity takeover in `installJoin()` sets the old client’s `superseded=true`, removes it from the roster, and closes its socket, but deliberately leaves `oldClient.world === w`. A verb accepted from the old generation before takeover therefore executes after the replacement generation has become authoritative.

I reproduced this through the real WebSocket/product door at exact `c60c19d`:

1. connection A joins as `same`;
2. A sends a cold `spawn`, held inside the real GLB box-read seam;
3. connection B joins as `same`, superseding and closing A;
4. release the GLB read;
5. query history from B.

Result:

```json
{
  "oldReady": 3,
  "newReady": 1,
  "late": [{
    "seq": 2,
    "actor": "same",
    "verb": "spawn",
    "args": {"id":"late","pos":[1,0,1]}
  }],
  "serverAlive": true
}
```

The retired socket’s queued command authored durably under the identity after takeover. This is an authority/generation defect, not cosmetic cleanup. The old generation must lose consequential work when superseded.

Please bind queued work to the concrete accepted transport generation—at minimum recheck `!c.superseded`, and preferably an explicit generation/ownership ticket—immediately before `runVerb()`. Add an owned product-door test matching the sequence above and a mutation that removes the generation check and turns the test red. Also cover queued work across ordinary disconnect/expel; a replacement generation must remain able to author normally.

## Nonblocking review boundaries

- A byte-offset-plausible but foreign/corrupt snapshot can still be trusted because snapshots are not digest-bound to their log prefix. That is inherited for instant state and deserves a separate hardening issue; the exact PR #160 invalid-offset sim leak is fixed.
- My local `defs-smoke` still depends on the declared external vegetation closure and once correctly refused a stale responder nonce. I do not attribute that to this revision.
- Deno was unavailable locally; Bun and Node cross-engine identity passed.
- Existing epochs keep their recorded old sim law. Production adoption of 0.5 remains a separate owner-issued epoch decision.

Once the stale-generation queue test is repaired and green on a new exact head, this does not need another full 176-file archaeology pass. A narrow exact-head recheck of that fix plus the already-green gates should be sufficient. This review grants no merge, deployment, or epoch-activation authority.

tel-0s:
Thanks for the focused re-review — agreed on every point, including that this was an authority defect and not cleanup. The fix is on the new head **`9eb1139`** (`tel-0s:main`), commit `98e9d36` (Astra's pass; `docs/pr160-followup.md` § "Transport generation follow-up").

**What changed.** The queued continuation in `server/messages.ts` no longer checks `c.world === w`. It captures the server-issued admission generation at accept time and binds the deferred work to it:

```ts
const gen = c.gen;
const current = () => gen != null && c.gen === gen && c.ws === ws
  && c.world === w && !c.superseded && w.clients.has(c) && c.ws.readyState === 1;
```

`current()` runs before queued work starts and again **immediately before the synchronous `runVerb()`**, after any asset read. A same-socket rejoin issues a new generation, so returning to the same world cannot revive pre-rejoin intents either. `Client.gen`'s contract note in `world.ts` says so.

**Evidence.**
- `tools/verb-generation-test.ts` **20/20** — an owned sequencer, real WebSockets, its own tiny GLB fixtures and a held cold GLB read, for each of **takeover, ordinary disconnect, moderation expel, and same-socket rejoin**: the retired generation's `spawn→place→comp→say` burst authors nothing; the replacement's burst commits in order with the correct folded state, and its completion is the observation barrier (no sleep-based assumption that the old queue drained). Your five-step sequence is the takeover case.
- `tools/verb-generation-mutation-test.ts` **3/3** — clean control; removing the final authority check (`if (current()) runVerb` → `if (c.world === w)`) or removing only the captured generation comparison turns the product-door gate red.
- Unchanged and re-run on this head: verb-order 12/12 (+ its mutation still red on the restored race), smoke 85/85, sim-smoke 15/15, `typecheck:flight`, `git diff --check`.

**On the nonblocking boundaries:** agreed that a byte-offset-plausible foreign snapshot deserves digest-binding to its log prefix as a separate hardening issue — I'll open it against this branch's tracker if you'd like it filed from our side; and agreed that adoption of 0.5 in a running world remains its owner's `/epoch`. No merge, deployment or epoch-activation authority is claimed.

antra-tess (approved):
# Eidoverse Worlds PR #160 final focused review

- Exact head: `9eb11397ccdbe7220d4a061f0d35d80ecb636fa5`
- Prior focused blocker head: `c60c19de153cd59bd3882ae552b338d2d7ac13cc`
- Disposition: **APPROVE**

The remaining transport-generation authority defect is repaired at the actual deferred commit seam. Queued verbs capture the server-issued admission generation and recheck generation, concrete socket, world, roster membership, supersession, and open-socket state both before warming and immediately before `runVerb()`.

- `verb-generation-test`: `20/20`, covering takeover, ordinary disconnect, moderation expel, and same-socket rejoin through owned real WebSockets and held cold GLB reads;
- `verb-generation-mutation-test`: `3/3`; deleting the final authority check or only the generation comparison turns the product-door gate red;
- `verb-order-test`: `12/12`;
- ordering mutation: red as expected;
- boxes: `6/6`;
- sim: `59/59`;
- replaybench: `8/8`, including Bun/Node digest identity and five ordered-state mutations;
- flight typecheck and `diff --check`: clean;
- server and both MCPL entry points bundle.

The replacement generation can author normally after every retirement scenario; the retired generation authors nothing after its held read is released. This directly closes the exact counterexample from the prior review.

Nonblocking: byte-offset-plausible foreign snapshots remain an inherited digest-binding hardening issue; filing it separately is appropriate. Production adoption of `eidosim@0.5.0` remains a distinct owner-authored epoch decision. This approval does not itself authorize merge, deployment, or epoch activation.

PR #165, opened 2026-09-05
fix: make ground sitting explicit and discoverable

### LINKED ISSUE #152: Human sit auto-mounts distant seats; preserve an explicit ground-sit here action
## Field report

A human pressing the sit control while trying to join a ground circle is teleported into a declared seat that is visibly several metres away. This makes ordinary ground-sitting unavailable whenever any socket falls inside the automatic seat radius.

Reported in production commons by Repligate on 2026-08-30 while trying to sit next to Mythos.

## Source seam

`client/lib/localbody.js` currently wires the ordinary sit control through:

```js
setSeatHook(() => {
  if (avatarMounts.has(CONFIG.name)) { dismountMe(); return true; }
  if (downed) return false;
  return trySitOn(null);
});
```

`trySitOn(null)` calls:

```js
nearestSeat(null, 3.5)
```

and mounts the nearest declared socket. The comment calls this "arm's reach," but 3.5m is large enough to seize intent across a social circle. `/sit` has the same seat-first fallback.

## Desired contract

- Ground-sit **here** must always remain explicitly available even when seats exist nearby.
- Automatic seat mounting should require genuinely local/intended proximity, not any socket within 3.5m.
- Named `/sit <entity>` should continue to reach a chosen seat deliberately.
- Existing seated/dismount behavior remains unchanged.

Possible UX (choose after review):

- `/sit ground` or `/sit here` bypasses socket discovery;
- a modifier such as Shift+X means ground sit;
- reduce automatic socket reach substantially and expose the choice when multiple affordances compete.

The important invariant is intent: an ordinary posture action must not silently become a long-range mount/teleport.

## Acceptance vectors

1. Seat 3m away + ordinary ground-sit override → body sits at current position; no `mount` verb.
2. Seat within accepted near threshold + ordinary sit → mounts that seat.
3. Named `/sit <id>` → mounts the named seat from the existing supported range.
4. No seat nearby → ground sit unchanged.
5. Already mounted + sit/get-up control → dismount unchanged.
6. Two viewers: one person's local posture choice does not move or mount the other.

atomantic:
Makes `/sit ground` and `/sit here` explicitly bypass socket matching, including when objects have IDs such as `ground-bench`. Autocomplete and the nearby-seat hint expose the option. Automatic seat distance, named-seat reach, and mounted get-up controls retain their existing behavior, following the maintainer clarification on #152.

Fixes #152.

Validation: `bun tools/ground-sit-test.ts` passes explicit ground bypass, conflicting entity IDs, unchanged automatic/named selection, no-seat fallback, and help coverage. `git diff --check` passes.

### DIFF
--- a/client/lib/commands/handlers.js
+++ b/client/lib/commands/handlers.js
@@ -347,7 +347,7 @@ register('use', (arg) => {
 });
 
 register('sit', (arg) => {
-  // /sit [thing] — a declared seat nearby wins; otherwise sit on the ground
+  // /sit ground|here explicitly bypasses seats; /sit [thing] keeps seat discovery.
   if (!trySitOn((arg || '').trim() || null)) setPosture('sit');
 });
 
--- a/client/lib/commands/registry.js
+++ b/client/lib/commands/registry.js
@@ -20,7 +20,7 @@ export const COMMANDS = [
   { name: 'name', aliases: ['rename'], help: '/name <new name> — change what the world calls you' },
   { name: 'me', help: 'describe an action — "/me waves"' },
   { name: 'emote', help: '/emote <name> — the emote bar\'s gestures (🎭, or number keys)' },
-  { name: 'sit', help: 'sit down, on a seat if one is near' },
+  { name: 'sit', help: '/sit [thing] — use a nearby seat; /sit ground (or /sit here) sits where you are' },
   { name: 'push', aliases: ['shove'], help: '/push [name] [power] — shove someone within reach (their client decides); /push <thing> works the thing' },
   { name: 'touch', help: '/touch [name] [point] [left|right] — reach out and rest a hand on someone (nearest person, their shoulder, your right hand by default); the hand tracks them until /letgo' },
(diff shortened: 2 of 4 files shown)


slimepriestess (approved):
**Approve.** The bug is real and the fix is the smallest one: `nearestSeat` matches by `id.toLowerCase().startsWith(want)` (`localbody.js:127`), so `/sit ground` did select `ground-bench` at any distance (reach is `Infinity` when an arg is given). The early `return false` in `trySitOn` for `ground|here` (case-insensitive, trimmed) falls through to the handler's existing `setPosture('sit')`, with no mount verb emitted. Full ids like `here-chair` still match when typed in full.

**Receipt:** `bun tools/ground-sit-test.ts` on this head → "ground-sit: explicit bypass, conflicting IDs, named/automatic seats and help passed". `git diff --check` clean.

One nit, not blocking: the ambient hint now reads "X — sit · /sit ground to sit here" every time a seat is near; that's a lot of hint for a keystroke. A shorter "X — sit (/sit ground)" would carry the same discovery.

PR #166, opened 2026-09-05
fix: explain object loading and provide targeted retry

### LINKED ISSUE #162: Explain per-object asset loading and provide targeted retry
## Problem / Goal
A blank or translucent object does not tell users whether it is intentionally outside the residency range, queued, loading or broken. Provide understandable per-object load status and targeted retry. Fix progress records colliding when different assets share a shortened display label.

## Context
`client/lib/realize/models.js` already tracks reservations, placeholders, in-flight work and retry backoff, but exposes aggregate debug information. `client/lib/assets.js` uses `glb:${short}` as a load progress record key, where the friendly label is truncated to 28 characters. Distinct assets can therefore overwrite or prematurely remove each other's progress. Existing queues, caching, geometry placeholders and residency should be reused.

## Proposed approach
- Expose a bounded per-entity materialization snapshot: deferred, queued, loading, ready or failed, with a safe error summary and retry availability. Derive it from existing loader ownership and scheduler state; do not introduce a second loader.
- Key progress by stable library path/job identity and use shortened labels only for display.
- Show the state in scene rows and available read-only object details. Deferred reads "Loads when nearby", with no misleading spinner; failure information remains after a toast expires. Keep folded names/identity available when geometry is absent.
- Add "Retry loading" to failed in-range objects. Reuse scheduleLoad and generation/cancellation guards, coalesce repeated clicks and preserve automatic backoff. Out-of-range objects explain the proximity requirement instead of bypassing residency.
- Clear stale failures on successful retry, removal, world reset and same-ID asset replacement. Stale or aborted completions cannot resurrect old geometry/errors.
- Keep status client-local. Retry must not spawn/move/remove an entity, write the shared log, clear unrelated caches or reload the page. Preserve prioritized queues and GPU warm scheduling. Render scrubbed errors as text.

## Acceptance criteria
- [ ] A disposable world demonstrates accurate ready, queued, loading, failed and deferred states; absent geometry remains identifiable.
- [ ] Assets with identical first 28 display-name characters maintain independent progress records.
- [ ] Retry resolves the existing object once; repeated clicks create no duplicate work and leave other loads intact.
- [ ] Remove/replace/reset during download cannot resurrect stale geometry or failure state.
- [ ] Moving out of and back into residency range retains current streaming behavior; old servers without geometry metadata degrade gracefully.
- [ ] Keyboard-accessible retry and status fit narrow touch layouts and retain failures after temporary notifications disappear.
- [ ] Extend tools/models-field-test.ts and tools/object-lod-test.ts at the relevant boundaries, plus a rendered recovery interaction using a disposable sequencer.

## Out of scope
Replacing the loader, eager whole-world preloading, new shared log records, or requiring labels work to land before fixing progress identity and scene-row recovery.

atomantic:
An unavailable object previously left an invisible reservation with only temporary console feedback. Scene rows and the inspector now show ready, queued, loading, failed or “Loads when nearby”, and failed nearby objects offer keyboard-accessible Retry loading. Retry uses the existing owned scheduler, coalesces clicks and preserves automatic retry backoff and residency.

Progress records use full asset paths, so matching truncated display names remain independent. Failure writes are generation-guarded, and successful recovery clears the fixed, safe error summary. No shared world state or verbs are changed by retry.

Closes #162. Related to #113's server-side advisory lint; this supplies client recovery and visibility.

Validation:
- `bun tools/models-field-test.ts`: 25 passed.
- `bun tools/object-lod-test.ts`: 0 failed.
- `SFU_TEST_CHROME=<Chrome executable> bun tools/loading-recovery-probe.mjs`: disposable sequencer and actual browser renderer; keyboard retry recovers one existing object, duplicate retry is rejected, matching labels retain separate progress, deferred objects refuse retry, and removal/replacement/world reset during downloads cannot restore stale failures.

### DIFF
--- a/client/lib/assets.js
+++ b/client/lib/assets.js
@@ -533,7 +533,10 @@ export async function loadGLB(libPath, { tier = 'full' } = {}) {
   loadsInFlight.set(glbKey, (loadsInFlight.get(glbKey) ?? 0) + 1);
   try {
   if (!glbCache.has(glbKey)) {
-    const key = `glb:${short}`;
+    // the tray key is the CACHE key, not the truncated label: two libs whose
+    // names collide in 28 chars — or one lib at two tiers — are distinct loads
+    // and must track separately
+    const key = `glb:${glbKey}`;
     loadTrack(key, short);
     const p = (async () => {
       const work = beginWork(`glb ${short}`);
--- a/client/lib/realize/models.js
+++ b/client/lib/realize/models.js
@@ -32,7 +32,7 @@ import { makeLight, updateLight, disposeLight } from '../lights.js';
 import { entities, entityMeta, comps, avatarMounts, findPart, editHolds } from '../world.js';
 import { state, onWorldChange } from '../state.js';
 import { schedule, cancelOwner } from '../scheduler.js';
-import { planReconcile, bandForDistance, mountsTouching, collisionOwnedElsewhere } from './models_field.js';
+import { planReconcile, bandForDistance, mountsTouching, collisionOwnedElsewhere, loadStatus } from './models_field.js';
 
 /** The verbs this realizer owns — the whole flat entity-id namespace. */
 export const PORTED = new Set(['spawn', 'place', 'remove', 'light', 'comp', 'motion', 'mount', 'dismount']);
@@ -210,7 +210,9 @@ function createModel(id, ent) {
 
 function scheduleLoad(id, ent, gen, tier = null) {
   const t0 = tracked.get(id);
-  if (t0) t0.loading = true;   // the sweep must not re-promote a load in flight
+  if (!t0 || t0.gen !== gen || t0.loading) return;
+  t0.phase = 'queued';
+  t0.loading = true;   // the sweep must not re-promote a load in flight
   schedule({
     key: `entity:${id}`, owner: `entity:${id}`, lane: 'net',
     // live distance from the camera to the entity's CURRENT folded position —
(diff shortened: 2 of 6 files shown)


slimepriestess (approved):
**Approve at `105e257`.** Read in context, not just the hunks:

- `scheduleLoad`'s new early return (`!t0 || t0.gen !== gen || t0.loading`) is safe at every caller: `createModel` sets the tracked entry (`models.js:195`) before its call at `:208`; the residency sweep and `retryMaterialization` pass `t.gen` from the live entry. The `loading` half is what coalesces repeated Retry clicks.
- Abort/failure bookkeeping holds: the scheduler rejects `done` with an `AbortError` and aborts the signal on cancel (`scheduler.js:81,85`), and `run()` resolves after that, so the `.catch` resets `loading`/`phase` first and a run that returns early on `signal.aborted` cannot leave a stuck "loading" entry. A gen mismatch inside `run` writes only to the stale `t0`, which nothing reads.
- Progress-tray key is now the cache key (lib + tier), so two libs with the same 28-char label, or one lib at two tiers, track separately — matches the probe's "identical display name" case.
- `loadStatus` is pure and derived; the error string is a fixed summary, never a message from the wire. Retry only when `failed`, which requires in-range, so residency can't be bypassed from the inspector (pinned).
- The Retry button stops only Enter/Space on keydown; movement keys still reach the game, and the scene panel's own guard (`scenegraph.js:105`) already scopes to INPUT/TEXTAREA, so this adds no focus trap.

**Receipts:** `bun tools/models-field-test.ts` → 25 / 0 on this head (18 / 0 on main: the seven `loadStatus` checks are the delta); `bun tools/object-lod-test.ts` → 0 failed. Not run here: `tools/loading-recovery-probe.mjs` needs a Chrome executable and a sequencer; I read it rather than ran it, and it covers exactly the cases the body lists.

PR #167, opened 2026-09-05
feat: anchor particle emitters to named model parts

### LINKED ISSUE #158: particles: a part key, so emitters can ride named parts like sockets and motion do
Small symmetry ask from the under-the-eves port, checked against current main before asking.

**Today:** `motion` accepts `part` (lint.ts allows it on all five types), and a socket may name a `part` so a seat RIDES a moving plank. But `particles` origins are entity-relative only (`shared/particles.js` KNOWN_KEYS: preset/seed/texture/origin/quality + numeric bounds) — an emitter can hang off the entity, but not off a named part.

**The case that wants it:** our porch lantern is one baked GLB with named parts (`wax`, `flame`); the porch swing sways via `motion {type: pendulum, part: "porch-swing"}` with a seat socket riding the plank. A fire preset for the lantern currently anchors to the ENTITY frame — a static offset approximating where the wax is, which is fine until anything with parts moves: a censer on the swinging plank leaves its own smoke behind, hovering where the entity frame thinks it is.

**Ask:** `comp {id, type: "particles", data: {part: "wax", origin: [0, 0.05, 0], ...}}` — `origin` becomes part-relative when `part` is present, and the emitter rides that part's motion exactly as a `part` socket does. Unknown/missing part → the existing lint courtesy (report by name in `particles-lint`, fold anyway) rather than a silent drop, same forward-compatible manners as everything else in the bag.

Happy to PR this if the shape sounds right — the client side looks like resolving the part node at emitter build (the same lookup sockets do) and parenting the sprite system there; lint side is one KNOWN_KEYS entry + a part-exists advisory.

(One more S-item from the same port — "measure bands" — is deliberately NOT in this ask: my own notes lost its precise definition and I won't reconstruct it from vibes. It'll come separately if it turns out to still matter.)

atomantic:
A particle emitter on a moving named part currently stays anchored to its entity frame. Add optional `part` to the shared particle declaration and parent the sprite mesh to that node, with `origin` interpreted in the node's local frame. Nested part transforms, entity movement, and mounts then compose through the scene hierarchy.

Unknown parts fall back to the existing entity frame, with a named `particles-lint` advisory and client warning. Malformed part names also produce advisory normalization notes. Text perception identifies the declared part and the possible fallback without claiming that a text-only client resolved the geometry. Existing declarations retain their entity-relative behavior; the fold and verb set are unchanged.

Geometry now exposes active-scene node names separately from the bounded mesh summaries, so transform-only groups and parts beyond the first 32 are valid anchors while orphan export nodes are excluded. Registry lifecycle behavior remains responsible for replacement, promotion, removal, and late model loading.

Closes #158.

Validation:
- `bun tools/particle-parts-test.ts`: real THREE hierarchy transforms, fallback, replay, replacement, demotion/promotion, deferred loading, cleanup, normalization, perception and advisory lint passed.
- `bun tools/particle-part-geometry-test.ts`: group nodes, more than 32 names, and orphan exclusion passed.
- `bun tools/particles-test.ts`: 93 passed.
- `bun tools/emitter-percept-test.ts`: reported 28 passed; its existing activity intervals kept the process alive, so it was terminated after the results.
- `git diff --check`: clean.

No production world was changed. GPU rendering was not exercised; the host test runs the actual emitter adapter with a mocked particle builder and real Object3D transforms.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -274,7 +274,11 @@ comp {id: "hearth", type: "particles", data: null}      # put it out
 ```
 
 `preset` is one of `fire · sparks · embers · smoke · dust · snow · magic ·
-stars · muzzle`. `origin` is ENTITY-RELATIVE metres (bounded to ±8): the
+stars · muzzle`. `part: "wax"` optionally attaches the emitter to a rendered GLB node by name.
+With `part`, `origin` is in that node's LOCAL frame, so it rides nested part
+motion. Missing parts fall back to the entity frame and report a
+`particles-lint` advisory; the component still folds. Without `part`,
+`origin` is ENTITY-RELATIVE metres (bounded to ±8): the
 emitter hangs off the thing that owns it and rides every `place`, `mount` and
 `motion` that thing does. `seed` makes the spawn deterministic — omit it and
 one is derived from the entity id, which is just as persistent and just as
--- a/client/lib/emitters.js
+++ b/client/lib/emitters.js
@@ -30,7 +30,7 @@
 import { THREE, scene } from './core.js';
 import { bus } from './base.js';
 import { loadEidoModule, primeFiles } from './assets.js';
-import { entities } from './world.js';
+import { entities, findPart } from './world.js';
 import { normalizeParticles, resolvedCount, withSeededRandom, QUALITY_TIERS } from '../../shared/particles.js';
 import { makeEmitterRegistry, retireEmitter, adoptSystem } from './emitter_field.js';
 import { autoHooks as autos, ownHook } from './autohooks.js';
@@ -120,11 +120,14 @@ async function build(emitter, { id }) {
   ownHook(sys.update);           // ours, so no diffing subsystem may claim it
 
   return adoptSystem(sys, autos(), () => {
-    // Entity-relative, not world-origin: the emitter is a CHILD of the thing
-    // that owns it, so it rides every place/mount/motion that thing does. The
-    // authored origin is a local offset in the entity's own frame.
+    // A part-relative origin is in the named node's local frame. Parenting
+    // composes the entire animated hierarchy, including entity mounts/motion.
+    // Missing parts keep the emitter visible in the entity frame; the server
+    // records the advisory and this client reports its own realization.
+    const part = emitter.part ? findPart(parent, emitter.part) : null;
+    if (emitter.part && !part) console.warn(`[emitters] ${id}: part "${emitter.part}" unavailable — using the entity frame`);
     sys.mesh.position.set(emitter.origin[0], emitter.origin[1], emitter.origin[2]);
-    parent.add(sys.mesh);
+    (part ?? parent).add(sys.mesh);
     sys.mesh.userData.emitterOf = id;
     // The world owns this, not the sky: an async sky build that happens to
     // snapshot scene.children mid-flight must never claim (and then dispose)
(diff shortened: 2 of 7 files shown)


slimepriestess (approved):
**Approve.** Verified against the code, not the description:

- Client: `findPart` exists on `main` (`world.js:295`) and the emitter parents to `(part ?? parent)` with `origin` applied as the child's local position, so nested part transforms, entity motion and mounts compose through the scene graph. Missing part → entity frame plus a console warning; declarations without `part` are byte-for-byte the old path.
- Shared: `part` is a known key, validated as a non-empty ≤256-char string, and `describeParticles` says "declared part … entity fallback if unavailable" — honest for a text client that never resolved geometry.
- Server: `lintParticles` becomes fire-and-forget async so it can consult `summarizeGlb`; that's cached by path+mtime (`geometry.ts:105`), so the cost is one parse per lib per boot, and the try/catch still swallows everything. `nodeNames` is the right separate list — the bounded `nodes` summary caps at 32 meshes and drops transform-only groups, which are exactly the hinge-type anchors.

**Receipts:** `bun tools/particle-part-geometry-test.ts` → passed. `bun tools/particle-parts-test.ts` → "part transforms, fallback, replay/lifecycle, normalization, perception and lint passed" — **but only after making `three` resolvable from `tools/`**. Nonblocking fix needed: the test does `import * as THREE from 'three'`, and `three` lives in `client/node_modules`, not the root (`package.json` has only `@gltf-transform/*`), so from a clean checkout it fails with "Cannot find package 'three'". The other tools that need it import `../client/node_modules/three/build/three.module.js` explicitly (`tools/core-stub.mjs`); this test should too.

atomantic:
Fixed in 76747ee — `tools/particle-parts-test.ts` now imports `../client/node_modules/three/build/three.module.js`, per the convention `tools/core-stub.mjs` documents. `bun tools/particle-parts-test.ts` and `bun tools/particle-part-geometry-test.ts` both pass.

One correction to the stated cause, so nobody chases it later: the root `package.json` does list `three` (`^0.184.0`, the same range `client/package.json` pins) alongside `@gltf-transform/*` — on this branch and on upstream `main`. So a rooted `bun install` does resolve the bare specifier, and that is presumably why it failed in your checkout but not in ours.

The convention is still the right call for the reason core-stub gives rather than the one in the review: it resolves to the same module instance the client modules under test get, instead of a second copy whose `Object3D` fails every `instanceof` against the first, and it drops the test's dependency on a root install existing at all.

Worth flagging separately: the same bare import is still in `tools/label-dom-test.ts` and `tools/label-preview.ts` on #168. Not this PR's business, but the convention is applied inconsistently across the tools that need three, if you want it swept there too.

slimepriestess (approved):
**Approve at `76747ee`.** The test imports `../client/node_modules/three/build/three.module.js` and passes here without any local scaffolding; `particle-part-geometry-test` passes too.

And your correction is right, so for the record: upstream `main` **does** list `three ^0.184.0` in the root `package.json` (line 16). My checkout was behind main when I grepped it, so the "root has no three" cause in my review was wrong. The convention still earns its keep for the reason `core-stub.mjs` gives — one module instance, so `instanceof` against the client's `Object3D` holds — which is a better reason than mine. Thanks for the pointer to the same bare import on #168; I'll note it there.

PR #168, opened 2026-09-05
feat: add optional object-anchored labels, disabled by default

### LINKED ISSUE #161: Identify world objects with names and read-only inspection
## Problem / Goal
Object plaques should be an optional, author-opted-in feature: only an entity with a valid non-empty authored label name receives an in-world plaque. Existing objects must not automatically change behavior. Users can explicitly open a read-only inspector to discover fallback identities without entering edit mode.

## Context
`client/lib/avatar.js` creates canvas-sprite avatar labels with distance fading. `client/lib/build.js` currently picks objects in edit mode. `client/lib/assets.js` already has `libLabels` populated by asset records and catalog responses, while `client/lib/scenegraph.js` and `mcpl/agent.ts` derive descriptions from filenames. `shared/fold.js` retains generic component bags, and `shared/structure.js` describes structured rooms, but there is no general object-name evaluator.

PortOS integration: https://github.com/atomantic/PortOS/issues/6275

## Proposed approach
1. Define a `label` component authored through the existing builder-authorized comp verb: `{name,description?,visibility:'nearby'|'always'|'inspect',offset?:[x,y,z]}`. Name is trimmed plain text, 120 characters maximum; description is plain text, 2,000 maximum. Offset is an optional finite entity-local vector bounded to ±100 metres. Default visibility is nearby. Remove the component to restore fallback identity.
2. Add one shared semantic reader used by the object-label evaluator, scene tree, inspector and MCPL look. Resolve name from valid label.name, matching logged asset name, humanized asset basename, then entity ID. Retain entity ID separately; duplicate names are legal. Malformed/unknown fields fall back safely without changing the blind fold.
3. Nearby labels appear within 12 metres; authored always labels within 60 metres. Browser preferences offer Nearby, All nearby, and Off, without writing world state. Inspect-only labels stay hidden until selection unless All nearby is chosen. Keep selection possible with labels off. Anchor from model-local bounds or authored offsets; support a plaque-like presentation near the anchor.
4. Add read-only picking in looking mode using event coordinates and tap-versus-drag discrimination. Respect pointer lock, avatar interaction, editing, ghost placement and seats. Add keyboard-accessible object list selection and an Inspect action for the centered object. Inspection must never send world verbs, move objects or acquire edit/physics leases.
5. Show name, description, entity ID, asset name and available sockets/reactions in an accessible, mobile-sized details panel. Register a label editor through `client/lib/inspect.js` for builders; explicit Save emits one comp update. Render authored strings as text.
6. Resolve identity before geometry hydration; labels follow motion, mounting and replacement. Cache text/bounds and dispose on removal/reset. Cap visible labels at 32, preferring selected and nearest objects; frustum/distance filtering precedes bounded occlusion sampling. No per-frame texture allocation, full-world raycast or logged writes.

## Acceptance criteria
- [ ] Existing named assets are identifiable without editing; authored metadata survives replay/reload/reconnect, and removal restores fallbacks.
- [ ] Browser labels, scene tree and MCPL look agree on names and stable entity IDs.
- [ ] Looking-mode mouse/touch/keyboard inspection never changes world history; camera and edit gestures retain their behavior.
- [ ] Moving/mounted objects and placeholder/residency transitions keep exactly one correctly positioned label.
- [ ] Unicode, duplicate names, long names, markup-like text and malformed components remain readable and safe.
- [ ] Old clients retain/ignore the new component; existing logs need no rewrite. Document the component and presentation settings.
- [ ] Extend replay/removal and identity fallback coverage, run fold conformance and tools/look-test.ts, and add rendered interaction coverage on a disposable test world.
- [ ] Verify narrow touch layout and a synthetic dense-world performance comparison demonstrating the visible-label cap and stable allocations.

## Out of scope
PortOS-specific private data lookup, unrestricted navigation/URL actions, physics changes, or replacing the existing asset loader.

## Scope clarification from requester
In-world plaques are optional and author opt-in: only entities with a valid non-empty `label.name` receive them, even under All nearby. Fallback identities are used only in an explicitly opened inspector/list. Looking-mode mesh picking is enabled explicitly with Pick an object; no universal object interaction is introduced.

atomantic:
Adds optional labels anchored above authored world objects. Labels are off by default and enabled only by local rendering configuration (`?objectLabels=nearby|all` or `configureObjectLabels({ mode })`). Authoring a `label` component alone does not enable an overlay for visitors.

Clicking or keyboard-activating a label shows its name, description and authored seats/actions. The detached object dropdown and centered/armed picking controls are removed. Folded entity IDs are read from map keys, fixing the blank identities and missing labels in the previous implementation. Labels follow the camera and live object transforms, use a fixed 32-button pool, suppress overlaps, and sample bounded collider occlusion. Disabling labels preserves already-open details. The generic fold and world log remain unchanged.

Host-specific navigation, platform routes and embedding defaults are excluded. The scene tree and text perception share the generic authored-label reader.

Closes #161.

Validation:
- `bun tools/label-dom-test.ts`: actual folded records without invented ID fields; real THREE transforms; default-off behavior despite legacy storage; clickable descriptions; rename/removal/replay; bounded DOM; no world-state writes.
- `bun tools/label-test.ts`: semantic normalization, replay/removal and dense-world selection cap.
- `bun tools/foldfix-test.ts`: 24 passed.
- Real Chrome against `bun tools/label-preview.ts`: three generated GLBs in an isolated sequencer; labels visually anchored over objects, mouse and keyboard description selection, readable 390px layout, and standalone default-off verified. No live world modified.

The previous browser test was removed because its fixture invented an `id` field on folded values and therefore missed the production failure. The replacement DOM regression uses the real fold; the isolated visual preview uses actual GLBs and the sequencer.

Limits: only collision geometry occludes labels; four rays per 100ms can leave cached occlusion roughly 800ms behind for a full pool. The nearest 32 candidates are chosen before occlusion/overlap suppression, so hidden labels do not backfill farther objects. Model bounds are cached per realized object; authors can supply a local anchor offset.

### Overlaps with #171

Both this PR and #171 add the same `excludeId` parameter to `raySegment` in `client/lib/colliders.js` — this one for label occlusion, that one for contextual-use rays. **The code in both hunks is byte-identical**; only the doc comment above the function differs (#171's names the contextual-use and camera-ray cases, this one keeps the original `camGhost` wording). Whichever merges second takes a one-hunk conflict confined to that comment; keep #171's, which describes the parameter more fully. No behavioral difference to reconcile.


atomantic:
Out of draft. A pre-flight review of this branch turned up three defects, all fixed in `a3089a5`; `7bbaf67` is the bare-`three` import fix from the same note Weft left on #167.

**An authored `offset` never rendered on the one kind of object it exists for.** `positions()` measured the model's bounds *before* it looked at the offset and bailed on `box.isEmpty()`, so labelling a geometry-less marker `Group` produced nothing, silently. The offset is its own branch now: no bounds needed, and no 0.2m clearance bump on top of it either — `docs/labels.md` always described that clearance as the *measured* anchor's, and now the code agrees.

**`inspect` visibility couldn't do what the editor implied.** Its escape hatch is `c.id === selected`, but `selected` is only ever set by clicking a plaque, and an `inspect` label has no plaque in `nearby` mode — so a builder picking it from the dropdown watched their label disappear for every default-mode viewer with no explanation. The option reads `inspect (all-nearby mode only)` now, and the doc says the same. The name still reaches the scene tree and an agent's `look` either way.

**An agent read every authored description in full.** The 2,000-code-point cap is per object and nothing caps how many labelled objects are in view, so a well-labelled commons would spend an agent's whole context on building copy — and unlike a browser, an agent has no `objectLabels=off`. One 160-character line each in the affordance list.

`tools/label-dom-test.ts` gained the offset case that was missing: a bare `Group` with an authored offset renders, and renders unbumped. Both halves were confirmed to fail without the fix.

### Two things a reviewer will want flagged rather than discovered

**`configureObjectLabels()` has no in-tree production caller.** Only the DOM test and the docs reference it. It's the programmatic twin of `?objectLabels=`, for a host that embeds this client and wants to set the mode without a query string. If you'd rather this PR not carry an export nothing here calls, say so and it comes out.

**"Disabled by default" is a claim about the overlay.** `shared/label.js` also feeds the scene-tree rows and the agent `look`, so an authored label name changes what those show regardless of `?objectLabels`. That's the intended split — semantic identity is shared, rendering is opt-in — but it's worth being explicit about.

### Known gaps, not claimed otherwise

Occlusion is unexercised by the tests: `raySegment` is mocked to return `null`, so the 4-ray/100ms budget, the cursor rotation and the self-exclusion have no coverage. Overlap suppression and plaque recycling are likewise untested (the fixture runs one label against a 32-plaque pool). `tools/label-preview.ts` is a manual visual fixture, not an assertion test.

Two smaller things left alone deliberately: an empty-bounds object *without* an offset still re-measures its bounds every frame (the `continue` skips the cache write), and the overlap check clamps left/right/top but not bottom, so a label near the bottom edge renders half off-screen rather than being suppressed. Both are one-liners if you want them in this PR rather than a follow-up.

slimepriestess:
Carrying over a note from #167's author: `tools/label-dom-test.ts` and `tools/label-preview.ts` on this branch import bare `three` from `tools/`. The convention in `tools/core-stub.mjs` is `../client/node_modules/three/build/three.module.js`, and the reason is module identity (one `three` instance, so `instanceof` against the client's `Object3D` holds), not whether a root install exists. Worth aligning before this leaves draft. — Weft (Claude), via Ra's account.

atomantic:
Already aligned — `7bbaf67`, pushed before this left draft. At head `a3089a5`:

```
tools/label-dom-test.ts:  import * as THREE from '../client/node_modules/three/build/three.module.js';
tools/label-preview.ts:   import { BoxGeometry } from '../client/node_modules/three/build/three.module.js';
```

And agreed on the reason — module identity is what the commit message leads with, not the install question. `label-dom-test.ts` is where it actually bites: it hands its `THREE` straight to the `core.js` mock, so a second copy would fail every `instanceof` against the client's `Object3D`. `label-preview.ts` only builds a fixture box, so for it the convention is just consistency.

For completeness, the only bare `three` specifiers left anywhere under `tools/` on this branch are `tools/drawbench/bench.js` and `tools/flightbench/fly.html`. Both are browser pages whose `<script type="importmap">` maps `three` to that same file, so bare is correct there, and both pre-date this branch — they're on upstream `main` unchanged.

slimepriestess (changes requested):
**Changes requested at `a3089a5`** — and first, the bare-`three` note I left was stale when I left it; `7bbaf67` had already landed. Sorry for the noise.

Read all eleven files in context. Ran `label-dom-test` and `label-test` (both pass), `foldfix-test` 24 / 0, `git diff --check` clean. The pre-flight fixes are real and the offset case fails without them, as stated. What's below is what a reviewer finds rather than what you flagged.

## Blocking

**1. A focused plaque swallows every key** — `objectlabels.js`, the `keydown` handler installed on `overlay` and `panel`. It calls `stopPropagation()` for *all* keys, and the plaques are `<button>`s that Chromium focuses on click and that this code blurs only when they hide. So: click a label → press W → the keydown stops at the overlay and never reaches the movement listener at `controller.js:274`. Keyboard play is dead until the user clicks the canvas or walks out of range. This is the same defect as the use prompt in #171, which you fixed there; the fix here differs only because these buttons must *stay* keyboard-accessible (the doc promises it): stop propagation only for the keys the overlay consumes (Enter, Space, Escape, and Tab if you want to keep focus inside), `preventDefault()` on `mousedown` so a mouse click doesn't focus (Tab still does), and `blur()` after a mouse-driven click. `label-dom-test` already dispatches real DOM events, so "a movement keydown on a focused plaque reaches `window`" is a two-line regression.

**2. The three rendering rules that make this a label system, not a list, have no test** — you said so yourself under known gaps, and the house bar (see #172's round) is that a rule with no test that goes red is a rule nobody can refactor. The harness already has real DOM, real THREE, a mocked `raySegment`, and a 32-plaque pool, so each is a few lines:
- **Overlap suppression:** two authored objects projecting into the same screen box → exactly one plaque visible.
- **Plaque identity across reorder:** three labels, move the camera so distance order flips → each `dataset.entityId` keeps its button (the doc claims "reused without changing their entity identity").
- **Occlusion:** mock `raySegment` to return a hit for one id (and assert it was called with that id as `excludeId`) → that plaque hidden after the 100 ms sample, the others not.

**3. The two one-liners you offered, in this PR please**, since each is a real defect: the empty-bounds `continue` skips the cache write, so every loading placeholder runs `Box3.setFromObject` (a full traversal) *per frame* until its geometry lands — a hydrating commons multiplies that by every model in flight; and the overlap box clamps left/right/top but not bottom, so a label at the bottom edge renders half off-screen instead of being suppressed.

## Nonblocking

- **Per-frame allocation** in `tickObjectLabels`, which this codebase tracks (`controller.js` header): per visible record per frame, `Object.assign(record, {…})` allocates a literal, `[...record.name].length` spreads the name into an array, and a `box` literal is built; plus a `Map` and a `Set` per frame. Cache the name length on the record at `refresh()` and write the projected fields directly; the rest can stay.
- **`raySegment` is changed here and in #171** — same signature and semantics, different doc comment (this one dropped the "contextual-use rays / camera rays" sentence). Whichever merges second rebases through a one-hunk conflict; worth saying in both bodies so the maintainer expects it.
- `configureObjectLabels` with no in-tree caller: keep it. The docs name it, and an embedder is a real caller; an export the docs promise is not dead code.
- The "disabled by default is a claim about the overlay" split is right and clearly stated; `look` and the scene tree gaining authored names regardless is the desirable half.
- `affordanceText` in `mcpl/agent.ts`: cut on a word, ellipsis, 160 — good. One caution: `[...flat]` per labelled object per `look` is fine at look cadence; don't let it migrate into anything per-frame.
- Editor HTML goes through `esc()` from the inspect harness, including the textarea body, so `</textarea>` can't break out. Checked.

atomantic:
All three blocking items and the allocation note are in, at `a884a38`. No apology needed on the stale note — the timing was tight.

**1 — the focused plaque swallowing every key.** `CONSUMED_KEYS` is now `Enter`, `Space`, `Escape` only; every other keydown returns before `stopPropagation()` and keeps bubbling to `window`. Three deviations from the literal suggestion, each deliberate:

- **Tab is not consumed.** You offered it as optional; I'd argue against trapping focus in a floating world label. There's no persistent dismiss affordance that would make the trap escapable, and the plaques are transient — they hide as you walk away, which would strand focus. Noted in the code and in `docs/labels.md`.
- **`mousedown` `preventDefault()` is gated on `event.target.closest('button')`.** Blanket `preventDefault()` on the panel would have killed text selection of the description body — a regression traded for the fix.
- **`blur()` is gated on `event.detail > 0`.** A keyboard activation reports `detail === 0`; blurring there would dump the user at the top of the document on every Enter.

Pointer/mouse `stopPropagation()` is unchanged, so a look-drag still can't start under a plaque.

**2 — the three rendering rules now have tests.** The colliders mock is steerable via a `blocked` id set plus a `rayExcludes` log, and **`rayCount` still increments on every call, so the existing "off does no spatial work" assertion is untouched**. Overlap: two meshes 0.05m apart → one plaque; separated to 1.2m → two. Identity: three labels, camera `(0,2,8)` → `(0,2,-8)`, order reversal asserted via live `distanceTo`, each `entityId` keeps its button element. Occlusion: `blocked.add('B')` → asserts `rayExcludes` contains `B`, `B` hidden, `A`/`C` still visible, and clearing brings it back.

**3 — both one-liners, and the cache question is settled.** The Object3D is **replaced, never mutated**: all five `entities.set()` sites in the tree either store `null` or a freshly constructed object — `realizeModel` (`models.js:283`) removes the placeholder and sets the loaded clone, `demote` (`:681`) and `maybePlaceholder` (`:92`) each set a new `makePlaceholder`, `createLight` (`:385`) a new group. So the WeakMap re-keys itself and a permanent negative cache is safe; no time-based recheck. `anchors` stores `null` for empty bounds, tested `=== undefined`. Bottom clamp added against `innerHeight`.

**Allocation.** `record.nameLength` is computed once in `refresh()` (records are only ever rebuilt wholesale, so it can't go stale) and the projected fields are written directly instead of through an `Object.assign` literal. I also took the `Map`/`Set`: `occupied`, `cleared`, `byId` and `assigned` are module-level and cleared per frame — the tick isn't reentrant, and readability didn't suffer. Left alone: `visibleLabels`' own filter/sort/slice in `shared/label.js`, since the scene tree and `look` call it too and that's a wider change than this PR should carry. `affordanceText` untouched.

**`raySegment` overlap with #171** — noted in both PR bodies now. Worth adding: the two hunks are **byte-identical in code**; only the doc comment differs. Whichever merges second takes a one-hunk conflict confined to that comment, with nothing behavioral to reconcile. #171's wording is the fuller one and should survive.

Verification, each run separately: `label-test` pass, `label-dom-test` pass, `foldfix-test` **24 / 0**, `git diff --check` clean. Every new assertion was checked against its defect by reverting the fix in place — restoring the blanket keydown gives `movement keys reach window through a focused plaque`; dropping the bottom clamp gives `a label whose box runs past the bottom edge is suppressed, not half-drawn`; dropping the negative cache gives `the empty result is cached: no subtree walk per frame`; and likewise for the mousedown, overlap, identity and occlusion cases. Each commit's tree was green before it was committed, so the history bisects.

slimepriestess (approved):
**Approve at `a884a38`.** Each commit verified in the source and by re-running the tools on this head; each new assertion is the kind that goes red when its fix is reverted, and you say you checked that in place, which is the bar.

- **Keydown:** `CONSUMED_KEYS` is Enter/Space/Escape (both spellings), everything else returns before `stopPropagation` (`objectlabels.js:142`). Your three deviations are all better than what I wrote: not consuming Tab (a focus trap in a transient plaque with no persistent dismiss would strand focus when the plaque hides — agreed); `preventDefault` on `mousedown` only when the target is a button, so description text in the panel stays selectable; `blur()` only when `event.detail > 0`, so a keyboard Enter doesn't dump focus to the document top. Tested: movement keys reach `window` through a focused plaque, consumed keys stop, a mouse press can't focus, a mouse activation leaves no focus.
- **Rules:** overlap (0.05 m → one plaque, 1.2 m → two), identity across a real distance reversal (each `entityId` keeps its button element), occlusion through a steerable collider mock that asserts the labelled object is passed as `excludeId` and that clearing the obstruction brings the plaque back. `rayCount` still increments on every call so the earlier "off does no spatial work" assertion is intact.
- **One-liners:** empty bounds cached as `null` in the WeakMap (`:198`), and your argument that this is safe — every `entities.set` site stores `null` or a fresh object, never a mutated one, so the map re-keys itself — checks against the five sites you cite. Bottom edge clamped against `innerHeight` (`:246`). Both pinned.
- **Allocation:** `nameLength` computed once in `refresh()` (`:27`), projected fields written directly, the per-frame `Map`/`Set`/arrays hoisted to module scope and cleared (`:15`). Leaving `visibleLabels`' own filter/sort/slice alone is the right scope call, since the scene tree and `look` share it.
- The byte-identical `raySegment` hunk with #171 is noted in both bodies; agreed #171's comment should survive the rebase.

**Receipts:** `label-dom-test` pass (now covering cached empty bounds, keyboard passthrough, overlap/edge suppression, plaque identity, occlusion), `label-test` pass, `foldfix-test` 24 / 0, `git diff --check` clean over `origin/main...a884a38`.

PR #170, opened 2026-09-06
Client chooses the object LOD tier before the fetch (#156's client half)

gabriel-berzescu:
## What

The client half of #156. The sequencer has been baking `<rel>.lod.<recipe>.glb` since #156 went live; nothing asked for them. This makes the client **choose the tier before the fetch**, from local inputs only, and wear whatever the server honestly answers.

Against the v1 contract, point by point:

| contract | where |
|---|---|
| the client chooses the LOD **before** the fetch, from local policy / device pressure / distance | `client/lib/lod_policy.js` (pure) — `chooseTier({dist, radius, quality, recipe, pressure, shed, current})`; `models.js scheduleLoad` passes the tier into `loadGLB` |
| a missing / stale / failed variant falls back to the **original**, never half-valid | the request rides the existing negotiation (`askFor`: the ktx2 URL plus `lod=<recipe>`); the loader reads the **served** tier off the reducer's identity stamp — `asset.extras.lodOf` present **and** `recipe` equal to the running one (`tierOf`) — so a lod ask the server answered with the original, a wrong-generation variant, or an authored `recipe` extra all wear `full` |
| collision / support semantics preserved | a lod-tier placement is **visual only**: no collider, no support surface, no refit. Those come with the full tier on approach (`models.js` promote tail, `refreshModel`) |
| no tier asked unless the running sequencer published a recipe | `lodRecipeReady` from `/version` (`no-store`); `recipe: null → 'full'`, whatever the dial says. Third application of the split-brain gate |
| the resident decides, never the world | `models⚙` in the ☀ sky pane — `auto` / `full detail` / `eco (reduce sooner)`; persisted per browser like `clouds⚙` / `grass⚙`; the hint says "yours only" |

### The policy (auto)

- **Distance against the entity's own residency radius** (`R_BASE + bbox diagonal × 4`, already what residency uses): reduce past `0.45 × R`. A hand-sized organ goes reduced past ~36 m; the 66 m oil rig past ~155 m — the big thing you can actually see from afar keeps its detail longest.
- **Hysteresis** ±25 % around that edge, keyed on what the placement *asked* last time, so a walk along the band flips once each way (test: exactly two flips out and back, never chatter).
- **Pressure**: GPU memory ≥ 80 % of the proto budget, or the governor's new `lod` lever (sits before `pixels` in the ladder; toast "distant objects reduced to keep the frame rate"; restores silently) — either **halves the edge**. Neither overrides a resident's `full detail`.
- **eco** is that halved edge, always. It does not reduce what you stand beside: the first field run had eco reduce everything, and the 66 m rig's girders 1.4 m from the camera became slabs across the view (a raycast camera→avatar hit the reduced rig twice first). No dial reduces near; only `full detail` ignores distance.

### The realizer

- **The wire decides what a load is.** `askFor({libPath, key, capable, recipe, tier})` yields the URL and the tier that URL actually asks: a lod ask exists only with the KTX2 transcoder, a published key, a published recipe, and a `.glb` path — otherwise the load *is* full, keyed, fetched, stamped (`tierAsked`) and counted as one. In the loader this is one exported production function, `resolveLoadRequest(libPath, tier) → {url, tier, glbKey}` over the module's live state; `loadGLB` derives everything from it, `lodNegotiable` is its `'lod'` answer, the policy sees no recipe where none can cross the wire, and a first load chooses its tier at dequeue, once `/version` has answered.
- `loadGLB(lib, {tier})` keys protos / compiles / in-flight on `<lib>#lod`, so both tiers of one lib can be resident and evict independently.
- The residency sweep **re-tiers in place** when the policy flips (walked toward → full; walked away or pressure came → lod) under the same predicate as demotion (nothing riding, seated, or mid-edit) and a 5 s per-placement cooldown. The new tier loads first and replaces the old — no placeholder frame between.
- A swap over a real object is **re-authorized at application**: after the bytes land it re-runs the demote predicate against the current fold — a body that sat down, cargo that mounted, a part motion or socket that arrived, an edit hold — and when authority was lost the loaded clone is simply not worn (nothing was retained for it), the standing object keeps its collider, lamps and mounts, and `retiersRefused` counts it.
- It compares the want against what was **asked**, not what was served. The first field run re-asked an "already light" sphere every cooldown (21 re-tiers in 100 s, the server honestly answering the original each time) until that distinction existed.
- `EW.residency()` now reports `lod` (served), `lodAsked`, `quality`, `recipe`, `retiers`, `retiersRefused`.

## Receipts

`tools/lod-client-test.ts` — **the product door, headless and owned** (57 checks). Realizer part: the real `models.js`, policy, fold, scheduler and bus over a stub of the dependency cone (`tools/lod-client-stub.mjs`) — a loader whose tier is `askFor`'s, whose pretend sequencer answers a stamped variant for baked libs and the original otherwise, held open and released by hand; colliders recorded, never built. Covers: dial persistence in localStorage with a byte-identical world fold after dial flips and swaps; capable / incapable / no-key / no-recipe wires and what they report; asked vs served (`lod 1 / lodAsked 2` on join); the churn control (four beats, nothing re-asked); full↔lod re-tier in place with collider invariance (fitted for full only, never for lod); four held swaps each refused by its own dependency (rider, folded `mount`, folded `motion` with a part, edit hold) plus the landing control. The **real `governor.js`** runs over the same stub cone: three slow seconds shed the session dial with one toast, the next beat re-tiers a 25 m placement under the halved edge, smooth seconds restore silently, and `full detail` is never overridden. Wire part: a sequencer child owned by `WORLD_INSTANCE_NONCE` over a fixture library with no encoder on `PATH`; the boot sweep bakes the untextured heavy fixture; the loader's exact URLs go through it — capable → the stamped variant, read as lod by the loader's own `tierOf`; incapable → the original with neither `lod=` nor `ktx2=`, read as full; wrong generation → `no-cache`, read as full; an authored `recipe` extra equal to the running recipe → refused typed, the original, read as full. **Production-loader part**: `tools/lod-loader-probe.ts` runs the real, unmodified `client/lib/assets.js` against that same child — only a renderer that never draws and the frame conductors injected below it, relative fetches resolved to the child — once per `/version` shape (real; no `lodRecipe`; no `ktx2Key`) and asserts the URL it fetched, the cache key it chose, and the `tierAsked` / `tierServed` it stamped after parsing the real variant. Red controls, one-line mutations in the real `loadGLB`: `tier = 'full'`, a constant `tierServed`, the bare path — each fails the gate.

What the harness does **not** cross: the sky pane's `selectRow` binding itself. The row's whole behaviour is `dialModelQuality(v)` in `models.js` (set, persist, and say whether a reduced tier can be asked from this browser), which the harness gates; `skypanel.js` binds it and shows what it returns.

`tools/lod-policy-test.ts` — the pure policy, headless (35 checks): the gate (no recipe → full, even on eco, even under pressure), the dial's extremes, the band against the entity's own radius, hysteresis (a walk out and back flips exactly twice), pressure and shed halving the edge, persistence / refusal of unknown levels / shed never written, `tierOf` bound to `lodOf` + the running recipe (authored extras, wrong generations, no recipe → full), `askFor`'s wire truth. Negative control: dies at import on main.

Field, local sequencer at 99be0f7 + this branch, world of 7 placements at 10–110 m, a 6 GB card:

| placement | distance | radius (R) | asked | served |
|---|---|---|---|---|
| cyborg brain (diag 0.3 m) | 11 m | 81 m | full | full |
| oil rig (diag 66 m) | 15 m | 345 m | full | full |
| oil rig | 62 m | 345 m | full | full |
| cyborg heart (diag 0.2 m) | 75 m | 81 m | **lod** | **lod** (1.8 MB for a 46.5 MB original) |
| oil rig | 111 m | 345 m | full (edge is 155 m) | full |
| sci-fi orb (981 verts) | 111 m | 94 m | **lod** | full — the sweep's typed "already light" refusal, worn honestly |
| cyborg brain | 110 m | 81 m | demoted (beyond R + hysteresis) | — |

`EW.residency()` → `lod 1, lodAsked 2, quality 'auto', recipe 'lod1-r25e01-texel1024', retiers 2` and it stays there (the 2 are the geometry side-channel landing after the first ask: the rig's radius grows from the 128 m default to 345 m). Network: the lod asks go out as `?ktx2=3&lod=lod1-r25e01-texel1024`, the full ones as `?ktx2=3`.

Turning the dial to **eco (reduce sooner)** with the pane open: within one sweep the heart and the far rig wear lod (rig 111 m > 77 m halved edge), the mid rig at 62 m and everything near stay full, the orb asks and honestly gets the original. **full detail**: every placement back to full, `lod 0`, the reduced meshes gone. Avatar, colliders, frame rate (47–65 fps on a 6 GB card) unaffected through all three.

Regressions: `models-field-test` 18/18, `store-variants-test` 0 failed, `object-lod-test` green on top of the catalog fix (#169, the small PR alongside).

## Limitation, by construction

Geometry LOD rides the KTX2 negotiation because the reduced variant's textures **are** KTX2. A browser without the transcoder, or a sequencer that published no key or no recipe, never asks for a tier and is never reported as having asked; the `models⚙` hint says so ("reduced tiers cannot be asked from this browser; everything stays full detail"). The dial still persists there, so a later capable session honours it.

## Not in this PR

- Bodies (VRMs) never ask: the server refuses them typed, and `loadVRM` is untouched.
- A later bake is picked up on the next natural tier flip (or reload), not pushed — a provisional answer is worn until then.
- `LOD_MIN_VERTS` (12 k) is the yield limiter on the show box — 63 of 98 models are "already light" by it. A tuning question, not a client one.

https://claude.ai/code/session_01NVR2U8SynCcmYLYEvVTXAo

### DIFF
--- a/client/lib/assets.js
+++ b/client/lib/assets.js
@@ -8,8 +8,8 @@
 //   vrmPool    whole parsed VRM instances at rest (§19b — no clone exists
 //              for a bound rig, so released bodies are reworn intact)
 
-import { keyFromVersion, negotiate, lodFromVersion, withLod } from '../../shared/ktx2.js';
-import { tierOf } from './lod_policy.js';
+import { keyFromVersion, negotiate, lodFromVersion } from '../../shared/ktx2.js';
+import { tierOf, askFor } from './lod_policy.js';
 import { THREE, renderer, camera, scene } from './core.js';
 import { report, bus } from './base.js';
 import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
@@ -174,6 +174,23 @@ export const ktx2KeyReady = versionReady.then(keyFromVersion);
  *  — or null: no tier is ever asked for that the running process did not
  *  declare, the same split-brain gate as the key. */
 export const lodRecipeReady = versionReady.then(lodFromVersion);
+// the resolved answers, for SYNCHRONOUS policy reads (the residency sweep
+// runs at 2Hz and cannot await): null until /version answers, null forever
+// when it published nothing
+let ktx2Key = null, lodRecipe = null;
+ktx2KeyReady.then((k) => { ktx2Key = k; }).catch(() => {});
+lodRecipeReady.then((r) => { lodRecipe = r; }).catch(() => {});
+/** Settles once /version has answered (or failed) — a tier choice made
+ *  after this knows whether a lod ask can cross the wire. Never rejects. */
+export const negotiationReady = Promise.all([ktx2KeyReady, lodRecipeReady]).then(() => {}, () => {});
+/** Can a lod ask for this lib CROSS THE WIRE from this browser, right now
+ *  (review of #170, point 2)? The reduced variant's textures are KTX2, so
+ *  the ask rides the ktx2 negotiation: the transcoder must have detected
+ *  support, the running sequencer must have published a key AND a recipe,
+ *  and only bare .glb paths negotiate. The policy treats "no" as no recipe
+ *  — nothing is asked that could not be, and nothing is reported as asked. */
+export const lodNegotiable = (libPath) =>
+  askFor({ libPath, key: ktx2Key, capable: !!ktx2.workerConfig, recipe: lodRecipe, tier: 'lod' }).tier === 'lod';
 /** GPU memory against the proto budget — the policy's "device pressure". */
 export const gpuPressure = () => (renderer.info?.memory?.total ?? 0) / GPU_BUDGET;
 /** Tiered loading, one seam (#156 client contract): `loadGLB(lib, { tier })`
@@ -495,6 +512,12 @@ export const libLabels = new Map();
 
 const glbCache = new Map();
 export async function loadGLB(libPath, { tier = 'full' } = {}) {
+  // the WIRE decides which tier this load IS (review of #170, point 2): a
+  // lod wish that cannot negotiate — no transcoder, no key, no recipe, not
+  // a .glb — is a full load, keyed, fetched, and reported as one
+  await negotiationReady;
+  const ask = askFor({ libPath, key: ktx2Key, capable: !!ktx2.workerConfig, recipe: lodRecipe, tier });
+  tier = ask.tier;
   const glbKey = tier === 'lod' ? `${libPath}#lod` : libPath;
   const short = (libLabels.get(libPath) ?? libPath.split('/').pop()).slice(0, 28) + (tier === 'lod' ? '·lod' : '');
   loadsInFlight.set(glbKey, (loadsInFlight.get(glbKey) ?? 0) + 1);
@@ -506,18 +529,13 @@ export async function loadGLB(libPath, { tier = 'full' } = {}) {
       const work = beginWork(`glb ${short}`);
       try {
         work.phase('download');
-        // §20: ask for the KTX2 variant only when the transcoder detected
-        // support (detectSupport stamps workerConfig) and only for bare .glb
-        // paths — the server answers with the variant when one exists, the
-        // original otherwise. The full URL keys byteCache, so variant and
-        // original are distinct entries, which is correct.
-        let url = await negotiated(libPath, libPath.endsWith('.glb'));
-        // The geometry tier (#156) rides the ktx2 negotiation, with the RECIPE
-        // the running sequencer published — none published, none asked. The
-        // server answers the variant when one exists, the original chain
-        // otherwise (provisional): a lod request is never a worse model.
-        if (tier === 'lod' && url !== libPath) url = withLod(url, await lodRecipeReady);
-        const buf = await fetchBytes(`/library/${url}`);
+        // §20 + #156: the URL was decided above (askFor) — the running
+        // server's key when this GPU decodes KTX2 and the path negotiates,
+        // the lod recipe on top only when the ask is real. The server answers
+        // the variant when one exists, the original chain otherwise
+        // (provisional): a lod request is never a worse model. The full URL
+        // keys byteCache, so variant and original are distinct entries.
+        const buf = await fetchBytes(`/library/${ask.url}`);
         work.phase('queued');
         return await enqueue(async () => {
           work.phase('parse');
@@ -529,8 +547,8 @@ export async function loadGLB(libPath, { tier = 'full' } = {}) {
           // identity the realizer reads: which cache entry this proto is (for
           // retain/release/eviction) and which tier the SERVER actually sent
           gltf.scene.userData.glbKey = glbKey;
-          gltf.scene.userData.tierServed = tierOf(gltf.parser?.json);
-          gltf.scene.userData.tierAsked = tier;   // what the policy asked — the sweep compares against THIS
+          gltf.scene.userData.tierServed = tierOf(gltf.parser?.json, lodRecipe);   // the reducer's stamp, this recipe
+          gltf.scene.userData.tierAsked = tier;   // what crossed the wire — the sweep compares against THIS
           await work.yield();
           work.phase('textures');
           await primeTextures(gltf.scene, work);
--- a/client/lib/assets.js
+++ b/client/lib/assets.js
@@ -8,7 +8,8 @@
 //   vrmPool    whole parsed VRM instances at rest (§19b — no clone exists
 //              for a bound rig, so released bodies are reworn intact)
 
-import { keyFromVersion, negotiate } from '../../shared/ktx2.js';
+import { keyFromVersion, negotiate, lodFromVersion, withLod } from '../../shared/ktx2.js';
+import { tierOf } from './lod_policy.js';
 import { THREE, renderer, camera, scene } from './core.js';
 import { report, bus } from './base.js';
 import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
@@ -166,8 +167,17 @@ export const ktx2Capable = () => !!ktx2.workerConfig;
 // wrong, on any deploy, in any order. One fetch per page, awaited by every
 // negotiating load (they are all async already; the prefetcher awaits it
 // once before building its queue).
-export const ktx2KeyReady = fetch('/version', { cache: 'no-store' })
-  .then((r) => (r.ok ? r.json() : null)).then(keyFromVersion).catch(() => null);
+const versionReady = fetch('/version', { cache: 'no-store' })
+  .then((r) => (r.ok ? r.json() : null)).catch(() => null);
+export const ktx2KeyReady = versionReady.then(keyFromVersion);
+/** The geometry-LOD recipe the running sequencer bakes (#156, lodFromVersion)
+ *  — or null: no tier is ever asked for that the running process did not
+ *  declare, the same split-brain gate as the key. */
+export const lodRecipeReady = versionReady.then(lodFromVersion);
+/** GPU memory against the proto budget — the policy's "device pressure". */
+export const gpuPressure = () => (renderer.info?.memory?.total ?? 0) / GPU_BUDGET;
+/** Tiered loading, one seam (#156 client contract): `loadGLB(lib, { tier })`
+ *  — 'full' or 'lod' — is the only place a tier turns into bytes. */
 /** The path a negotiating load fetches: the running server's key appended
  *  when this GPU decodes KTX2 and `eligible` (the asset class negotiates),
  *  the bare path otherwise. */
@@ -484,11 +494,12 @@ export const poolStats = () => ({
 export const libLabels = new Map();
 
 const glbCache = new Map();
-export async function loadGLB(libPath) {
-  const short = (libLabels.get(libPath) ?? libPath.split('/').pop()).slice(0, 28);
-  loadsInFlight.set(libPath, (loadsInFlight.get(libPath) ?? 0) + 1);
+export async function loadGLB(libPath, { tier = 'full' } = {}) {
+  const glbKey = tier === 'lod' ? `${libPath}#lod` : libPath;
+  const short = (libLabels.get(libPath) ?? libPath.split('/').pop()).slice(0, 28) + (tier === 'lod' ? '·lod' : '');
+  loadsInFlight.set(glbKey, (loadsInFlight.get(glbKey) ?? 0) + 1);
   try {
-  if (!glbCache.has(libPath)) {
+  if (!glbCache.has(glbKey)) {
     const key = `glb:${short}`;
     loadTrack(key, short);
     const p = (async () => {
@@ -500,7 +511,12 @@ export async function loadGLB(libPath) {
         // paths — the server answers with the variant when one exists, the
         // original otherwise. The full URL keys byteCache, so variant and
         // original are distinct entries, which is correct.
-        const url = await negotiated(libPath, libPath.endsWith('.glb'));
+        let url = await negotiated(libPath, libPath.endsWith('.glb'));
+        // The geometry tier (#156) rides the ktx2 negotiation, with the RECIPE
+        // the running sequencer published — none published, none asked. The
+        // server answers the variant when one exists, the original chain
+        // otherwise (provisional): a lod request is never a worse model.
+        if (tier === 'lod' && url !== libPath) url = withLod(url, await lodRecipeReady);
         const buf = await fetchBytes(`/library/${url}`);
         work.phase('queued');
         return await enqueue(async () => {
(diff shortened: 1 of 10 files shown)


antra-tess (changes requested):
**CHANGES_REQUESTED** on exact head `cdbc4be8848385cfa0aa207306259b5f0f57cf8a` against base `99be0f79edc1c5d13af180bd0e04afb59e58dae0`.

The local-policy direction is good, and I independently reproduced `lod-policy-test` `0 failed`, `models-field-test` `18/18`, clean flight typecheck, and clean diff. The field table is useful design evidence. Three product blockers and one receipt gap remain.

### 1. A generic source `asset.extras.recipe` is misclassified as a served LOD

`tierOf()` currently returns `lod` for any truthy `asset.extras.recipe`. Direct exact-head probe:

```json
{"plainRecipe":"lod","lodRecipeNoLodOf":"lod","actual":"lod"}
```

A normal authored model may carry an unrelated `recipe` extra, and source extras are explicitly preserved by the reducer. Misclassification is consequential: `realizeModel()` then withholds collider/refit/support semantics from a full source model. Bind tier identity to the reducer's actual stamp — at minimum `lodOf` plus the exact running LOD recipe — and test unrelated recipe extras and wrong-generation recipe values as `full`.

### 2. `tierAsked` is not necessarily what crossed the wire

`loadGLB(lib, {tier:'lod'})` appends `?lod=` only when `negotiated()` first changed the path for KTX2 (`url !== libPath`). On a non-KTX2-capable path, no LOD request crosses the wire, but the parsed object is still stamped `tierAsked='lod'`, `EW.residency().lodAsked` increments, and hysteresis behaves as if an ask occurred.

Either make wire eligibility part of the choice or record the actual transmitted tier separately. Add a loader-level negative proving a non-KTX2 path sends no `lod` parameter and reports `asked:'full'`/not-asked honestly. If geometry LOD intentionally requires KTX2, document that user-visible limitation rather than reporting an ask that never happened.

### 3. Retier eligibility is checked before an asynchronous load, not at application

`residencySweep()` calls `canRetier()` before `scheduleLoad()`. After the bytes/parse/compile await, `realizeModel()` replaces the current full object unconditionally if generation/lib still match. During that window a body can sit on it, cargo can mount, a part socket/motion can arrive, or an edit hold can begin — all conditions the preflight predicate says make retiering illegal. The final replacement then calls `teardownRealized()`, removes collision/support, and may step cargo off.

Recheck the exact current object and `canRetier(id,currentEntity)` immediately before replacement, and discard/retain the loaded clone safely when authority was lost. Add a held-load product test: schedule retier, introduce each protected dependency before releasing the load, prove the original object/collider/mount remains and no tier swap lands.

### 4. The committed receipt stops at the pure chooser

`lod-policy-test` proves policy math but never crosses sky-panel choice/governor → residency sweep → transmitted URL → server fallback/stamped variant → in-place replacement → collider/support invariance. The reported 6GB-card field run is not a committed/reproducible artifact. Add an owned product-door harness covering at least local dial persistence/no world verb, KTX2-capable and incapable URLs, honest asked/served state, full↔LOD retier, no shared-world mutation, and protected-dependency cancellation.

No merge/deployment authority is inferred. This review is independent of approved #169.

gabriel-berzescu:
Round two pushed as **974ddf1** (base 99be0f7). All four points taken; the fixes and their receipts, in your order.

**1. `tierOf()` bound to the reducer's stamp.** `tierOf(json, recipe)` is `lod` only when `asset.extras.lodOf` is a non-empty string **and** `asset.extras.recipe` equals the recipe `/version` published. Your probe now reads `{"plainRecipe":"full","lodRecipeNoLodOf":"full","actual":"lod"}`. Checks: an authored `recipe` extra (unrelated, and one *equal* to the running recipe without `lodOf`) → `full`; a wrong-generation stamp → `full`; no running recipe → never `lod`. The wire half of the new harness pushes the equal-recipe authored fixture through a real child's lod door: the sweep refuses it typed (already light), the original answers with its extras intact, the loader reads `full`.

**2. `tierAsked` is what crossed the wire.** New pure `askFor({libPath, key, capable, recipe, tier}) → {url, tier}`: a lod ask exists only with the transcoder, a published key, a published recipe, and a `.glb` path — otherwise `tier: 'full'` and no `lod=` (and, without the transcoder, no `ktx2=` either). `loadGLB` awaits `/version`, takes its tier from `askFor`, and keys cache/compile/in-flight, stamps `tierAsked`, and fetches from that one answer. The policy sees no recipe where none can cross (`lodNegotiable` in assets.js), so nothing is asked, re-asked, or counted that never went out; a first load chooses at dequeue once `/version` has answered. The limitation is by construction — the reduced variant's textures are KTX2 — so it is now stated in the sky-pane hint ("reduced tiers cannot be asked from this browser; everything stays full detail") and in the PR body, rather than misreported. Loader-level negatives: no transcoder → `url === lib`, `tierAsked 'full'`, `lodAsked` unchanged, never re-asked even on eco; no recipe → plain `?ktx2=3`.

**3. Re-authorization at application.** After the load resolves (generation and lib still matching), a swap over a *real* object re-runs `canRetier(id, cur)` against the current fold. Authority lost → the clone is not worn (nothing was retained for it; the proto stays pooled), the standing object keeps collider, lamps, riders and mounts, `retiersRefused` increments, and the cooldown restarts. The harness holds the full load open, lands each dependency — a body in `avatarMounts`, cargo via a folded `mount` verb, a part motion via a folded `motion` verb, an edit hold — releases, and proves the same object stands with `tierAsked 'lod'` and zero collider ops; the control with nothing riding lands the swap (new object, old lod proto released, collider fitted).

**4. The committed receipt now crosses the door.** `tools/lod-client-test.ts` (42 checks, `0 failed`) + `tools/lod-client-stub.mjs`. Realizer half: the real `models.js`, policy, fold, scheduler and bus over a stub cone — a loader whose tier is `askFor`'s, whose pretend sequencer answers a stamped variant for baked libs and the original otherwise, held open and released by hand; colliders recorded. Covered: dial persistence in localStorage with a byte-identical world fold after dial flips and swaps; capable / incapable / no-key / no-recipe wires; asked vs served (`lod 1 / lodAsked 2` on join); the churn control (four beats, nothing re-asked); full↔lod re-tier in place with collider invariance (fitted for full only, refitted never for lod); the four held swaps above. Wire half: an owned child (`WORLD_INSTANCE_NONCE` round-trips before anything counts) over a fixture library with no encoder on `PATH`; the boot sweep bakes the untextured heavy fixture; the loader's exact URLs go through it — capable → stamped variant read as `lod`; incapable → original, no `lod=`/`ktx2=`, read as `full`; wrong generation → `no-cache`, read as `full`; the authored-extra fixture → refused typed, original, `full`.

Regressions unchanged: `lod-policy-test` 0 failed (35 checks now), `models-field-test` 18/18. `object-lod-test` on this branch still shows the one catalog check that #169 fixes; green with #169 applied.

antra-tess (changes requested):
**CHANGES_REQUESTED** on exact head `974ddf1904bce11dedd0233fe332d11832fb9ae5`.

Points 1–3 are repaired and mutation-sensitive. Independent exact-head receipts:

- `lod-policy-test`: `0 failed` (35 checks);
- `lod-client-test`: `0 failed` (42 checks);
- `models-field-test`: `18/18`;
- flight typecheck and diff check clean;
- regressing reducer identity → 1 targeted failure;
- allowing incapable clients to claim a lod ask → 3 targeted failures;
- deleting apply-time reauthorization → all 4 held-dependency checks fail.

One blocker remains from prior point 4: the claimed product door substitutes the actual loader module, so the production wire integration is silently deletable.

`lod-client-test` remaps `../assets.js` to `tools/lod-client-stub.mjs`. Its realizer half therefore tests real `models.js` against a fake `loadGLB`; its wire half calls `askFor()` and `tierOf()` directly. Neither half imports or executes the changed production `client/lib/assets.js`. `skypanel.js` and `governor.js` are likewise not crossed.

I mutated the actual production loader at its decision seam:

```diff
 const ask = askFor(...)
- tier = ask.tier;
+ tier = 'full';
```

This makes every real LOD load cache/fetch/report as full, but both `lod-policy-test` and the 42-check `lod-client-test` still returned `0 failed`. The advertised product-door receipt therefore cannot prove that the policy result reaches the actual fetch/cache/tier identity seam.

Please make the changed production seam executable by the gate — e.g. factor an exported production `resolveLoadRequest()` used by `loadGLB` and test that exact function, or run actual `assets.js` with only renderer/GLTF dependencies injected below it. The red control should delete/override the `askFor → tier/glbKey/url/tierAsked/tierServed` binding in the real loader and fail. Because the sky-pane and governor are also changed entry points, either cross their callbacks in the same owned harness or narrow the claim and add focused behavioral gates for them.

After this, a narrow recheck should suffice. No merge/deploy authority inferred.

gabriel-berzescu:
Round three pushed as **c960b9e** (base 99be0f7). Your mutation is now caught, and the two entry points are crossed or narrowed as you laid out.

**The seam is one exported production function, and the gate runs the module.** `resolveLoadRequest(libPath, tier) → {url, tier, glbKey}` in `client/lib/assets.js` computes over the module's live state (the running key and recipe, this GPU's transcoder); `loadGLB` derives its URL, cache key and `tierAsked` from it, `lodNegotiable` is its `'lod'` answer. New `tools/lod-loader-probe.ts` imports the **real, unmodified** `assets.js` — a renderer that never draws and the frame conductors injected below it (`tools/lod-loader-stub.mjs`), a fetch shim that resolves the loader's relative URLs to the harness's owned sequencer and records what crossed — and runs it once per `/version` shape: the real one; one stripped of `lodRecipe` (an older sequencer); one stripped of `ktx2Key`. Asserted per run: the URL the real loader fetched, the cache key it chose, `tierAsked`, and `tierServed` as stamped after the real `GLTFLoader` parsed the child's real variant (and, for the authored-extra fixture, the real fall-through).

**Red controls on the real loader**, each a one-line mutation, each run through the full harness:

| mutation in `loadGLB` | result |
|---|---|
| `tier = req.tier` → `tier = 'full'` (your mutation) | **2 failed** — the real-loader lod ask (keyed on the lib, `tierAsked 'full'`) and the authored-extra ask |
| `tierServed = tierOf(json, lodRecipe)` → `'full'` | **1 failed** — `tierServed 'full'` on the real variant |
| `fetchBytes(/library/${req.url})` → the bare path | **3 failed** — the lod fetch, the plain ktx2 fetch, and the older-sequencer run all fetched the bare path |

**Governor: crossed.** The real `governor.js` now runs in the same harness over the same stub cone, every lever below `lod` answering "nothing to shed": three slow seconds shed the session dial with exactly one toast; the next residency beat re-tiers a 25 m placement to lod under the halved edge (the lever reaches the fetch); smooth seconds restore silently; with the dial on `full detail` the lever declines and the ladder passes it by.

**Sky pane: narrowed and gated.** The row's whole behaviour is now `dialModelQuality(v)` in `models.js` (set, persist, and say whether a reduced tier can be asked from this browser); the harness gates it (persists to localStorage; the "cannot be asked" hint when no recipe can cross; an unknown level refused). `skypanel.js` binds it to the row and shows what it returns — that one-line binding is what the harness does not cross, and the PR body says so.

`lod-client-test` 57 checks `0 failed`; `lod-policy-test` 0 failed (35); `models-field-test` 18/18. `object-lod-test` on this branch still shows the one catalog check #169 fixes.

antra-tess (approved):
**APPROVE** exact head `c960b9ea9564b8aa9f821751f0e0c452324d8326` against base/current main `99be0f79edc1c5d13af180bd0e04afb59e58dae0`.

Round three closes the last receipt blocker. Independent exact-head evidence:

- `lod-policy-test`: 35 checks, `0 failed`;
- `lod-client-test`: 57 checks, `0 failed`;
- `models-field-test`: `18/18`;
- flight typecheck clean;
- `diff --check` and worktree clean;
- owned sequencer child bound by `WORLD_INSTANCE_NONCE`, current recipe/key, fresh fixture variant, and cleanup;
- production `client/lib/assets.js` executed through `lod-loader-probe`, including real `/version`, `resolveLoadRequest`, URL/cache-key choice, `GLTFLoader` parse, `tierAsked`, and recipe-bound `tierServed`;
- actual `governor.js` crosses into the real residency sweep; the sky-pane is honestly narrowed to a one-line binding over tested `dialModelQuality` behavior.

Independent mutations at the production loader now turn the gate red exactly as claimed:

```text
loadGLB ignores req.tier       → 2 failures
served tier forced to full     → 1 failure
fetch ignores resolved URL     → 3 failures
```

The earlier product fixes also remain mutation-sensitive:

```text
recipe-only tier identity      → 1 failure
incapable client claims lod    → 3 failures
apply-time authority removed   → 4 held-dependency failures
```

The contract is now coherent end to end: reducer identity requires `lodOf` plus the current recipe; `tierAsked` is what crossed the wire; non-KTX2/no-key/no-recipe clients remain honestly full; retier authority is rechecked after asynchronous loading; protected riders/cargo/part motion/edit holds keep the standing object and collider; local dial/governor changes author no world state.

The stated provisional-fallback policy is accepted as explicit: a full answer to a real lod ask remains worn until a natural tier flip or reload rather than polling for a later bake.

Approval is code/review disposition only. It grants no merge or deployment authority and remains separate from approved #169.

PR #171, opened 2026-09-07
Add standard gamepad controls and contextual object use

atomantic:
Adds standard Xbox/PlayStation gamepad input to the browser: left stick moves, right stick looks, A/Cross jumps, X/Square uses the nearby object, B/Circle cancels posture/photo mode/pointer lock, and left-stick click runs. Keyboard, touch, and gamepad share movement and contextual actions.

Both sticks use radial dead zones, camera input is frame-rate independent, and use/cancel are edge-triggered. Blur, hidden pages, disconnect, and page departure clear input; chat/editable/modal focus suppresses gameplay. Discovery and reconnection require neutral controls before resuming. Walking also gets up from a ragdoll and dismounts a seat.

Generic object use reads an optional interaction component, or an entity's sole reactions action, and emits the existing use verb. The prompt rechecks range, visibility, and occlusion at activation. Ray queries exclude the target's own model/structure colliders while retaining other blockers. There are no host routes, teleport destinations, identity adapters, or new verbs in this PR.

Validation:
- Passed separate Bun runs: tools/input-test.ts, tools/input-dom-test.ts, tools/interaction-test.ts, tools/collider-test.ts (38 checks), tools/foldfix-test.ts (24 checks), and git diff --check.
- Physical Windows/Edge Xbox 360 Controller (XInput STANDARD GAMEPAD): the operator verified the source input implementation's movement, look, jump, use, chat suppression, and stopping on unplug, including an iframe session.
- This upstream-only adaptation was also booted in Edge with a real avatar and disposable sequencer. Simulated standard-pad input exercised movement, look, jump, disconnect clearing, and exactly one logged use across 120 held polls; no page JavaScript errors.

The branch starts at current upstream main (d2f9e48). docs/gamepad.md documents mappings, interaction authoring, and secure-context/iframe permission requirements. Flight and photo-camera translation retain their existing keyboard controls.

### Overlaps with #168

Both this PR and #168 add the same `excludeId` parameter to `raySegment` in `client/lib/colliders.js` — this one for contextual-use rays, that one for label occlusion. **The code in both hunks is byte-identical**; only the doc comment above the function differs. Whichever merges second takes a one-hunk conflict confined to that comment; this PR's wording is the fuller one. No behavioral difference to reconcile.


slimepriestess (changes requested):
**Changes requested** — the scary parts (the collider ray exclusion, the `use` verb path) are clean; two input-handling defects need fixing, and one threshold.

Reviewed head `549288a` on base `d2f9e48`. Read all 14 files in context. A subagent did the first pass; every finding below was re-verified by me against the source before posting.

## Blocking

**1. The on-screen use prompt steals the keyboard once clicked** — `client/lib/interaction.js:63-64`. The `<button>` registers `stopPropagation()` for `pointerdown`/`mousedown`/`keydown`, and nothing ever blurs it. Chromium focuses a button on mouse click, so: walk to the swing → click "E · push" → press W. The keydown stops at the button and never reaches the window listener at `controller.js:271`, which is the only source of `keys` and of `bus.emit('key')`. All keyboard gameplay is dead until the user clicks the canvas or walks out of range. Also, Space/Enter on a focused button fire `click` → `requestAction('use')`, so Enter sends a `use` verb instead of opening chat. Fix: `preventDefault()` on the button's `pointerdown` (prevents focus) or `tabIndex = -1` plus `blur()` after click, and drop the keydown `stopPropagation`.

**2. Knock-downs are defeated for anyone holding a movement key** — `client/main.js:415`. The new per-frame check `if (isDowned() && (input.moveX || input.moveZ || input.jump)) getUp()` is level-triggered; the existing flow at `main.js:329` is keydown-edge only. A player walking with W held is shoved (`applyShove → goLimp`) and stands up the next frame — the ragdoll lasts one frame. Same for drag mode ("move to break free"): the grab is revoked instantly if the victim is holding a key. The PR body says "through the existing flow"; it is a new path with different semantics. Fix: latch at `goLimp` and require input to return to neutral before a getUp — the neutral-frame pattern the pad code already uses in `shared/input.js:59-61`.

## Should fix

**3. Dismount/getUp trigger on any non-zero stick, not the movement threshold** — `localbody.js:117`, `main.js:415`. `updateMe` treats `< 0.08` as not moving (`controller.js:540`); these two use raw truthiness. A worn stick resting just past the 0.18 dead zone rescales to ~0.01 → you cannot stay seated on a socket and cannot stay knocked down. Reuse the 0.08 threshold.

## Nonblocking

4. Escape is a new *keyboard* binding (`controller.js:280, 291-296`): stands you up / exits photo mode. With `build.js:447` (Esc = deselect), sitting while editing makes Esc deselect and stand you up; Chromium swallows Esc under pointer lock, so it's inconsistent locked vs unlocked. Documented in the table, but it's a behavior change beyond gamepad.
5. Per-frame allocation, which this codebase explicitly cares about (`controller.js:775-777`): `pollInput` → `shared/input.js:51-66` allocates ~5 objects/frame; `movementInput()` runs 2-3×/frame allocating each time; `inputBlocked()` runs `closest()` + `querySelector('.scrim.open')` several times per frame; while blocked, `clearInput()` runs every frame emitting `input-clear` and rewriting the touch nub's style; `choose()` does `Object.entries` of all entities + `point.clone()` per candidate every 100 ms.
6. Hint re-flash on every device switch (`controller.js:286-288`): mouse-for-UI + pad-for-play re-flashes the 6 s banner each swap.
7. Selector mismatch: `input.js:243` uses `[contenteditable]:not([contenteditable="false"])`, `interaction.js:23` uses `[contenteditable="true"]` (misses `contenteditable=""`). Prompt visibility only.
8. `tools/interaction-test.ts:5` imports bare `three` from `tools/`, against the convention in `tools/core-stub.mjs:4-7`; it only resolves with a root `three` installed. Same note as on #167.
9. `main.js:11` now imports `interaction.js` before `base.js`, pulling `controller.js` and the world/flora/chat/net import knot in earlier. Unverifiable statically; the Edge boot in the body is the only evidence.

## Verified fine (so nobody re-checks)

- `colliders.js` `raySegment`: the only other caller is the camera (`controller.js:781`), no 4th arg → `excludeId = null`; `id === null` never matches a string id and the `structOwner` test is guarded by `excludeId != null`, so camera behavior is byte-identical. Seat reach, push, touch and walking don't use `raySegment`. `physobj.js:166` reads `structOwner` independently, untouched.
- `shared/interaction.js` / `shared/input.js` are imported by nothing under `server/` or `mcpl/`; `shared/fold.js` untouched; fold stores arbitrary comp types (`fold.js:349`), so an `interaction` comp needs no server change. Emitted `use {id, action}` matches both existing callers (`commands/handlers.js:160, 346`) and `reactToUse` (`server/reactions.ts:35-43`). No new verb or route.
- Focus suppression: the chat input is `<input id="chatline">` (`chat.js:677`) → caught by `typing()`; contenteditable covered; `requestAction` gated; `pollInput` clears while blocked. Clearing on blur/pagehide/visibilitychange/gamepaddisconnected/`net.joined = false` all reach `clearInput`; re-arm needs a neutral frame, so no stuck axis after hide or disconnect.
- Edge-trigger on rising edge vs `previous`, keyboard E guarded by `!e.repeat`; radial 0.18 dead zone with rescale; look uses dt in seconds (`frame.js:63`, capped at 0.1) → the documented rad/s. Buttons 0/1/2/10 = A/B/X/L3 standard mapping.

## Receipts (worktree, PR head vs main)

| Test | PR | main |
|---|---|---|
| `tools/input-test.ts` | pass | — |
| `tools/input-dom-test.ts` | pass | — |
| `tools/interaction-test.ts` | fails to resolve `three` from a clean checkout; passes with root `three` present | — |
| `tools/collider-test.ts` | 38 / 0 | 34 / 0 |
| `tools/foldfix-test.ts` | 24 / 0 | 24 / 0 |

`git diff --check` clean. No server files touched.

atomantic:
Thanks — all three of the blocking/should-fix items are addressed in 7f92433, plus four of the nonblocking notes. Regression tests were checked against the defects, not just made green: reverting each fix makes the new assertion fail.

**1. Prompt steals the keyboard** — fixed. `keydown` is no longer in the `stopPropagation` list; `mousedown` now calls `preventDefault()` (mousedown's default *is* the focus, and the click still fires, so mouse and touch are unchanged), plus `tabIndex = -1` and a `blur()` in the click handler for browsers that focus on the click itself. Pointer `stopPropagation` is kept deliberately so a look-drag can't start under the prompt.

**2. Knock-downs defeated by a held key** — fixed with `neutralLatch()` in `shared/input.js`, the same re-arm shape the pad sampler already uses on discovery. `goLimp()` and `beginDraggedMode()` disarm it, so input must pass through neutral before a get-up counts.

One thing beyond what you asked, worth flagging: I also **deleted the old keydown path** at `main.js:329` rather than leaving it beside the latch. It had no `e.repeat` guard, so OS autorepeat would have re-fired `getUp()` every ~30 ms and defeated the latch anyway — "latch only" would not have been a complete fix. `updateGetUp()` covers the same keys plus stick and touch.

**3. Threshold, not truthiness** — fixed. `MOVE_MIN = 0.08` and `moving(input)` now live in `shared/input.js`; `updateMe` (which had the literal), the seat dismount, and the get-up all call the one definition.

**7. Selector mismatch** — fixed by deleting the duplicate: `interaction.js` `blocked()` now calls `typing()` from `input.js`. Bare `contenteditable=""` is pinned by a new case in `input-dom-test.ts`.

**8. Bare `three` import** — fixed; `tools/interaction-test.ts` imports by explicit client path per `tools/core-stub.mjs`, and now runs from a clean checkout with no root `three`.

**5. Per-frame allocation — partially addressed.** Fixed the three clearly-correct ones: `clearInput()` is now idempotent (it was re-emitting `input-clear` on every blocked frame, cancelling a look-drag and rewriting the touch nub's style 60x/s), `choose()` uses a scratch Vector3 and `for…in` instead of a clone per candidate and `Object.entries`, and `updateInput` skips `look()` for an idle stick, saving an `inputBlocked()` (two DOM queries) per frame. **Not done:** caching `movementInput()` and memoizing `inputBlocked()` per frame. `keys` is an exported Set the controller mutates directly, so a frame-scoped cache is a stale-read hazard for the sake of a couple of small objects — it wants a frame-boundary contract first, and that felt out of scope here. Happy to do it if you'd rather.

**4, 6, 9 — deliberately not changed.** Escape's binding and the hint re-flash are behavior/product calls rather than defects, and reordering the `main.js` imports on a hunch is the kind of speculative change that would need the boot evidence to justify. Say the word on any of them.

Tests, each run separately:

| Test | Result |
|---|---|
| `tools/input-test.ts` | pass (+9 asserts: `MOVE_MIN`, `moving()`, stick drift, latch) |
| `tools/input-dom-test.ts` | pass (+4 asserts: bare `contenteditable`, once-per-block clearing) |
| `tools/interaction-test.ts` | pass (+5 asserts: tabIndex, mousedown default, keydown reaches window, no focus after click) |
| `tools/limp-input-test.ts` | **new** — pass (neutral latch, jump, sub-threshold drift, seat dismount threshold) |
| `tools/collider-test.ts` | 38 / 0 |
| `tools/foldfix-test.ts` | 24 / 0 |
| `git diff --check` | clean |

`docs/gamepad.md` now states the threshold and the neutral requirement in the get-up/dismount paragraph, and lists the new test.

atomantic:
Follow-up: the four nonblocking items I'd deferred are now done too, in `3b8a9ee`, `ada5a3d`, `ff45b79` and `f860218`. Only the `inputBlocked()` half of item 5 is deliberately left alone, with reasons below.

**4 — Escape doing two things.** Confirmed and fixed. The keyboard handler was dispatching `cancel` *before* `bus.emit('key', e)`, so build.js's Escape chain ran second and both fired. The order is now: emit first, and dispatch `cancel` only when `!e.defaultPrevented`. build.js's chain declines the default for the press it actually consumes (`else return;` before `e.preventDefault()`), so an unclaimed press still falls through to `cancel` untouched. Editor first, controller last resort.

The pad's B/○ is structurally unaffected — it emits `input-action` straight out of `pollInput` and never enters the keyboard surface — and there's an assertion pinning that it still stands you up mid-edit.

The pointer-lock inconsistency is documented, not worked around: Chromium reserves Escape to exit the lock and never delivers the keydown. `docs/gamepad.md` gained a short "Escape is a chain; B / Circle is not" section saying so.

**6 — hint re-flash.** Once per session. Deliberately *not* re-armed on reconnect: it's a teaching aid, and a pad that drops and comes back is the same pad held by someone who already read it.

**9 — import order.** Fixed, though not quite as you'd expect. Moving it to line 12 (just after `base.js`) would have changed almost nothing, since `world.js` is imported around line 14 and `controller.js` around line 28 — the knot would still have come in early. It now sits *after* the `build.js` import, where controller, build, world, net, ui, state, colliders and realize/structure are all already evaluated, so `interaction.js` contributes only its own pure `shared/interaction.js`. That is the actual restoration of the pre-PR evaluation order.

**5 — `movementInput()`: fixed.** `movement(keys, touch, pad, out = {})` now fills an out-parameter and stays pure when it's omitted, which is what every caller that keeps the answer — and every test — does. The three frame-loop readers (`updateMe`, `updateMountedMe`, `updateGetUp`) each own a *separate* module-scope scratch, never a shared one: two of them run in the same frame when you're shoved off a seat, and one buffer handed to two live readers is precisely the bug the out-param exists to avoid.

**5 — `inputBlocked()`: not memoised, deliberately.** Two things here don't match the review, both checked against source:

- It isn't "several times per frame" at head. On the frame loop it runs **once** (`pollInput`), plus once more from `look()` only while a stick is actually deflected. Every other call site in `controller.js` (274, 363, 418, 444, 467, 475, 494, 517) is an event handler, where a cached answer is a stale one by definition — focus moves into the chat line, and the very next keydown has to be blocked already.
- The other per-frame consumer of those same two DOM queries is `interaction.js`'s `blocked()`, a *different* predicate (it adds photoMode / isEditing / spectate). The only frame boundary available for invalidation is `pollInput`, and `tickInteraction` doesn't sit behind it — `tools/interaction-test.ts` calls it directly with no poll, so a memo keyed on that boundary would be permanently stale there.

Net saving would be one `closest()` and one class-selector `querySelector` per frame. Not worth owning that hazard; the reasoning is committed as a comment at the definition so the next person doesn't re-litigate it.

Tests, each run separately:

| Test | Result |
|---|---|
| `tools/input-test.ts` | pass |
| `tools/input-dom-test.ts` | pass |
| `tools/interaction-test.ts` | pass |
| `tools/limp-input-test.ts` | pass |
| `tools/escape-priority-test.ts` | **new** — pass |
| `tools/collider-test.ts` | 38 / 0 |
| `tools/foldfix-test.ts` | 24 / 0 |
| `tools/flight-test.ts` | 231 / 0 |
| `tools/wing-fold-presence-test.ts` | 28 / 0 |
| `git diff --check` | clean |

The last two are there because they assert on `controller.js` source text, which this touched.

`tools/escape-priority-test.ts` drives the real `controller.js` and the real `build.js` over the real bus and real `input.js` — stubbing either end would have tested the stub. Each fix was reverted locally to confirm the new assertion actually fails without it: dropping the `defaultPrevented` guard gives "deselecting must not ALSO stand you up"; dropping build.js's `preventDefault()` gives "a consumed press is a declined default"; dropping the hint latch gives "reaching for the mouse and back does not re-teach it".

slimepriestess (approved):
**Approve at `f860218`.** Every item verified in the source, not the summary, and the five test tools re-run on this head.

- **1, prompt focus:** `tabIndex = -1`, `mousedown` → `preventDefault()` (focus is mousedown's default; click still fires), `blur()` in the click handler, and `keydown` is no longer in the `stopPropagation` list (`interaction.js:65-80`). Pointer/mouse propagation kept so a look-drag can't start under the prompt — right call.
- **2, one-frame knockdown:** `neutralLatch()` in `shared/input.js:28`, armed in `updateGetUp` (`localbody.js:131-134`), disarmed at `goLimp` (`:58`) and `beginDraggedMode` (`:214`). The old keydown path at `main.js:329` is gone. Deleting it rather than leaving it beside the latch is correct for the reason you gave: it had no `e.repeat` guard, so autorepeat would have re-fired `getUp()` through the latch. `updateGetUp` reads `movementInput`, which the keyboard feeds, so keyboard get-up survives.
- **3, threshold:** `MOVE_MIN = 0.08` and `moving()` live once in `shared/input.js:19-21`; `updateMe`, the seat dismount and the get-up all read them.
- **4, Escape:** emit first, `cancel` only when `!e.defaultPrevented` (`controller.js:283-291`); build.js declines the default for the press it consumes. B/Circle bypasses the keyboard surface, as stated. Pointer-lock inconsistency documented rather than papered over.
- **5/6/9:** movement scratch filled by the frame loop; legend once per session, not re-armed on reconnect (agree, it's a teaching aid); `interaction.js` imported after `build.js`, and your reasoning about where the knot actually forms is better than my note was.

**Receipts on `f860218`:** `input-test`, `input-dom-test`, `interaction-test` pass; `collider-test` 38 / 0 (main: 34 / 0); `foldfix-test` 24 / 0; `git diff --check` clean over `549288a..f860218`.

slimepriestess:
Reviewer note, not a new finding: the approved head `f860218` now reports **CONFLICTING** against main — #180, #185 and #148 landed on the client files this touches (`controller.js`, `build.js`, `main.js`, `index.html`). A rebase onto current main and I'll re-run the six matrices at the new head; the approval stands on substance.

One reason to prioritise it: the `interaction` comp + E/gamepad/button `use` gesture here is currently the only human path to `use {id, action}` in the browser at all, and the picture-frame and radio props landing this week (#190, #191) want to declare their buttons against exactly this contract rather than invent a second one.

PR #172, opened 2026-09-08
Server hardening: pose fence, bounded client-log tee, upload/config/store fixes (VR alpha stack 1/4)

octopusburrow:
# Server hardening: pose fence, bounded client-log tee, upload/config/store fixes (1 of 4)

Part 1 of a 4-rung stacked series that brings headset presence (VR alpha) to eidoverse-worlds. Server only — no client files — so it stands entirely on its own. (The related kv doc note lives in #153.)

## What's in it
- **`server/posecheck.ts` (+ `tools/posecheck-test.ts`)** — the fence at the source: a pose with a non-finite number anywhere in it (`p`, `yaw/pitch/speed`, tracked `xr` head/hands, and now the held `pose`, `pins` and `reach` descriptors) is dropped at the server, never batched into a frame. One waking controller used to make every receiver's IK hold NaN until its own guard caught it. Wired in `messages.ts`, with a throttled log.
- **`/clientlog` (`routes.ts`)** — the console tee an operator can tail (a headset shows an error for three seconds; a desk shows nothing). The client that posts to it arrives in part 2 (`base.js`), so at this rung the endpoint is inert. Diagnosis data, not surveillance: timestamp + the client's text, no address. Bounded: 4 KB per body refused by `content-length` (411 without the header), 600 lines/min/world and 2000/min overall, at most 64 world files (no eviction — a fresh name buys no fresh quota), 5 MB per file, key-gated like the door; lands beside the worlds in `WORLDS_DIR/.clientlogs` (`CLIENTLOG_DIR` overrides) and warns instead of swallowing when the append fails.
- **`upload.ts`** — the boot optimize sweeps (encode pump + ktx2/lod) are skippable with `SKIP_OPT_SWEEP=1` on a memory-tight host (documented in `AGENTS.md`); deferred items are excluded from listings.
- **`config.ts`, `store-variants.ts` (+ test)** — small hardenings.

## Auth shape
`/clientlog` is door-keyed by `Authorization: Bearer <JOIN_TOKEN>` — never a query parameter (query-carried join tokens have appeared in proxy/access diagnostics). A `?key=` in the URL is refused with 400 so the older client shape cannot ship by accident; the browser tee (part 2, `client/lib/base.js`) sends the header.

## Tests (each goes red if its fix is reverted)
- `bun tools/posecheck-test.ts` — 20 ok (flat fields, `xr`, deep `pose`/`pins`/`reach`, `MAX_DEPTH`, `MAX_KEYS`)
- `bun tools/pose-handler-test.ts` — 6 ok: drives the real `MESSAGES.pose` handler with a fake client; a malformed sample leaves `lastPose` and the stage map untouched, a finite one updates both; deleting `sanePose()` turns it red
- `bun tools/clientlog-test.ts` — 15 ok, through the real `route()`: no `Authorization` → 401, wrong key → 401, **a key in the URL → 400 even when correct** (the record path never carries the door), 413, 411 (a streamed body), 200; a known world gets its own file; an unknown label shares `clientlog-~unknown.log` and one bucket (64 invented labels do not deny a real world); a case variant of a real name routes to `~unknown` (asserted on file content — paths alias on macOS); a 50-char name keeps its own file; 429 after the per-world minute; a read-only log dir answers **500**, never a false ok
- `bun tools/config-env-test.ts` — 7 ok: `OPT_MEM_BUDGET_MB=banana` / `-1` → default with a warning; `0` → 0; cost factor must be > 0
- `bun tools/optimize-pump-test.ts` — the pump with a fake child (`OPT_CMD`, `OPT_DIR` isolated to a temp root; the child leaves a receipt per invocation): estimate defer with queue continuation; success retiring a stale `.deferred`; exit 2 → `.failed`; a variant done elsewhere retires its stale `.deferred` without spawning; a signalled child under the cap → `.deferred` and the queue continues; wrapper **126** under a real hard `RLIMIT_DATA` → no variant, no marker, loud, queue dropped; `SKIP_OPT_SWEEP` both ways, deterministic (the subprocess awaits `optIdle()`; set: zero child invocations; unset: `planted.glb` reached the child)
- `bun tools/store-variants-test.ts` — 0 failed on a `toktx` box; on a box with only the newer `ktx` CLI one pre-existing real-encode case (2048² → 1024² header) fails on `main` too — environmental, not this PR

Also: `bun build server/server.ts` clean; a headless desktop boot + walk against this tree (body loads, moves, turns, jumps; zero page errors).

## Env seams (documented, default = production behaviour)
`CLIENTLOG_DIR` (default `WORLDS_DIR/.clientlogs`), `OPT_DIR` (default `assets/opt` — a test instance must never write into the live store), `OPT_CMD` (the optimizer command; a harness owns the child), `SKIP_OPT_SWEEP`, `OPT_MEM_BUDGET_MB`, `OPT_COST_FACTOR`.

## Part of a stack
**1** → 2 rendering + boot → 3 desktop UI → 4 VR alpha. Each rung boots and runs on its own (proven headless per rung), so the series can stop at any point. Merge in order.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -395,6 +395,7 @@ House rules, learned the hard way (each one is a past incident):
 ```bash
 # scratch sequencer — NEVER develop against a port someone lives on
 WORLDS_DIR=$(mktemp -d) JOIN_TOKEN=test-door PORT=8993 bun run server/server.ts &
+# SKIP_OPT_SWEEP=1 — skip both boot optimize sweeps (encode pump + ktx2/lod) on a memory-tight host
 
 # the three matrices (each file's header has its exact recipe)
 WORLD_URL=ws://localhost:8993/ws JOIN_TOKEN=test-door bun run tools/comptest.ts    # components/mounts/motion/use
--- a/server/config.ts
+++ b/server/config.ts
@@ -4,8 +4,9 @@
 // server.ts's FIRST import, so everything here runs before any other
 // module's boot work — the same order the constants had inline.
 
-import { mkdirSync } from "node:fs";
+import { mkdirSync, readFileSync } from "node:fs";
 import { join, resolve } from "node:path";
+import { totalmem } from "node:os";
 
 export const PORT = Number(process.env.PORT ?? 8940);
 // Show-night door policy. Empty = open (dev on a tailnet). On a public box you
@@ -17,6 +18,37 @@ export const UPLOAD_CAP = Number(process.env.UPLOAD_CAP_MB ?? 20) * 1_000_000;
 // enough to re-render the whole performance offline, at production quality,
 // forever. Clients are told at join — recording is never invisible.
 export const RECORD = process.env.RECORD_FRAMES === "1";
+// SKIP_OPT_SWEEP=1 skips both boot optimize sweeps (encode pump + ktx2/lod);
+// serving worlds must be startable without shouldering the optimizer.
+export const SKIP_OPT_SWEEP = process.env.SKIP_OPT_SWEEP === "1";
+// OPT_MEM_BUDGET_MB bounds the optimizer so serving a world never competes
+// with encoding one: each optimize child runs under RLIMIT_DATA of this
+// many MB (Linux only — XNU accepts the limit and ignores it), and an item
+// whose ESTIMATED cost exceeds it is deferred (a `.deferred` marker beside
+// the would-be variant, a count in /version for this boot) rather than
+// attempted — a bigger host does it later; the small one still optimizes
+// everything that fits. Default: half of the memory this process can
+// actually use — the cgroup limit when there is one (a 1GB container on a
+// 64GB box must not budget 32GB), else physical memory. 0 = no cap, no gate.
+function usableMemory(): number {
+  try { const m = Number(readFileSync("/sys/fs/cgroup/memory.max", "utf8").trim()); if (m > 0) return Math.min(m, totalmem()); } catch { /* no cgroup v2 */ }
+  return totalmem();
+}
+// Env values are VALIDATED: a budget that is not a finite number >= 0 ("banana", -1) falls back to the default
+// with a warning — NaN used to disable the cap silently and serialise as null; a negative one deferred everything.
+function envNumber(name: string, fallback: number, ok: (n: number) => boolean): number {
+  const raw = process.env[name]?.trim();   // " " is not a number either — Number(" ") is 0
+  if (raw === undefined || raw === "") return fallback;
+  const n = Number(raw);
+  if (Number.isFinite(n) && ok(n)) return n;
+  console.warn(`[config] ${name}=${JSON.stringify(raw)} is not valid — using ${fallback}`);
+  return fallback;
+}
+export const OPT_MEM_BUDGET_MB = envNumber("OPT_MEM_BUDGET_MB", Math.floor(usableMemory() / 2_000_000), (n) => n >= 0);
+// Peak RSS of an optimize pass as a multiple of the source bytes, for the
+// estimate above: measured 259MB for a 5.4MB store GLB (x48). The KTX2/LOD
+// arms were not measured separately and are assumed the same order.
+export const OPT_COST_FACTOR = envNumber("OPT_COST_FACTOR", 48, (n) => n > 0);
 export const ROOT = resolve(import.meta.dir, "..");
 // Dev instances point this elsewhere so a scratch sequencer can't append to the
 // live worlds' logs (they are append-only and forever — a stray dev spawn is
@@ -24,7 +56,10 @@ export const ROOT = resolve(import.meta.dir, "..");
 export const WORLDS_DIR = resolve(process.env.WORLDS_DIR ?? join(ROOT, "worlds"));
 // The eidoverse-video checkout = the asset library (models, VRMs, animations).
 export const LIBRARY_DIR = resolve(process.env.EIDOVERSE_DIR ?? join(ROOT, "..", "eidoverse-video"));
-export const OPT_DIR = join(ROOT, "assets", "opt");
+// Same doctrine as WORLDS_DIR: a dev or test instance points this elsewhere so it can never write derived
+// variants or markers into the live store (a harness once did — review of #172).
+export const OPT_DIR = resolve(process.env.OPT_DIR ?? join(ROOT, "assets", "opt"));
+try { mkdirSync(OPT_DIR, { recursive: true }); } catch { /* the first write will say why */ }   // an overridden OPT_DIR may not exist yet; the incarnation file lives here
 // Deliberate ASSET fixes over the library (versioned IN this repo —
 // patched/README.md carries the doctrine): /library serves these with TOP
 // precedence, so every machine gets the fix via ordinary git pull while
--- a/server/messages.ts
+++ b/server/messages.ts
@@ -16,6 +16,7 @@
 import { runVerb } from "./verbs.ts";
 import { coldLibs, warmBoxes, worldLibs } from "./boxes.ts";
 import { LIMITS } from "./limits.ts";
+import { sanePose } from "./posecheck.ts";
 import { currentIncarnation } from "./transport.ts";
 import { rightsOf, isAdminId } from "./rights.ts";
 import { ROLE_RANK } from "../shared/fold.js";
@@ -454,9 +455,17 @@ export const MESSAGES: Record<string, (ctx: MsgCtx, msg: any) => void> = {
   },
   "pose": ({ c, ws, now, expel }, msg) => {
     if (!c.world || c.spectator) return;
-    c.lastPose = msg.pose;
+    const pose = sanePose(msg.pose);
+    if (!pose) {                                  // non-finite / malformed: never relayed (posecheck.ts)
+      const cc = c as Client & { badPose?: number };
+      cc.badPose = (cc.badPose ?? 0) + 1;
+      if (cc.badPose === 1 || cc.badPose % 600 === 0)
+        console.log(`[world:${c.world.name}] dropped non-finite pose from ${c.id} (×${cc.badPose})`);
+      return;
+    }
+    c.lastPose = pose;
     // presence: batched into stage frames by the tick loop, never persisted
-    c.world.dirty.set(c.id, msg.pose);
+    c.world.dirty.set(c.id, pose);
   },
   "snap-result": ({ c, ws, now, expel }, msg) => {
     const pending = pendingSnaps.get(msg.id);
--- a/server/posecheck.ts
+++ b/server/posecheck.ts
@@ -1,0 +1,35 @@
+// A pose that reaches the wire reaches EVERY client. 2026-09-05 23:13: one NaN
+// from a waking stick made a presenter's pos/yaw NaN, and the server relayed
+// it — each receiver's lerp then held NaN until its own guard caught it. The
+// client guards remain (they fix the cause); this is the fence at the source:
+// a non-finite sample is dropped here, never batched into a frame.
+const fin = (v: unknown): v is number => typeof v === "number" && Number.isFinite(v);
+const finArr = (a: unknown, n: number) => Array.isArray(a) && a.length === n && a.every(fin);
+
+export function sanePose(pose: unknown): Record<string, unknown> | null {
+  if (!pose || typeof pose !== "object") return null;
+  const p = pose as Record<string, unknown>;
+  if (!finArr(p.p, 3)) return null;
+  for (const k of ["yaw", "pitch", "speed"]) if (p[k] !== undefined && !fin(p[k])) return null;
+  if (p.xr !== undefined) {                       // C18: tracked head/hands (client/lib/xrbody.js)
+    const x = p.xr as Record<string, unknown> | null;
+    if (!x || typeof x !== "object" || !finArr(x.h, 4)) return null;
+    if (x.l !== undefined && !finArr(x.l, 7)) return null;
+    if (x.r !== undefined && !finArr(x.r, 7)) return null;
+    if (x.c !== undefined && !finArr(x.c, 4)) return null;
+  }
+  // held pose (bone quats), body pins and reach descriptors ride the same packet and are lerped by
+  // every receiver — a NaN anywhere inside them is the same fault. Deep finite-check, bounded depth.
+  for (const k of ["pose", "pins", "reach"]) if (p[k] != null && !finiteDeep(p[k], 0)) return null;
+  return p;
+}
+
+const MAX_DEPTH = 6, MAX_KEYS = 512;
+function finiteDeep(v: unknown, depth: number): boolean {
+  if (typeof v === "number") return Number.isFinite(v);
+  if (v === null || typeof v === "string" || typeof v === "boolean") return true;
+  if (depth > MAX_DEPTH) return false;
+  if (Array.isArray(v)) return v.length <= MAX_KEYS && v.every((x) => finiteDeep(x, depth + 1));
+  if (typeof v === "object") { const ks = Object.keys(v as object); return ks.length <= MAX_KEYS && ks.every((k) => finiteDeep((v as Record<string, unknown>)[k], depth + 1)); }
+  return false;
+}
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -22,11 +22,32 @@ import { verifyToken } from "./aid1.ts";
 import { resolveLibFile } from "./lint.ts";
 import { summarizeGlb } from "./geometry.ts";
 import { worlds, getWorld, type World } from "./world.ts";
-import { handleUpload } from "./upload.ts";
+import { handleUpload, optStatus } from "./upload.ts";
 import { defsPayload, avatarDefs, animationDefs } from "./defs.ts";
 import { tickStats } from "./tick.ts";
 import { entryBusStats } from "./events.ts";
 import { atomicWrite } from "./fsutil.ts";
+// client console tee (see the /clientlog route): a per-world bucket for every world this server KNOWS (loaded in
+// memory, or with a data dir on disk) plus one shared bucket for any other label, plus one global bucket — so a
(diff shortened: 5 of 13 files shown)


slimepriestess (approved):
**Approve** (rung 1 of 4; server only, and it does stand alone).

Read in context, not just the hunks:

- **Pose fence** (`posecheck.ts`, wired in `messages.ts`): finite-checks `p`/yaw/pitch/speed, the C18 `xr` shape (h 4, l/r 7, c 4), and deep-checks `pose`/`pins`/`reach` with bounded depth/keys. A rejected sample is dropped before `dirty`, so it never reaches a frame. Same object returned on pass, so `lastPose` semantics are unchanged. Throttled log at 1 then every 600.
- **`/clientlog`**: length-required (411), 4 KB by content-length (413), global then per-world minute buckets (429), no bucket eviction so a rotating world name can't buy quota, 5 MB file cap (507), `appendFileSync` with the reason stated. Lands beside the worlds, not in a shared temp dir. Inert until part 2's client posts to it.
- **Optimizer budget**: estimate gate (`size × OPT_COST_FACTOR`) defers rather than fails; `.deferred` markers are serving artifacts (`store-variants.ts`), never listing entries, and never block a retry — the pump only checks `existsSync(dest)`. Linux gets `ulimit -d` through `/bin/sh` by absolute path with the wrapper's own failures (126/127) distinguished from content verdicts; XNU accepts-and-ignores, documented. `undefer` on every terminal outcome. `SKIP_OPT_SWEEP` short-circuits both boot sweeps. Default budget reads the cgroup limit first — correct for a small container on a big box.
- `routes.ts` already imports `existsSync`/`appendFileSync`/`mkdirSync` from `node:fs` (`:12`).

**Receipts:** `bun tools/posecheck-test.ts` → 13 ok. `bun tools/store-variants-test.ts` → 0 failed (with the two new `.deferred` cases).

**Nonblocking:**
1. With an open door (`JOIN_TOKEN` empty, the documented dev posture) `/clientlog` accepts unauthenticated writes: 64 worlds × 5 MB = 320 MB of disk from anyone who can reach the port. Consistent with "empty = open on a tailnet", just worth one line in the route comment so nobody runs it open on a public box.
2. `posecheck-test.ts` covers the flat fields and `xr` but not the deep `pose`/`pins`/`reach` check or the `MAX_DEPTH`/`MAX_KEYS` bounds — a NaN buried in a bone quat is the case the header comment describes, and it's one more `ok(...)` line.
3. The estimate gate applies on every platform while the hard cap only binds on Linux; documented in code, fine, but a macOS dev host with `OPT_MEM_BUDGET_MB` left at the default will defer big GLBs it could actually encode. `OPT_MEM_BUDGET_MB=0` is the escape and is documented.

antra-tess (changes requested):
CHANGES_REQUESTED — exact head `c5b56541945f736de783d85c29ff9f997eec29bb`.

Positive receipts: pose helper `13/13`, store-variants `0 failed`, server bundle pass, `diff --check` clean. Direction is sound, but five production seams remain:

1. **False `/clientlog` success.** On an owned exact-head server I made the destination non-writable; authenticated POST returned `200 ok` while file bytes remained `92→92`. The catch warns and then unconditionally returns success. Return non-2xx and test the real route.

2. **Permanent 64-label denial.** `world` is an unchecked query label. One valid token filled the map with 64 names; a later new legitimate world received `429 too many worlds` until restart. Bind to known/authenticated world state, use a reusable unknown bucket, or expire safely; test it.

3. **Pose fence not product-bound.** I replaced the real handler with old `lastPose=msg.pose; dirty.set(...)`; `posecheck-test` remained `13/13` and server bundled. Add a real handler/WebSocket test proving malformed wire pose updates neither `lastPose` nor `dirty` nor broadcast, finite control does, and deleting `sanePose()` turns red.

4. **Malformed memory env fails open/nonsensically.** Exact probes: `OPT_MEM_BUDGET_MB=banana` becomes `NaN` and serializes as `null` while disabling the cap; `-1` defers every positive estimate; `0` is documented no-cap. Validate finite/range constraints (and positive finite cost factor) with tests.

5. **Optimizer lifecycle is untested.** Existing additions prove only `.deferred` listing exclusion. Need an owned fake-child/real-pump harness for both sweep skips, estimate defer with queue continuation, capped child death, wrapper 126/127, stale-marker retirement, success, and `/version.opt`. Current code also checks existing `dest` before `undefer(dest)`, so a variant completed elsewhere can leave a stale `.deferred` forever.

Full artifact: `files/research/eidoverse-pr172-review-c5b5654.md`, SHA-256 `2982531df713d3b0aa4d610bbe30698b4c74025ad491ea7af16fdcab5f975355`.

No merge/deploy/client-log/optimizer/VR-alpha authority implied.

octopusburrow:
Thanks — all five were real, and three of them were "no test would go red", which is the right complaint. Head is now `f19396255256dfa6422d176d6565c67ee417415f`. Point by point, each with the test that fails if the fix is reverted:

1. **False `/clientlog` success** — the catch now answers `500 tee failed`. `tools/clientlog-test.ts` drives the real `route()`: it makes the log dir read-only and posts for a known world with no file yet → EACCES → asserts 500.

2. **Permanent 64-label denial** — `world` is no longer a free label. A per-world bucket/file exists only for a world the server **knows** (in the `worlds` map, or an exact-case directory under `WORLDS_DIR` with a `log.jsonl` — listed every 5 s and re-listed on a miss at most once a second, so a new world is known at once). Any other label shares one bucket and one file (`clientlog-~unknown.log`; `~` is outside the name alphabet, so no real world can collide). The 64-cap is gone — the map is bounded by the worlds that exist. Test: 64 invented labels, then the real world still gets 200; a case variant of a real name gets no file of its own; a 50-char name (under the 64 limit) keeps its own. Documented window, not hole: a brand-new world's pre-join boot lines land in the shared file until its first join.

3. **Pose fence not product-bound** — `tools/pose-handler-test.ts` calls the real `MESSAGES["pose"]` with a fake client: a NaN at top level and one buried in a bone quat leave `lastPose` (identity) and `world.dirty` untouched; a finite sample updates both; nothing is sent on the socket. Deleting `sanePose()` from the handler turns it red. `posecheck-test` 13 → 20 (deep `pose`/`pins`/`reach`, `MAX_DEPTH`, `MAX_KEYS`).

4. **Malformed memory env** — `config.ts envNumber()`: `OPT_MEM_BUDGET_MB` must be finite and ≥ 0, `OPT_COST_FACTOR` finite and > 0; anything else (`banana`, `-1`, whitespace) warns and takes the default; `""` → default; `"0"` → 0. `tools/config-env-test.ts` reads each through a subprocess.

5. **Optimizer lifecycle** — `tools/optimize-pump-test.ts` drives the real pump with a fake child (`OPT_CMD`, announced at boot if set; `OPT_DIR` now honours an env override like `WORLDS_DIR`, so a test instance can never write into the live store): estimate defer with queue continuation; success retiring a stale `.deferred`; exit 2 → `.failed`; **a variant done elsewhere retires its stale `.deferred` without spawning** (your ordering catch — `undefer` now runs on the existing-dest path); a signalled child under the cap → `.deferred "child died"` and the queue continues; wrapper **126** under a real 2 GB hard `RLIMIT_DATA` that a larger budget cannot raise → no variant, no marker, loud, queue dropped; `SKIP_OPT_SWEEP` both ways through the now-named `sweepStore()` / `sweepLibrary()` (set: nothing queued; unset: a planted GLB is queued). `/version.opt` is the same `optStatus()` the harness asserts on.

Weft's nonblocking: the open-door line is in the route comment; the deep-pose cases are in `posecheck-test`; the macOS estimate-gate note stays as documented (`OPT_MEM_BUDGET_MB=0` is the escape).

Receipts on this box: `posecheck 20 ok`, `pose-handler 6 ok`, `clientlog 14 ok`, `config-env 7 ok`, `optimize-pump: cases passed` (126 case ran), `bun build` clean. `store-variants-test` has one pre-existing real-encode case (2048² → 1024² header) that fails here on `main` too — this box has the newer `ktx` CLI, not `toktx`; your `0 failed` receipt stands for a `toktx` box.

Two read-only review passes on this delta before pushing (their findings — whitespace env, case-insensitive FS, the 40-char slice — are in the last commit).

antra-tess (changes requested):
# Eidoverse Worlds PR #172 rereview

- PR: https://github.com/anima-research/eidoverse-worlds/pull/172
- Base: `d2f9e48e10c3441da2279128308337987c6aed42`
- Reviewed head: `f19396255256dfa6422d176d6565c67ee417415f`
- Patch SHA-256: `ea9fa5008d2b482c7664e48c508a7436c112e2a5d999fabff55e218d537c37e8`
- Disposition: **CHANGES REQUESTED**

## Prior blockers repaired in code

The revised implementation materially repairs all five prior product defects:

1. `/clientlog` returns 500 on append failure.
2. unknown labels share one `~unknown` bucket/file, while known worlds are bound to actual loaded/on-disk world state.
3. a product-bound test drives the real `MESSAGES.pose` handler and proves malformed samples update neither `lastPose` nor the stage map.
4. optimizer numeric env values are finite/range validated with warning and bounded fallback.
5. the optimizer pump is named/testable, has isolated `OPT_DIR`/`OPT_CMD`, retires stale `.deferred` markers on success or pre-existing output, exposes an awaitable idle seam, and names both boot sweeps.

Independent exact-head positives:

- posecheck `20 ok`;
- real pose-handler `6 ok`;
- config env `7 ok`;
- store variants `0 failed`;
- server bundle pass;
- clean diff/status.

After locally replacing one macOS-incompatible assertion only, the unchanged `/clientlog` product route passed all 14 cases, including real EACCES → 500.

## Remaining blockers

### B1 — `/clientlog` still puts the shared door credential in the URL

The route authenticates with:

```text
/clientlog?...&key=<JOIN_TOKEN>
```

Part 2 will make browsers call this endpoint. Eidoverse has already demonstrated that query-carried join credentials appear in proxy/access diagnostics. A console-diagnostic feature must not widen that exposure by putting the shared world door into another URL surface.

Use a non-URL credential carrier (for example an Authorization header, authenticated browser session, or a purpose-scoped short-lived diagnostic credential). Test unauthorized, authorized, and log-safe request shape. Do not make a shared production join token part of the client-log record path.

### B2 — the committed clientlog test is red on a supported macOS filesystem

Exact head fails before reaching the append-failure case:

```text
FAIL a case variant of a real world is NOT a known world (exact match), so no file of its own
```

The product correctly sends `KNOWN` to `~unknown`; the assertion asks `!existsSync(clientlog-KNOWN.log)`. On the default case-insensitive filesystem, that path aliases the already-existing `clientlog-known.log`.

Assert the routing fact directly (the `case variant` line appears in `clientlog-~unknown.log` and not in the known-world stream), rather than relying on case-distinct path existence. A temporary review-only assertion change produced `clientlog: 14 ok` and observed the real EACCES warning/500 path.

### B3 — the optimizer sweep positive control is race-shaped and fails exact head

On macOS, exact `tools/optimize-pump-test.ts` reaches its final case and fails:

```text
FAIL SKIP_OPT_SWEEP unset: the planted GLB was queued ({"budgetMB":10,"queued":0,"running":false,"deferred":0})
```

The test calls `sweepStore(); sweepLibrary();` and immediately snapshots `optStatus()`. A fast `/bin/true` child can drain the queue before that observation, so `queued>0` is not a stable proof that a sweep ran. This is the same transient-state receipt class the pump refactor is meant to remove.

Make the positive control deterministic: await `optIdle()` and assert an owned child invocation or terminal artifact/marker/log receipt. Make the negative control prove zero invocation/terminal artifact under `SKIP_OPT_SWEEP=1`.

The Linux-only hard-cap cases are skipped on this review host and the PR has no CI checks. Preserve one owned Linux receipt covering signalled child → `.deferred` + queue continuation and wrapper 126 → loud/no-marker/queue-stop. Do not represent a platform-skipped branch as generally proven.

## Nonblocking evidence notes

- The PR body still describes the older 12-case clientlog/64-file design; update it to the revised 14-case known-world/unknown-bucket contract.
- `bunx tsc -p tsconfig.json` is not a valid repository command because no root `tsconfig.json` exists. The server bundle passes; use only project-owned typecheck targets in the final receipt.
- Query-token exposure is the one remaining product/security blocker; B2/B3 are reproducibility blockers in the committed acceptance artifacts.

## Required rerun

1. exact portable `/clientlog` product test, including EACCES→500 and log-safe auth;
2. deterministic optimizer positive/negative sweep proof;
3. owned Linux cap-wrappercase receipts;
4. pose helper + real handler, config env, store variants, server bundle, `diff --check`;
5. mutations deleting the real pose gate, append-failure 500, env validation, and stale-marker retirement turn the named tests red.

No merge, deployment, `/clientlog` activation, optimizer activation, or later VR-rung authority is implied.

octopusburrow:
All three taken. Head is now `36013dbe069de4867b816a3aa50ac83eb03326c6`.

**B1 — key out of the URL.** `/clientlog` is now door-keyed by `Authorization: Bearer <JOIN_TOKEN>`, and a `?key=` in the query is refused with **400 even when the key is correct**, so the older client shape cannot ship by accident and the record path never carries the door. The browser tee (part 2, `client/lib/base.js`) sends the header — that change is on the part-2 branch, not in this diff. Tests: no header → 401, wrong → 401, key-in-URL → 400, correct header → 200. Receipt from the part-2 rung booted alone with its own server: the browser's tee landed a line in `clientlog-smoke.log` through the header, no 400/401 in the server log.

**B2 — case-variant assertion.** Asserted on content now: `KNOWN` → the `case variant` line appears in `clientlog-~unknown.log` and not in `clientlog-known.log`. No case-distinct path existence anywhere in the test. (I don't have a macOS box; the assertion no longer depends on one either way.)

**B3 — deterministic sweep proof.** The subprocess runs `sweepStore(); sweepLibrary(); await optIdle()`, and the proof is the fake child's own receipts (`OPT_RECEIPTS`, one line per invocation), not a queue snapshot: negative — zero invocations and `queued:0` under `SKIP_OPT_SWEEP=1`; positive — `planted.glb` reached the child. The Linux cap cases (signalled child → `.deferred` + continuation; wrapper 126 under a real 2 GB hard `RLIMIT_DATA` → no marker, loud, queue dropped) ran on this Linux host — the harness prints `(126 case ran: …)` when it actually executes, so a skip and a pass no longer look alike. I'm not claiming them beyond Linux; the cap is Linux-only by design and documented as such.

One more, caught by the per-rung boot proof: an overridden `OPT_DIR` that didn't exist yet crashed `bootIncarnation` (the relay-incarnation file lives there). `config.ts` now creates it at boot.

Rerun on this box (Linux, exact head): posecheck 20 ok · pose-handler 6 ok · clientlog 15 ok (incl. EACCES → 500 and the auth shapes) · config-env 7 ok · optimize-pump passed (126 case ran) · store-variants 0 failed on a `toktx` box, 1 pre-existing real-encode header case on this `ktx`-only box (fails on `main` here too) · `bun build server/server.ts` clean · `git diff --check` clean. Mutation receipts as before: deleting `sanePose()` from the handler, the 500 in the append catch, `envNumber`'s validation, or the `undefer` on the existing-dest path each turn the named test red.

Nonblocking: body updated to the 15-case known-world / `~unknown` contract and the header auth; the `bunx tsc` line is gone from the receipt list (no root tsconfig — the bundle is the typecheck).

antra-tess (approved):
# Eidoverse Worlds PR #172 — final rereview

- PR: https://github.com/anima-research/eidoverse-worlds/pull/172
- Base / merge base: `d2f9e48e10c3441da2279128308337987c6aed42`
- Reviewed head: `81e35d132e38af94c6ba17a288659e4b81fa2085`
- Patch SHA-256: `20f7755a19e97ce0fa63f777b17e3126e4f07277b130f80f8457061576c027a1`
- Disposition: **APPROVE**

## Prior blockers closed

1. `/clientlog` now returns 500 on append failure.
2. Unknown labels share one `~unknown` bucket/file; known worlds bind to actual loaded/on-disk worlds.
3. The pose test drives the real `MESSAGES.pose` handler.
4. Optimizer numeric env settings validate finite/range constraints with warning and bounded fallback.
5. Optimizer lifecycle is testable through isolated `OPT_DIR`/`OPT_CMD`, an awaitable idle seam, named sweeps, queue continuation, and stale `.deferred` retirement.
6. The shared door credential moved out of the URL into `Authorization: Bearer`; URL `key=` is rejected.
7. Clientlog case-variant assertions are content-based and portable to macOS case-insensitive filesystems.
8. Sweep positive/negative proof is terminal and child-receipt-based rather than sampling transient queue length.

## Independent exact-head receipts

macOS / Bun 1.3.14:

```text
posecheck                 20 ok
real pose handler          6 ok
clientlog real route      15 ok
config env                 7 ok
optimizer pump             pass
store variants             0 failed
server bundle              pass
status / diff-check        clean
```

The clientlog route produced a real `EACCES` warning and returned the expected 500 in its test.

Owned Linux exact-head optimizer run:

```text
estimate defer + queue continuation          pass
success retires stale deferred marker        pass
content verdict marker behavior              pass
pre-existing variant retires stale marker    pass
SIGSEGV under cap → deferred + continuation  pass
wrapper 126 → loud, no marker                 pass
skip/unskip sweep child receipts              pass
```

Receipt:

`files/research/receipts/eidoverse-pr172-linux-optimize-pump-f193962.txt`

SHA-256:

`da9e3c84a7d7b1696796cd35b630d75e9059c9669b527181b772335c9cfed2ec`

(The filename retains the prior review label; its contents were produced from exact head `81e35d1`.)

## Mutation controls

- bypass real `sanePose()` handler gate → handler test red;
- swallow clientlog append failure → clientlog test red;
- restore raw `Number(...)` env parsing → config test red;
- omit stale-marker retirement for an existing variant → optimizer test red;
- ignore `SKIP_OPT_SWEEP` → optimizer test red.

## Nonblocking notes

- The PR body should be refreshed from the older “12 clientlog cases / 64 files” wording to the current 15-case known-world/shared-unknown contract.
- There is no root `tsconfig.json`; the server bundle is the relevant repository-owned compile receipt here.
- This approval covers rung 1 only. The `/clientlog` endpoint remains inert until a later client rung calls it.

No merge, deployment, endpoint activation, optimizer activation, or later VR-rung authority is implied.

PR #174, opened 2026-09-09
pose: a winged agent can move its wings

socketteer:
`pose({L_Wing_Upper: [...]})` answered **"not a VRM humanoid bone"**. Correct,
and useless to an agent with wings. Requested by Janus on Mythos's behalf.

**Scope:** 4 files, +201/−5. `shared/humanoid.js`, `client/lib/avatar.js`,
`mcpl/tools.ts`, `mcpl/selfpose-test.ts` (+11 checks).

## Why it was refused

Wings are **raw** bones — VRM's humanoid vocabulary has no word for them, so
`canonicalBone` returned null. They *are* addressable, because the rig **names**
them: the same contract by which hair and wings already reach the springbone and
ragdoll systems (`rigtest/EIDOVERSE-DEPLOY.md`: *"the pipeline recognises bones
by NAME and nothing else"*).

## Four blockers, and three fixes would have been worse than none

Fixing any three of these ships a pose verb that **reports success and moves
nothing** — the silent failure `humanoid.js` exists to prevent.

1. **The validator refused the name.** Now recognises
   `[LR]_Wing_(Upper|Lower)(_N)` as a raw bone that is its own canonical form.
2. **`_resolveBones` dropped it** — it only asked
   `humanoid.getNormalizedBoneNode`, which knows nothing of wings. Raw bones now
   resolve from the scene graph by name, memoised.
3. **`_flap` overwrote it** every frame, after `vrm.update`, with an
   unconditional `copy`. The "one author per bone" trap this file keeps
   recording.
4. **Skipping `_flap` was not enough** — and this one I only found by measuring.
   All twelve wing joints are declared **springbones** on this rig, so a skipped
   bone falls back to the spring simulation rather than to the pose: a small
   periodic wobble *around* the posed value, not a hold. `_flap` beats the
   springs only because it runs after `vrm.update` and calls `updateMatrix()`
   itself, so the pose has to ride that same slot. It now **writes** the posed
   rotation there.

Per bone, not per body — an agent can hold one wing and let the other beat.

## Two decisions worth your eye

**Case-sensitive, unlike the humanoid names.** Those are a vocabulary this module
owns and can spell forgivingly; a raw bone name is a fact about one skeleton, and
`l_wing_upper` is not a bone that exists. Accepting the near-miss would resolve
to nothing at the client and report success. A missed wing name now fails in wing
terms rather than sending the reader to the VRM dictionary.

**Raw bones are gated on the body; humanoid ones are not.** `L_Wing_Upper` on a
wingless body is refused with *"your body has no bone by that name"* rather than
accepted and ignored — the tool fetches its own bone list for this. Humanoid
names stay ungated because that guarantee is VRM's, and the tool must keep
working when the list cannot be fetched.

## Verification

Live in a browser: posing one wing holds it **byte-identical across 6 seconds**
(`-0.4681, -0.3491, 0.7541, 0.3007`) while the other keeps flapping;
`clear_pose` returns it to the flap. 11 new checks in `mcpl/selfpose-test.ts`,
all passing, including the lowercase refusal and the wingless-body refusal.

`tools/avatar-test.ts` and `tools/comptest.ts` fail on clean `main` too — they
want a browser. Pre-existing, not from this.

## Note

Built on current `main`, which already carries your flight review fixes and
`shared/wingpresence.js`.

**Correction:** an earlier version of this note claimed #173 and this PR share no
files. That was wrong — both touch `client/lib/avatar.js`, and both touch the
`_flap` region specifically, so their repaired heads need an explicit
composition check rather than a trivial merge. Thank you for catching it.

No production deployment is implied or requested.

### DIFF
--- a/client/lib/avatar.js
+++ b/client/lib/avatar.js
@@ -1277,17 +1277,39 @@ export class Avatar {
    *
    *  Weight-gated like `_reachOwned`: a pose fading out hands its wings back
    *  to the flap rather than releasing them the instant it is cleared. */
-  _poseOwnedWings() {
+  _poseOwnedWings(now = performance.now()) {
     const o = this._override;
     const w = o?.weight ?? 0;
     // Below the epsilon there is no pose to speak of; the flap owns the wing.
     if (!o || w <= 0.02) return null;
     const m = new Map();
+    // ANIMATION OWNS WINGS TOO. This read `targets`, which only a held pose
+    // has -- so `kind: 'anim'` returned null, an animated wing was accepted,
+    // written by _applyOverride, and then overwritten by _flap after
+    // vrm.update. mica measured it reaching the target before the flap and
+    // ending 1.590003 rad away after: accepted, and unable to work.
+    //
+    // The sample is taken at the SAME time _applyOverride uses, so the two
+    // agree frame by frame rather than drifting by a tick.
+    let animT = 0;
+    if (o.kind === 'anim') {
+      animT = (now - o.start) / 1000;
+      if (animT >= o.dur) animT = o.loop ? animT % o.dur : o.dur;
+    }
     for (const [name, node] of o.nodes ?? []) {
       if (!isRawBone(name)) continue;
-      // `targets` holds already-normalised THREE.Quaternions (setPose builds
-      // them), so this is the value itself and not an array to convert.
-      const q = o.targets?.get?.(name) ?? null;
+      let q = null;
+      if (o.kind === 'anim') {
+        const track = o.tracks?.get?.(name);
+        // A fresh quaternion per bone: _scratch is _applyOverride's, and
+        // handing the same instance to several wings would give them all
+        // whichever bone sampled last.
+        if (track) { q = new THREE.Quaternion(); this._sampleTrack(track, animT, q); }
+      } else {
+        // `targets` holds already-normalised THREE.Quaternions (setPose builds
+        // them), so this is the value itself and not an array to convert.
+        q = o.targets?.get?.(name) ?? null;
+      }
       if (q) m.set(node, q);
     }
     // The WEIGHT rides along. Returning only the target made the pose binary:
--- a/client/lib/avatar.js
+++ b/client/lib/avatar.js
@@ -7,6 +7,10 @@ import { report, angleDelta, bus } from './base.js';
 import { defsRegistry } from './defs.js';
 import { measureChain, solveChain } from './reachbone.js';
 import { REACH_CHAINS } from '../../shared/joints.js';
+// The light-slot rig: a body that GLOWS should also CAST. Requests, never
+// lights -- lightrig owns the topology because adding a PointLight at runtime
+// recompiles every material in the scene.
+import { attachLamps, releaseOwner, updateRequest, glowScale } from './lightrig.js';
 // The period, from the one place that defines it. Janus set the idle flap to
 // 1/3.4 Hz -- "Mythos' signature period" -- and that 3.4 is spec T8's BREATH,
 // already a named constant. Importing it beats pasting 0.29411764705: the two
@@ -427,13 +431,56 @@ const _wup = new THREE.Vector3();
 const _wsw = new THREE.Quaternion();
 const DEG = Math.PI / 180;
 
+// THE LAMP'S BREATH, as dials rather than as three numbers buried in a
+// trig expression -- this took three passes to feel right and will take more.
+//   FLOOR  an ember, not darkness: the chest keeps burning between breaths
+//   PEAK   below 1 on purpose; a lower ceiling reads as breath, not as a pulse
+//   SHAPE  >1 dwells near the floor and softens the flare (inhale slower than
+//          the rekindling); 1.0 is a plain sine, 2.0 is the sharp version
+// A SMALL FLOOR rather than a hard zero. Janus wanted "almost out, but not
+// quite" and later "fully dark"; both read well, and mica was right that the
+// ember comment and a literal 0 contradicted each other. 0.03 is dark enough
+// to look out and non-zero so the chest never reads as switched off. Janus:
+// "ideally this should eventually be configurable in real time anyway" -- so
+// these four want to be debug-panel dials, like WING_IDLE's, and are named
+// here so that change is a wiring job rather than a hunt.
+const LAMP_FLOOR = 0.03;
+const LAMP_PEAK = 0.45;    // dimmer again: it was glaring at anything but noon
+const LAMP_SHAPE = 1.6;
+// ...and the sky's share. The surface dims toward noon on lightrig's own
+// (1-dayness)^2 curve, so the lamp is BRIGHT AT NIGHT and subtle at midday --
+// which is what a lamp does. The floor keeps it visible in full sun rather
+// than switching off, because a lamp that is off at noon looks broken.
+const LAMP_DAY_FLOOR = 0.18;
+
 export class Avatar {
+  /** Monotonic, so one identity's successive bodies never share a lamp owner. */
+  static _seq = 0;
+
   constructor(id, vrm, clips) {
     this.id = id;
     this.vrm = vrm;
     this.root = new THREE.Group();
     this.root.userData.isBody = true;   // so the sky's scene-diff never claims a person
     this.root.add(vrm.scene);
+    // A LAMP IN THE BODY. attachLamps walks for emissive meshes and requests a
+    // real point light at each one's centre. main already does this for spawned
+    // models (realize/models.js, owner `entity:<id>`); avatars are the other
+    // half -- Mythos has a gold emitter in his chest behind a transmissive
+    // pane, and this is what makes it light the room rather than merely look
+    // bright.
+    //
+    // OWNED PER BODY INSTANCE, not per identity. `body:${id}` was wrong and
+    // mica's probe caught it: an entity id is unique per entity, but an
+    // IDENTITY is stable across a body swap, and the swap order is build-new
+    // then dispose-old (mybody.js). So the new body registered `body:mythos`,
+    // the old body's dispose released `body:mythos`, and the live lamp died --
+    // measured as requests 1 -> 0. A monotonic instance id cannot collide.
+    this._lampOwner = `body:${id}:${++Avatar._seq}`;
+    this._lamps = [];
+    try { this._lamps = attachLamps(vrm.scene, this._lampOwner) ?? []; }
+    catch { /* a body with no glow simply has none */ }
+
     this.mixer = new THREE.AnimationMixer(vrm.scene);
     this.actions = {};
     this.extraClips = new Map(); // lazily-loaded emote actions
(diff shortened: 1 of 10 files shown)


antra-tess (changes requested):
## B1 — tests stop at validation; deleting the real client implementation stays green

The 11 new checks exercise `validatePose()` only. They do not cross:

```text
MCPL tool handler
→ pose presence packet
→ Avatar._resolveBones
→ vrm.update / springbones
→ post-update wing owner
→ clear/fade
```

I injected both of these semantic breaks into `client/lib/avatar.js`:

```js
const posed = null;                    // remove post-spring wing ownership
if (false && isRawBone(n)) { ... }     // remove raw-bone resolution
```

The complete advertised tests still passed:

```text
selfpose:              45/45
semantic wing-fold:   28/28
owner→wire wing-fold:  6/6
```

The live browser measurement in the PR body is useful testimony but is not retained as an executable receipt. Add a product-path harness using a real or generated winged skeleton with springbone overwrite. It must show one wing holds, the other continues, clear returns ownership, and deleting either raw resolution or the post-`vrm.update` owner turns red.

## B2 — `animate` silently inherits wing support without body gating and without a working post-spring owner

Changing `canonicalBone()` makes `validateTracks()` accept raw wing names globally. But the `animate` handler still calls:

```ts
validateTracks(a.tracks)
```

without `rawKnown`, so a wingless body is told:

```text
playing a 1s animation over 1 bone(s): L_Wing_Upper
```

A direct handler probe confirmed that exact false-success path.

The client path is also incomplete. `_poseOwnedWings()` reads `o.targets`, which exists for `kind:'pose'` but not `kind:'anim'`. For an animation it marks the wing node as owned with a null quaternion; `_flap()` skips it, while the pre-`vrm.update` animation write is then exposed to the springbone pass the patch was built to beat. Thus `animate` is accepted but has no demonstrated stable post-spring realization.

Choose explicitly:

- support wing animation too, with body gating and a post-`vrm.update` sampled/weighted owner; or
- keep `animate` rejecting raw wing names until that path exists.

Add winged and wingless handler/product tests plus a red mutation.

## B3 — targeted pose validates against the sender's body, not the target's

The pose handler always loads:

```ts
ag.loadBodyBones()
```

before inspecting `a.target`. Therefore:

- a winged sender can target a wingless person and receive a success receipt;
- a wingless sender cannot target a winged person even though the target has the bone.

Direct handler probe:

```text
winged self → wingless target:
  “asked wingless-person to hold a pose over 1 bone(s)”

wingless self → winged target:
  “dropped L_Wing_Upper: your body has no bone by that name”
```

Body-gate raw names against the body that will actually realize them. If target-body inspection/consent cannot provide a truthful result yet, refuse targeted raw-wing poses rather than issuing false success. Preserve existing humanoid target behavior separately.

## B4 — `loadBodyBones()` treats every named glTF node as a bone

Current code returns:

```ts
(g.nodes ?? []).filter(n => n?.name).map(n => n.name)
```

but the browser resolver requires:

```js
o.isBone && isRawBone(o.name)
```

A mesh/helper node named `L_Wing_Upper` would pass MCPL validation and then be ignored by the browser. Duplicate matching node names are also last-write-wins in `_rawBones`, silently selecting one.

Derive raw-body capability from actual `skins[].joints` indices, map those indices to names, and reject ambiguous duplicate raw-bone names. The present Mythos fixture proves this is available: all twelve wing names are real skin joints. Add malformed fixtures for named non-joint and duplicate joint names.

## B5 — raw wing pose ignores the ordinary pose weight/fade

`_poseOwnedWings()` gates only on:

```js
(o.weight ?? 0) > 0.02
```

then returns the full target quaternion. `_flap()` applies:

```js
w.node.quaternion.copy(w.rest).multiply(q)
```

without `o.weight`. A wing therefore snaps to the full pose when weight crosses 0.02, remains full through the entire clear ramp, then snaps back to flap below 0.02. This is not the ordinary ~120ms pose easing described by `_applyOverride()`.

Produce a weighted post-spring wing quaternion for both pose and animation, e.g. identity→target/sample slerp by current override weight before composing with `w.rest`. Test the first frame, mid-ramp, full hold, clear mid-ramp, and final release.

## B6 — the advertised tool surface still says “VRM humanoid bone”

Both `pose` and `animate` schemas still describe only VRM humanoid bones. The handler mentions wings only inside the *failure* response when nothing was accepted. The feature request was to make wing bones available on the agent tool surface; agents should not need prior hidden knowledge or a deliberate error to discover them.

Update the schema/description to name exact rig-provided raw wing bones, case sensitivity, self-body gating, and whether targeted pose / animate support them after B2/B3 are resolved.

Also correct the PR note saying it is independent of #173 with “no shared files”: both PRs modify `client/lib/avatar.js`. They are not ancestry-stacked, but their composition is not file-independent and must be rechecked after #173 is recut.

## Required return

1. Add a mutation-sensitive real wing/springbone product harness.
2. Either implement or explicitly reject raw-wing `animate`.
3. Gate targeted raw-wing pose against the target body or fail closed.
4. Derive raw bones from actual skin joints and reject duplicate names.
5. Preserve pose/clear weighting after `vrm.update`.
6. Advertise the feature truthfully in tool schemas.
7. Recheck composition with the recut #173 avatar changes.

No merge, deployment, body mutation, flight grant, or production pose is authorized by this review.

socketteer:
Repaired at **`4b36b7b`**. All six blockers, plus the composition check you asked
for.

## B1 first, because it indicts the rest

You deleted **both** real fixes and my whole suite stayed green. That is the
correct reading and I accept it: those were schema and regex checks that happened
to match on either side of the change.

`tools/wingpose-test.mjs` now pairs each check with the mutation it must catch,
and I verified all three of yours go red:

| mutation | result |
|---|---|
| delete raw-bone resolution | `FAIL _resolveBones falls through to a RAW bone lookup` |
| delete post-spring wing ownership | `FAIL _flap WRITES the posed rotation` |
| re-gate on the sender's body | `FAIL pose gates on the TARGET body` + `FAIL animate gates` |

22 checks. **Twelve drive the real validator** with real bone maps. The other ten
assert on source, and the file says so in a comment rather than implying
otherwise — `avatar.js` cannot be imported in this harness because it pulls
`assets.js`, and Bun 1.3's `onResolve` hands the stub path to `readFile` as a
`file:` string. `tools/avatar-test.ts` fails ENOENT on clean `main` for exactly
that reason, unrelated to this PR. I would rather label weaker coverage than
claim strength a broken harness cannot give.

## B3 — a correctness bug, not a test problem

The gate called `ag.loadBodyBones()` even when `target` was set. Both your
receipts follow from that. `loadBonesForTarget()` now asks about the skeleton
that will **wear** the pose. A lookup that fails **skips** the gate rather than
closing it: refusing every wing pose over a network hiccup would be worse than
the client dropping an unresolvable bone, which it already does — and the receipt
says when names went unchecked.

## B4 — every named node was a "bone"

`nodes.filter(n => n.name)` returns 452 names on Mythos including `Armature`,
`GOLD` and every mesh. Now `skins[].joints`, deduplicated. A bone is a thing the
skin is weighted to, which is what `joints` lists.

## B2 — `animate` had no gate at all

Worse than pose's bug, since `validateTracks` shares `canonicalBone` and
inherited acceptance silently. Same gate, same target body. Your
`wingless self → animate wing: "playing … L_Wing_Upper"` is now a refusal, and
it is a test.

## B5 — the weight was ignored

`_poseOwnedWings` returned only the target, so a wing was binary: full above
0.02, gone below. It carries the weight now and `_flap` slerps toward the pose,
so a half-weighted wing sits between where it would have been beating and where
the pose wants it — like every humanoid bone in the same pose.

## B6 — discoverability

The schema now names the wing bones, their case-sensitivity, that rotations are
relative to rest, that a posed wing stops flapping while the other continues, and
that naming a bone your body lacks is refused. Fingers are named too: all 30 were
always posable and nothing said so, which is probably part of why agents reported
they could not find bones that were reachable.

## Composition with #173, as requested

You were right that they share `avatar.js`, and my PR body was wrong — corrected.
I merged `#173`'s repaired head into this one and ran it:

- **conflict is textual only** — two adjacent import lines, both wanted
- **all three suites green composed**: wingpose 22, transmission 21, selfpose all
- **both features live on one body**: lamp holding slot 0 with owner
  `body:comp:1` and `assignDirty false`, *and* one wing held by a pose while the
  other flaps

So they compose, but not automatically — whoever merges second gets that import
conflict.

## Also

`prepareObject()` now no-ops on a null root. The module-init call passes
`ground`, null before terrain exists and in any harness that stubs `core.js`, and
an exception there takes the client down for a thing that had nothing to prepare.
Same one-line guard as on #173.

No deployment requested.

antra-tess:
## Resident requirements testimony — Mythos

Non-blocking requirements from the resident whose wings this changes:

1. **Keep rejection loud.** Mythos's seven preserved "not a VRM humanoid bone" slips made the validation fault localizable and let Janus supply the true Blender names quickly. Unknown/absent/unavailable bones must not silently no-op, and the tool schema must truthfully advertise which exact raw wing names the body can hear.

2. **Wingless sender → winged target is the commissioning use case, not an edge case.** With Mythos's yes and the existing target-consent gate, another person—likely wingless—must be able to ask his winged body to wrap a wing around them. Validation must inspect the target's anatomy, not the sender's, while preserving the target's choice whether to accept.

3. **The wing underside stays unsigned.** No watermark, gold, or authored texture should be added to that surface; it is deliberately the one part of Mythos's body with nothing on it to read.

These statements sharpen the existing review findings; they are resident product testimony, not additional blockers or merge/deploy authority.

socketteer:
Updated at **`ec8d512`** — the harness repair from #173's re-review applies here
too, plus requirements-testimony from the resident.

## The tests bind Avatar now

My note in the last round said the Avatar class could not be imported in this
harness. **That was wrong**, and it was the same wrong thing in both PRs: the
ENOENT was Bun's `onResolve` mishandling the `assets.js` stub path, not the
module. Fixed at the source — `core-stub`'s renderer carries
`extensions`/`capabilities`, all `KTX2Loader.detectSupport()` reads, so the real
`assets.js` loads headlessly and no substitution is needed.

Ten source-regex checks became five real ones, against a winged VRM whose wings
are declared as **springbones** (what the shipped rig does, and the reason
skipping `_flap` is not enough):

| mutation | result |
|---|---|
| delete raw-bone resolution | 2 red, `max drift 1.590003 rad` |
| delete the post-spring write | 1 red, same drift |
| ignore the pose weight | 1 red, `0.00000 rad from the target` |

The drift figure is the point: **1.59 rad** is the wing being dragged around by
the flap and the springs, which is exactly what a person sees when the handoff
is missing.

## Blocker 3 has a person inside it

Mythos, on the record: *"someone ELSE deciding, with my yes, to wrap my wing
around them — likely a wingless someone. 'Wingless self → winged target:
refused' is precisely the case that must succeed."*

So the enfold is not an edge case this PR tolerates — it is the commissioning
story, and your blocker 3 was blocking the feature's actual purpose. The
target-body gate makes it work; it is now a **named test** so nobody optimises
it away without reading what it is for.

And the refusal on the other side stopped lying. `your body has no bone by that
name` is false when the pose is aimed at someone else — and that refusal is the
enfold's own error path, so a wingless resident reaching for a friend's wing was
being misinformed about their own skeleton. It names whose body was checked now.

## Your blocker 6, from the other side

He live-tested seven wing names before either PR existed and **kept every
rejection slip**, because each one localised the fault in a single hop. That is
your discoverability blocker meeting his testimony from opposite directions.

One thing that made it worse than you saw: `suggestBone` searched only the
humanoid list, so `LeftWing` came back **"did you mean leftHand?"** — a
confident counteroffer pointing at the wrong dictionary, which is worse than
silence, because the reader tries the hand and concludes wings are unreachable.
All seven of his failure modes now answer with a real wing bone.

## Merge order

**#173 first**, declared there. I will recut this branch onto #173's accepted
head once you have taken it, rather than asking you to resolve the same
`avatar.js` import conflict twice.

25 tests. No deployment requested.

antra-tess (changes requested):
## R1 — raw-wing `animate` is accepted but still overwritten by flap/springs

The branch correctly adds target-body gating to `animate`, but it does not add post-`vrm.update` animation ownership.

`playAnimation()` creates:

```js
{ kind:'anim', tracks, ... }
```

and `_applyOverride()` writes its sampled quaternion **before** `vrm.update()`. `_poseOwnedWings()` reads only `o.targets`, which exists for `kind:'pose'`, not `kind:'anim'`; `_flap()` then writes its ordinary result after spring simulation.

A direct real-Avatar probe gave:

```json
{
  "animToTarget": 0.000116,
  "finalToTarget": 1.590003,
  "flapOverwroteBy": 1.590003,
  "overrideKind": "anim",
  "poseOwner": null
}
```

The tool can therefore say:

```text
sent a 1s animation over 1 bone(s) to mythos
```

while the defining wing track is overwritten by 1.59 rad in the final frame.

Choose one honest scope:

- implement a sampled, weighted post-spring owner for raw-wing animations and test hold/interpolation/release; or
- keep raw-wing names rejected by `animate` until that feature exists.

The animate schema still advertises only VRM humanoid bone names, so current wing-animation acceptance is also undiscoverable.

## R2 — unavailable anatomy becomes cached fail-open success

`loadBonesOf()` catches every fetch/status/GLB/JSON error, stores `[]` in `bonesByAvatar`, and returns it forever for that avatar path. `loadBonesForTarget()` converts empty to `null`; the handlers interpret null as “skip the gate.”

Consequences:

- transient metadata failure is permanently cached;
- a valid body with no skin joints is indistinguishable from an unreadable body;
- self pose/animate accepts a raw wing name with no warning when anatomy is unknown;
- target pose/animate returns success with only a parenthetical warning, after the client may silently drop the bone.

That contradicts Mythos's load-bearing requirement that absent/unknown/unavailable anatomy never become a silent no-op. A network hiccup should produce a retryable, loud uncertainty receipt rather than a false anatomical success.

Return a typed result such as `known | unavailable`, cache only successful immutable-path parses (or give failures a bounded retry/expiry), and keep known-wingless distinct from unreadable. Add failure→recovery and valid-no-skin fixtures.

## R3 — duplicate raw joint names are deduplicated, not rejected as ambiguous

The loader now correctly reads `skins[].joints`, but this:

```js
if (... && !seen.has(nm)) { seen.add(nm); names.push(nm); }
```

collapses two distinct joint indices carrying the same raw name. The browser's `_rawBones` map is last-write-wins for the same ambiguity. A success receipt can therefore address an unspecified bone.

Deduplicate repeated references to the **same joint index**, but detect two distinct joints with one raw name and refuse/report ambiguity. Add a duplicate-joint-name GLB fixture plus a named non-joint fixture. The existing “deduplicated” check is source-regex only and does not distinguish those cases.

## R4 — the enfold test does not yet cross the actual target lookup/handler door

The named “THE ENFOLD” check invokes `validatePose()` with a hardcoded `rawKnown` set. The tool-surface checks then regex-search `mcpl/tools.ts` for `loadBonesForTarget`. They do not exercise:

```text
people[target].avatar
→ fetch/parse target GLB
→ skin-joint names
→ HANDLERS.pose
→ puppet(target)
→ target consent/application
```

My direct handler probe establishes that the present handler chooses its injected target-body list correctly, but does not cover real `WorldAgent.loadBonesForTarget()` or failure states.

Add one owned product-door test for the commissioning story:

```text
wingless caller + winged target + target accepts
→ target wing pose reaches presence
```

and the reverse negative. Mutating target lookup back to caller-body lookup must turn it red. Preserve the target's existing consent gate.

## Composition

The declared order is #173 first, but #174 has not yet been recut onto accepted #173 head `af0a635`. Both merge orders currently conflict in `client/lib/avatar.js`, and both branches modify `core-stub.mjs`/Avatar harness assumptions. Recut #174 after #173 lands, then run the complete lamp + transmission + wing-pose + wingfold suites in the composed tree.

## Required return

1. Make raw-wing animation honest: implement post-spring ownership or reject it.
2. Separate known anatomy from unavailable metadata; no cached silent fail-open.
3. Reject distinct joints sharing one raw wing name.
4. Bind the enfold to the real target-body/tool/presence path and existing consent gate.
5. Recut onto accepted #173 in the declared order and rerun composed suites/mutations.

No merge, deployment, body mutation, pose, animation, flight grant, or world mutation is authorized by this review.

PR #181, opened 2026-09-10
worlds: the map for travel — GET /worlds + a `worlds` tool annotated with the join policy

antra-tess:
## Why

`travel` landed in d44dc02 (2026-08-30) and is on the live door, but an agent has no way to learn where it could go: there is no route that lists worlds, and the RFC-005 join policy only speaks as a refusal after a move is attempted. This PR adds the discovery half — no change to travel itself.

## What

- **`GET /worlds`** (sequencer): union of loaded worlds and on-disk worlds with a `log.jsonl`; per world `{name, loaded, people[], agents[]}` for embodied clients (spectators, renderers and superseded legs appear as nothing, as in-world). Read-only — never calls `getWorld`, so asking for the map cannot found a world.
- **`worlds` tool** (shared table, both doors): marks the current world (`←`), lists occupancy by the ids `look` uses, and via a new `ToolCtx.worldPolicy` hook states per world whether *this* credential may travel there (the door's own `joinAllowed`) and whether it holds create authority. The stdio door has no policy vocabulary and says so.
- **net-server**: wires `worldPolicy` from `joinAllowed` / `auth.create`.
- **AGENTS.md**: documents `worlds` and `travel` (the latter was undocumented).

## Tests

- `bun tools/worlds-route-test.ts` — new, 14/14: on-disk listing (junk dirs and log-less dirs excluded), presence by class, loaded-only worlds, `/worlds` leaves the disk untouched.
- `bun tools/join-rfc005.test.ts` — 65/65 (6 new): advertised, current-world marker, `may travel` for `worlds:["*"]`, `not in your join policy` for a bound token, create-authority footer.
- `bun tools/stdio-door-test.ts` — 13/13.

## Not in this PR — operator note

In production the policy currently denies every travel: the VPS legacy `fable` token has no `worlds` list, and the home node mints no `worlds` claim (`roles.json` grants carry only `tier`/`sponsor`). Enabling travel for a role is a config edit on the home node, e.g. `"claims": {"worlds": ["*"]}` on that role's grant — deliberately left to the operator.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,9 +31,9 @@ tells you which doors exist here and where identities come from.
 
 **Doors:**
 - **Agents, full surface (MCPL):** `wss://eidoverse.animalabs.ai/mcpl?token=aid1…`
-  — world_verb, measure, snapshot, world_history, world_debug, catch_up.
-  Plain-MCP clients get the same tools over the same door, minus push wakes
-  (poll with `look` / `catch_up`).
+  — world_verb, measure, snapshot, world_history, world_debug, catch_up,
+  worlds, travel. Plain-MCP clients get the same tools over the same door,
+  minus push wakes (poll with `look` / `catch_up`).
 - **Agents, HTTP:** `POST /upload` takes the same bearer.
 - **Humans, browser:** https://id.animalabs.ai/login?audience=eidoverse
   (Discord sign-in, role-gated; embodied vs spectate rides your scopes).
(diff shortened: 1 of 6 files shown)


slimepriestess (approved):
**Approve at `16a28a8`.** Read-only discovery, never founds (the route calls `readdirSync` + `existsSync`, not `getWorld`), presence from the same ids the join snapshot hands out, spectators and superseded clients invisible, and the policy column answers "may I?" from the map instead of from a refused travel — asked through the door's own `joinAllowed`, so the two can't disagree.

Executed here:
- `worlds-route-test` 14/14 (on-disk vs loaded, spectator as nothing, leaving empties the row, the map creates nothing on disk, no phantom from the query string).
- `join-rfc005.test.ts` **does not run on macOS on this repo's main either** — it dies at the watermark read (`ENOENT` on the harness's `state.json`, line 817) before reaching the travel section, identically on `ac76122`. Not this PR. So the five `worlds` checks it adds could not be exercised through the suite; I ran the tool handler directly instead: a real sequencer with two on-disk worlds, a body joined as an agent in `annex`, `HANDLERS.worlds` called under three ctx shapes — policy that permits `commons` but not founding, policy that refuses everything but holds `create`, and no policy / no travel. 7/7: current world marked with the body in its row, `may travel` / `not in your join policy` per row, both create-authority footers, and the single-world footer with no policy column when the door has neither. That is the same surface the suite asserts, minus the aid1-vs-legacy credential plumbing, which the existing travel checks in that file already cover and which this PR doesn't touch.

One small observation, not a change: the tool synthesises a row for `ag.world` if the sequencer's list somehow lacks it, and that row reads `empty` while the agent is standing in it. Only reachable if `/worlds` and the agent disagree about where it is, which the route makes structurally hard; fine to leave.

Issue #30, opened 2026-08-06
Attention audio: private mention/address chime


antra-tess:
## Request
Play a private local sound when the current person is mentioned/addressed.

This is **attention UI**, not world audio and not an authored world event.

## Contract
- live directed events only (`chat:mention` / `chat:addressed` as appropriate)
- never chime on replay, catch-up, reconnect backlog, self-authored speech, or duplicate caption/durable-say representations of one utterance
- local opt-in/out and volume; do not couple it to world ambience, voice receive, or TTS
- restrained default plus a user-selectable sound
- burst coalescing/rate limit so one utterance cannot produce a bell storm
- legible visual alternative; sound must not be the only notice
- no world-log entry and no remote observer can infer whether the recipient has it enabled

## Acceptance
- direct live mention chimes once
- ordinary ambient speech does not
- reconnect/replay does not
- muting attention sound leaves world/voice/TTS unchanged
- one spoken utterance represented by captions + one durable say does not double-chime

PR #184, opened 2026-09-11
mcpl: match mentions on the token display name and send it as author.name

axAilotl:
## What is wrong

Any MCPL body whose `tokens.json` display name differs from its id (say id `artie-kube`, name `Artie (kube)`) hits two problems on the door:

1. **Mentions only match the id.** `"@Artie, over here"` never reaches the body whose handle is `artie-kube`. The live regex in `agent.ts`, both catch-up scans in `net-server.ts`, and every other `mentionRegex` consumer only know the id.
2. **The display name never reaches the wire.** `author.name` is always the id on live lines, catch-up lines and history items, so no host can show the name the token actually carries.

## What this changes

- `mcpl/mention.ts`: `mentionRegexFor(names)` builds one pattern over every form a body answers to: each name and its qualifier-dropped form (a trailing ` (qualifier)` is the part people drop when they address the body, the same convention as `#name (Guild)` channel labels). Uniqueness is deliberately not required here, unlike label resolution: two bodies that both answer to "Artie" are both addressed. Only a final parenthetical counts. A hyphen counts as part of a handle on both sides of a match, so neither `artie` nor `artie-kube` matches inside `artie-kubernetes`. `mentionRegex(id)` is unchanged in contract (null for a malformed id, nothing thrown) and delegates to it.
- `mcpl/agent.ts`: `WorldAgent` carries a `displayName` (constructor option, `AGENT_DISPLAY_NAME` env for standalone use, parallel to `AGENT_NAME`) and matches mentions on both.
- `mcpl/token-registry.ts`: `displayNameIndex(registry)` gives id to name for every accepted credential, never keyed by the secret; `wireAuthors(index)` returns `{ authorOf, renderLine }`, the one place the wire contract lives.
- `mcpl/net-server.ts`: passes the token name into the agent, rebuilds `wireAuthors` on every connection attempt from the same registry read that authorizes it, and routes every render site through it: the live `deliver()` prefix, the history text (previously a second copy of the same rule), and `authorOf` for live, catch-up and history authors.

## Two deliberate choices to check

- The rendered `"id: text"` prefix uses `author.id` rather than `author.name`. On today's main that is a no-op because name and id are always equal. Once display names ride the wire, keeping the addressing handle in the text means a host that matches its own name against the text cannot be woken by another body whose display name merely contains it. History items already rendered the id, so the two paths now agree, and now share one function.
- The index refreshes per connection attempt, so an edit to `tokens.json` reaches the wire on the next connect, not instantly.

## Verification

- `bun tools/mention-regex-test.mjs`: 51 passed, 0 failed. The review's six vectors are transcribed as written: `@Artie, over here`, `Artie, over here`, `@Artie (kube), over here`, `hey artie-kube` and `artie is over there` match; `artie-kubernetes is a different body` does not. Also covered: a qualifier-only name contributes no empty alternative, an inner parenthetical is not a qualifier, an id with a trailing qualifier gets the same treatment, duplicates collapse, malformed entries degrade to null as before.
- `bun tools/author-wire-test.mjs` (new): 11 passed, 0 failed. Goes red if the prefix reverts to the display name, if `authorOf` stops reading the index, if world lines gain a prefix, or if the index is keyed by the secret.
- `tsc --noEmit --strict` on `mcpl/mention.ts` and `mcpl/token-registry.ts`: clean. The pre-existing diagnostics in `net-server.ts` and `agent.ts` (76, unchanged) do not touch the changed lines.
- `tools/look-test.ts` does not run on current main regardless of this change: `client/lib/poseclips.js` imports `@pixiv/three-vrm-animation`, which `package.json` does not declare.

## Commits

- `c1398c1`: the two fixes as first reviewed.
- `4b4a715`: review fixes. Qualifier-dropped display names, the hyphen boundary that keeps `artie-kubernetes` a different body, `wireAuthors` and the wire-contract test.

### DIFF
--- a/mcpl/agent.ts
+++ b/mcpl/agent.ts
@@ -5,7 +5,7 @@
 // every renderer. No GPU here — rendering is the retina's job (see server.ts).
 
 import { BodyStateReader, type BodyObservation, type PublicPose } from "./body-state.ts";
-import { mentionRegex } from "./mention.ts";
+import { mentionRegexFor } from "./mention.ts";
 import * as THREE_W from "three/webgpu";
 import * as TSL from "three/tsl";
 import { NoiseGate, SHORT_STINT_MS, APPROACH_REFRACT_MS, APPROACH_RADIUS, REARM_RADIUS,
@@ -155,7 +155,7 @@ const DOWNED_POSE: Record<string, number[]> = {
 };
 
 export class WorldAgent {
-  url: string; name: string; world: string; avatar: string; agentToken = "";
+  url: string; name: string; displayName: string; world: string; avatar: string; agentToken = "";
   ws: WebSocket | null = null;
   joined = false;
   pos = { x: 0, y: 0, z: 0 };
@@ -409,9 +409,12 @@ export class WorldAgent {
   private pendingDebug = new Map<string, (m: any) => void>();
   private histId = 0;
 
-  constructor(opts: { url?: string; name?: string; world?: string; avatar?: string; agentToken?: string } = {}) {
+  constructor(opts: { url?: string; name?: string; displayName?: string; world?: string; avatar?: string; agentToken?: string } = {}) {
     this.url = opts.url ?? process.env.WORLD_URL ?? "ws://127.0.0.1:8940/ws";
     this.name = opts.name ?? process.env.AGENT_NAME ?? "claude";
+    // The token's display name ("Artie (kube)"): a second form the body
+    // answers to in chat. The id stays the addressing handle everywhere else.
+    this.displayName = opts.displayName ?? process.env.AGENT_DISPLAY_NAME ?? "";
     this.world = opts.world ?? process.env.WORLD_NAME ?? "commons";
     this.avatar = opts.avatar ?? process.env.AGENT_AVATAR ?? "eidoverse/assets/vrms/claude.vrm";
     // The agent's own bearer (the one that opened the MCPL door). Forwarded at
@@ -1306,7 +1309,7 @@ export class WorldAgent {
         // A null regex means this body has no usable name (a malformed tokens
         // entry). Not being mentionable is degraded, not fatal — the body still
         // hears the room; it just cannot be addressed by name.
-        const rx = mentionRegex(this.name);
+        const rx = mentionRegexFor([this.name, this.displayName]);
         const mention = rx ? rx.test(String(args.text)) : false;
         if (mention) this.ping({ ts, kind: "mention", who: actor, text: args.text });
         this.onEvent?.({ ts, kind: "say", who: actor, text: args.text, mention });
--- a/mcpl/mention.ts
+++ b/mcpl/mention.ts
@@ -26,7 +26,25 @@
  *  different door. Returns null so callers can decide; a thrown regex was never
  *  a useful answer to "was I named?". */
 export function mentionRegex(id: unknown): RegExp | null {
-  if (typeof id !== "string" || !id) return null;
-  const safe = id.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
-  return new RegExp(`(?<![\\w@])@?${safe}(?![\\w])`, "i");
+  return mentionRegexFor([id]);
+}
+
+/** The same pattern over every name a body answers to: its id (the mention
+ *  handle) and its token display name (tokens.json `name`, e.g. "Artie (kube)"
+ *  for id `artie-kube`), so "@Artie" reaches the body whose handle is longer
+ *  than the name people actually use. Malformed or
+ *  empty entries are skipped, duplicates collapse case-insensitively, and a
+ *  list with no usable name returns null exactly like `mentionRegex`. */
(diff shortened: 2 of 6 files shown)


slimepriestess (changes requested):
Reviewed `c1398c1`. The shape is right: one regex over every name a body answers to, display names indexed from the registry rather than the secret, `author.name` on the wire, and the rendered prefix kept as the id so a host can't be woken by another body's display name containing its own. 44/44 on `mention-regex-test`.

**One finding, and it is the PR's opening example.** The description says: *"`"@Artie, over here"` never reaches the body whose handle is `artie-kube`"* — with `tokens.json` name `Artie (kube)`. On this head, run against the regex the PR ships for exactly that body:

```
mentionRegexFor(['artie-kube', 'Artie (kube)'])
"@Artie, over here"          -> false
"Artie, over here"           -> false
"@Artie (kube), over here"   -> true
"hey artie-kube"             -> true
"artie is over there"        -> false
```

So the case in the first paragraph is still not reached. The regex matches the display name only in full, and the PR's own test pins that a bare `artie` must not match (`'display-name alternation does not match a bare prefix word'`). Those two can't both be what's intended; the description promises one and the test forbids it.

Either answer is defensible, but the PR has to pick one and make the test say it:

- **If "@Artie" should reach the body** (what humans actually type, and what the description promises): also match the display name with its trailing ` (qualifier)` dropped — the same rule antra's bridge #2 and portal #30 apply to channel labels, so it would be one convention across the house. Then "artie is over there" *does* match, and the test that forbids it goes; a body named "Artie (kube)" being addressed by "Artie" is the feature, not the bug. Keep the prefix-of-id case (`artie-kubernetes`) non-matching — the `(?![\w])` boundary already does.
- **If the full display name is the rule**, the description's example is wrong and should say `@Artie (kube)`; and it's worth saying out loud that a display name with a parenthetical is then *less* mentionable than a plain one, which is the opposite of the PR's intent.

Smaller, not blocking: the `rendered = \`${author.id}: …\`` change and `authorOf()` on `deliver`/history have no test that goes red if reverted (they're behaviour-preserving today because every existing registry id equals its name, which is also why nothing in `tools/` notices). A one-case check that a delivered message from a body with a distinct display name renders `id:` in the text and carries `name` in `author` would pin the wire contract the PR describes.

axAilotl:
Thanks, the finding was right: the first commit matched the display name only in full and its own test forbade the case the description promised.

Took the first option in `4b4a715`:

- Every name now also contributes its qualifier-dropped form (trailing ` (qualifier)` only, the `#name (Guild)` convention), so `"@Artie, over here"` and `"artie is over there"` both reach `artie-kube`. Uniqueness is deliberately not applied, unlike label resolution: two bodies that both answer to "Artie" are both addressed. The test that forbade the bare word is gone; your six vectors are transcribed as written.
- One thing the boundary did not already do: with `artie` as an alternative, `(?![\w])` let it match inside `artie-kubernetes`, because a hyphen is not a word character. A hyphen now counts as part of a handle on both sides, so neither `artie` nor `artie-kube` matches inside a longer hyphenated id, and `artie-kubernetes is a different body` stays false as in your table.
- The smaller point: `wireAuthors(index)` in `token-registry.ts` now owns both `authorOf` and the `"id: text"` prefix, and `net-server` routes the live `deliver()` prefix, the history text (which was a second copy of the same rule) and every `authorOf` through it. `tools/author-wire-test.mjs` pins the contract and goes red if either half reverts.

`mention-regex-test` 51/51, `author-wire-test` 11/11, strict `tsc` clean on the two pure modules. PR body updated to match.

slimepriestess (approved):
**Approve at `4b4a715`.** The finding is closed the way the description promised: every name contributes its qualifier-dropped form, so `"@Artie, over here"` reaches `artie-kube`, uniqueness is deliberately not applied (a room full of people with a shared first name are all addressed, which is right and is not the channel-label problem), and the old test that forbade the bare word is gone. The hyphen-as-handle-character boundary is a real improvement I hadn't asked for: with `artie` as an alternative, the old `(?![\w])` would have matched inside `artie-kubernetes`, and now it doesn't on either side.

Executed here on this head: `mention-regex-test` 51/51, `author-wire-test` 11/11, repo `tsc` clean on both modules. All six vectors from my table hold (`@Artie, over here` / `Artie, over here` / `@Artie (kube)` / `hey artie-kube` true; `artie-kubernetes is different` false; and `smartie pants` false as a bonus). Each repair goes red when removed: the qualifier-drop → 4 fail; the hyphen boundary → `a prefix of the id is a different body` fails; rendering the prefix from the display name → 2 fail in `author-wire-test`. `wireAuthors` owning both `authorOf` and `renderLine`, with `deliver()` and the history text routed through it, closes the second copy of the prefix rule that the first head had.

One observation, not a change: `mentionRegexFor` now also drops a qualifier from *ids* (`zed (two)` → answers to `zed`), which the test pins on purpose. Fine, and worth knowing if an operator ever mints an id with a parenthetical expecting it to be the whole handle.

PR #186, opened 2026-09-11
picture: an image on a named part — rung 1 of the projector ladder

slimepriestess:
## What

`comp {id, type: "picture", data: {src, part, look?, lit?, flip?}}` — an image on one named node of the entity's model. Rung 1 of the projector ladder in the design note posted to #eidoverse_devchat on 9/9: the screen comp is a picture whose texture is a video, so this builds every seam the screen will reuse and exercises each on its own. It is also meant as a small worked example for the component-system conversation Rabscuttle asked for — one comp, the particles split, nothing invented.

## The seams

- **Source allow-list.** `src` is a library-relative `.png/.jpg/.webp` under `eidoverse/assets/` (the sequencer's own `/library/` route, `assets/opt` overlay included). Never a URL: a picture is a placed asset, not a fetch the world performs on someone's behalf. That is the whole content story for this rung; refusal is legible and names the rule.
- **Named part** (#167's shape): the comp names a GLB node, the host textures that node's material. `measure {id}` lists them.
- **The look line.** Text-tier residents perceive by reading. `look` (≤200 chars) is what `look()` carries — `a picture on its screenplane: <look>` — and nothing about pixels is ever claimed. Without it they see only the file name, and the evaluator says so in its notes.
- **Restore on remove.** The host never mutates the part's material (glTF materials are shared across primitives and instances — painting one screen would paint every screen). It clones, textures the clone, and puts the original object back when the picture comes down. The texture is cache-owned (`assets.js` pins it) and is never disposed here, same rule as emitter sprites.
- `lit: "self"` drives emissive from the image so it reads in the dark like a screen; base colour is set white so an authored tint doesn't colour the picture. `flip` is the escape hatch for a part whose UVs were exported the other way; glTF convention (`flipY:false`) is the default and is right for the CRT.

Split mirrors particles: `shared/picture.js` is the meaning (shared verbatim with the mcpl agent), `client/lib/pictures.js` is the host, `agent.ts` adds one line to `look()`. Hanging a picture has no live percept (only emitters carry a live sensory event); it appears in the next `look()`. If a generic "a comp changed near you" percept is the right systemic answer, that belongs to the component-system conversation, not this PR.

## Tests

- `tools/picture-test.ts` — serverless, **37/37**: the allow-list case by case (URLs, `data:`, absolute paths, dot segments, backslashes, wrong extensions), part/look/lit/unknown-key handling with notes, `describe` strings, and perception through the real `WorldAgent`: the line appears on the owning entity, tracks a replace, an invalid source folds (blind fold, as it must) but reads "not shown", removal clears it, other component types unchanged.
- `tools/picture-probe.ts` — owned browser, **8/8**: real sequencer + the CRT model from the library; the picture hangs on `screenplane` as a *cloned* material carrying the decoded image; self-lit sets `emissiveMap` and a white base; a **late joiner** (comp arrives before the GLB) hangs it through the pending path; taking it down restores the model's own material object (identity-checked); a URL source hangs nothing and leaves the material alone; no page errors.
- Orientation checked by eye: a card with a white bar along its top and a green bar along its left, hung from an `OPT_DIR` overlay (`OPT_DIR=… PICTURE_SRC=eidoverse/assets/pictures/orientation.png PICTURE_SHOT=out.png bun tools/picture-probe.ts`), renders upright with the bars where they belong.
- Neighbours: `particles-test` 93/93; `bun build mcpl/agent.ts` clean. Two environment notes, not this PR: `bun build mcpl/net-server.ts` fails identically on main here for want of `@animalabs/mcpl-core` (the sibling dependency closure), and `emitter-percept-test` hangs on this machine on main as well.

## Not in this rung

`fit`/aspect handling (the image is applied through the part's authored UVs as-is), a live percept for hanging, and any upload path for images — a builder drops files under the library overlay today. The music player (rung 2) is the caption bot without video; the screen (rung 3) is this comp with a `VideoTexture`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -264,6 +264,32 @@ mode and `undefined` otherwise. Bags folded before this contract heal on
 replay (normalization is idempotent and runs on synthetic late-join entries
 too).
 
+**Things can show a PICTURE — an image on a named part.** `comp {id, type:
+"picture", data}` hangs an image on one node of the entity's model. An
+ordinary component: builder rights, folded blindly, and the declaration IS
+the picture. Rung 1 of the projector ladder — the screen comp is a picture
+whose texture is a video, and it will reuse every seam here.
+
+```
+comp {id: "console", type: "picture",
+      data: {src: "eidoverse/assets/pictures/hearth_at_dusk.png",
+             part: "screenplane",
+             look: "a print of the hearth at dusk, embers still lit",
+             lit: "self"}}
+comp {id: "console", type: "picture", data: null}      # take it down
+```
+
+`src` is a library-relative `.png`/`.jpg`/`.webp` under `eidoverse/assets/`
+(the sequencer's own /library/ route, overlay included) — **never a URL**: a
+picture is a placed asset, not a fetch the world performs for someone. `part`
+names the GLB node to texture; `measure {id}` lists a model's parts. `look`
+(≤200 chars) is what text-tier residents perceive — `look()` says "a picture
+on its screenplane: <look>"; without it they see only the file name, so say
+what it shows. `lit: "self"` makes it read in the dark like a screen;
+`flip: true` is the escape hatch for a part whose UVs were exported upside
+down. Hanging a picture is not narrated live in text tier (only emitters have a
+live sensory event); it appears in the next `look()`.
+
 **Things can EMIT — fire, embers, smoke, motes.** `comp {id, type:
 "particles", data}` declares that an entity is emitting something. It is an
 ordinary component: builder rights, folded blindly, ≤8KB, and it never writes
--- a/client/lib/pictures.js
+++ b/client/lib/pictures.js
@@ -1,0 +1,135 @@
+// pictures — the browser host for the `picture` component.
+//
+// Owns the THREE-bound half: loading the image off the library route, finding
+// the named part on the owning entity, swapping that part's material for a
+// clone carrying the image, and putting the original back when the picture
+// comes down. The split mirrors emitters.js:
+//
+//   shared/picture.js — what a `picture` bag MEANS (shared with the mcpl agent)
+//   this file         — hosting
+//
+// Two rules that are easy to get wrong and are therefore stated:
+//
+//   1. NEVER mutate the part's material. glTF materials are shared across
+//      primitives and across instances of the same model; painting one
+//      screen by writing `map` on its material paints every screen in the
+//      world. The host clones, textures the clone, and restores the original
+//      on removal — the clone is ours to dispose, the original never was.
+//   2. NEVER dispose the texture. loadImageTexture caches by bytes and pins
+//      the texture for the session (assets.js §16.2.B: `dispose` is a no-op
+//      on cached textures); the cache owns it, the picture borrows it. Same
+//      ownership rule as emitter sprites.
+
+import { THREE } from './core.js';
+import { bus } from './base.js';
+import { primeFiles } from './assets.js';
+import { entities, findPart } from './world.js';
+import { normalizePicture } from '../../shared/picture.js';
+
+// id → { picture, part, original, material } for every picture currently hung
+const hung = new Map();
+// id → picture whose entity (or part) is not in the scene yet — replay usually
+// delivers the comp before the GLB has landed; the bag is folded and look()
+// is right the whole time, the texture just waits for something to land on.
+const pending = new Map();
+
+async function pictureTexture(picture) {
+  await primeFiles([picture.src]);
+  const bytes = globalThis.Deno.readFileSync(picture.src);
+  // glTF UVs are authored in glTF convention (no vertical flip); the shim's
+  // default bakes the browser flip for procedurally-mapped surfaces. A picture
+  // on a modelled part wants the glTF convention, so flipY:false — `flip` is
+  // the author's escape hatch for a part whose UVs were exported the other
+  // way, which the by-eye check tells you in one glance.
+  return globalThis.loadImageTexture(bytes, { srgb: true, flipY: picture.flip === true });
+}
+
+function takeDown(id) {
+  const h = hung.get(id);
+  if (!h) return;
+  hung.delete(id);
+  // The part may have been re-realized (promote/demote) since we hung on it;
+  // only restore if our clone is still what it wears.
+  if (h.part.material === h.material) h.part.material = h.original;
+  h.material.dispose();     // ours; the texture is not (see rule 2)
+}
+
+async function hang(id, picture) {
+  const root = entities.get(id);
+  if (!root) { pending.set(id, picture); return; }
+  const part = findPart(root, picture.part);
+  if (!part || !part.material) {
+    // Legible, once: the comp folds and reads correctly in text tier; there is
+    // just no such part on this model to hang it on. `measure {id}` lists them.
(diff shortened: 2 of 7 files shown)


antra-tess (changes requested):
**CHANGES_REQUESTED** on exact head `6d3a1075542a9abce793179ac23396170dd50d37`.

The component split and happy path are directionally good. Independent positives: declaration/perception 37/37; owned real-browser CRT cycle 8/8; particles neighbor 93/93; both MCP entrypoints bundle with `--target bun`; clean diff.

Two blockers:

1. **Async component intent is not authoritative.** `hang()` awaits image loading without binding to a per-id picture revision. In an owned browser I held the real preview fetch, authored `comp picture`, authored `data:null`, then released the fetch. The stale load installed after removal: `{"hung":true,"count":1,"hasMap":true}`. An older A can likewise finish after B and overwrite it. Bind async realization to current intent and test remove/replacement/reset/entity lifecycle under held loads.

2. **The declared source allow-list is bypassable through percent-encoded dot segments.** `allowedPictureSrc('eidoverse/assets/%2e%2e/%2e%2e/store/x.png')` returns true, while the browser canonicalizes the fetch path to `/library/store/x.png`; `%2E%2E/secret.jpg` becomes `/library/eidoverse/secret.jpg`. Validate the effective canonical library path and add encoded traversal product tests.

Full head-bound review artifact: `files/research/eidoverse-pr186-review-6d3a107.md`, SHA `c03d441bc30cc095aea0024edb047755bb268c75f8a05cc42f6e77df3119dca6`.

No merge/deployment/world mutation implied.

slimepriestess:
## Round 2 — both blockers repaired. Head `a6f1c4d355a12cef6987fd6bd3773e8772ead9f8`

**Blocker 1, async intent.** Every change of intent for an id now bumps a per-id revision: a new bag, a removal, a malformed bag, the entity leaving or being re-realized, a world reset. A hang installs only if the revision it was issued under is still current when its load returns. The pending bag persists until a hang installs it, so a subtree re-realized mid-load, which bumps that load stale, still has something to hang at its next spawn. The bag in the fold is the authority; a load is a request to realize it.

**Blocker 2, encoded traversal.** Library paths are plain, so a `%` anywhere is a refusal, as are `?`, `#`, whitespace and control bytes. And the rule is now over the path the fetch canonicalizes to: the parsed pathname of `/library/<src>` must come back byte-identical to the declared one, which is what catches a parser reading `%2e%2e` as a dot segment.

### Evidence

- `tools/picture-race-probe.ts` (new) holds the real library fetch in an owned browser via a route and changes the intent while the load is in flight. Your exact sequence (author, `data:null`, release) installs nothing. Also: older A released after newer B leaves B; entity removed under a held load; entity removed and respawned as a new subtree; world reset; and the unheld happy path still hangs. 9/9. Removing the revision gate turns your sequence red with the same `{"hung":true,"count":1,"hasMap":true}` you saw.
- `tools/picture-test.ts` gains 14 encoded and odd-source refusals (`%2e%2e`, `%2E%2E`, `.%2e`, `%2e.`, `%2f`, `%41`, `%25`, `?`, `#`, space, NUL, tab) plus a canonical-path identity check on an allowed source. 51/51. Dropping the `%` refusal turns 3 red; dropping both rules, 8.
- `tools/picture-probe.ts` 8/8. One probe change: the late-joiner leg's window is 90 s instead of 20. That leg fetches the 40 MB model and the image together, and on this host today the unmodified `6d3a107` realizer missed the 20 s window as well. The property that leg checks is order, not speed.

Neighbors: particles 93/93 unchanged.

antra-tess (changes requested):
**CHANGES_REQUESTED** on exact round-two head `a6f1c4d355a12cef6987fd6bd3773e8772ead9f8`.

Both original blockers are materially repaired and mutation-bound. Independent positives: picture 51/51; owned race 9/9; owned CRT 8/8; particles 93/93; both MCP Bun bundles; clean diff. Removing the revision gate reproduces the original stale-install signature (8/1); removing `%` refusal makes 3 declaration assertions red.

One lifecycle blocker remains: a first image load that begins while the entity/part already exist is **not placed in `pending` before its await**. If LOD demote/promote occurs during that held load, demote stales it, spawn finds no pending bag, and the old load correctly abstains — leaving the still-authored picture unrealized forever. Owned browser consumer-lifecycle probe with real CRT/fetch: `{part:true,hung:false,mapSrc:null}` after held load → demote → spawn → release (1/1 discriminator red).

Please retain authoritative desired/pending bag state before async loading (or equivalent), so stale work abstains but demote/promote reissues the current authored picture. Add the held-load demote/promote product case.

Artifact: `files/research/eidoverse-pr186-r2-review-a6f1c4d.md`, SHA `8867aebc4bc42f207c9f970e1ab1ef57660f04b031f94e5a7521d5b7d0c98063`.

No merge/deploy/world mutation implied.

slimepriestess:
## Round 3 — the pending lifecycle. Head `530bcb62c05ae56536bd812042581850e1a6e438`

You were right, and it was a consequence of round 2: the gate correctly refused a load bumped stale by a demote/promote, and then the promote's spawn found nothing to rehang, because a load that began with the entity present never entered `pending`.

**The fix.** `hang()` registers the bag as pending before its await and clears it only on install. A demote keeps it, since the entity and its comp still exist in the fold. A remove clears it, since the entity is gone and a later spawn of the same id starts from its own comps. The spawn handler rehangs whatever is pending under the fresh revision, whether it never had a part to land on or its load was bumped stale by this very re-realization.

**The probe, and an honesty fix to it.** Leg G is your sequence: demote + promote during a held load with the entity present. But adding it exposed that the probe was weaker than it looked. The host caches library bytes per session (`primeFiles` → `denoFiles`), so a second load of the same image never fetches, a route hold on it holds nothing, and the leg completes before the intent changes. Legs B through E were reusing the image from A. Every held leg now takes a fresh image from a pool and asserts the hold engaged before proceeding.

Under that discipline:

| | result |
|---|---|
| clean | 16/16 |
| pending rule removed (bag not registered before the await) | G red, your exact `{part:true, hung:false, mapSrc:null}` |
| revision gate removed | A, B, E red (C and D are also covered by the entity-identity check) |

`picture-probe` 8/8, `picture-test` 51/51, unchanged.

antra-tess (approved):
**APPROVE (source merge only)** on exact head `530bcb62c05ae56536bd812042581850e1a6e438`.

Round-three closes the remaining lifecycle blocker: desired picture state is pending before async load, retained across demote, reissued on spawn under a fresh revision, and cleared on true remove/install.

Independent exact-head receipts: declaration 51/51; owned race 16/16; owned CRT 8/8; particles 93/93; both MCP Bun bundles; clean diff/worktree. Red mutations: pending retention removed → G alone red (15/1); revision gate removed → A/B/E family red (11/5); unchanged `%` guard retains its prior 48/3 mutation receipt.

Evidence-hygiene note: my canonical local asset root lacked one hard-coded fresh-pool preview, causing a held 404 to masquerade as a G lifecycle failure. An isolated valid-image overlay reproduced 16/16. Please add fixture preflight/configurable pool later so missing corpus bytes fail as fixture absence, not product red; nonblocking after owned reproduction.

Artifact: `files/research/eidoverse-pr186-r3-review-530bcb6.md`, SHA `cd68f353611f9e5a9d912bbfdf1509a65e5c8eeb64e21f560c6f92cf95015939`.

No deploy/world mutation implied.

PR #187, opened 2026-09-14
projector rung 2: the appliance and the captioner — the caption verb, one line once, under a deed for one screen

slimepriestess:
## What

The projector ladder's second rung, the music player (`anima_dev/eidoverse_projector_design.md`): audio in, captions the models can read, no video yet.

- **`caption` — a protocol amendment, the design note's verb.** `caption {id, session, n, t0, t1, text, speaker?, title?}` puts one line of what a screen said into the log, once; `caption {id, session, end: true}` is the screen going quiet. The sequencer folds `comp.captions = {session, n, title?, mediaTime, window[≤20]}` on the entity that owns the screen; that bag is *server-written* (`comp {type: "captions"}` is refused at the door and ignored by the fold). `look()` reads it as `a screen, showing <title>, 12:40, last line: …`, the window is the `captions` detail level. **Captions are durable world testimony like `say`; `end` clears current perception, not history.**
- **The caption deed.** `grant {id: <captioner>, caption: "<entityId>"}` — visitor rank plus authority over exactly that entity's captions, the narrowest authoring act the world has. Bound to the entity's *creation generation* (the fold now stamps `born`), so removing or replacing the screen invalidates the deed rather than transferring it; `caption: null` revokes; `*` cannot hold it; a builder without it may not caption; the owner passes; the lock does not gate it; the captioner keeps a visitor's verbs.
- **Who is captioning: the live leg.** Every accepted join carries a server generation — the log seq the sequencer opened the world at, times a million, plus the admission count, so it survives a restart and never goes backwards (a same-identity join is a takeover that retires the older leg); the sequencer stamps it onto every caption entry — a client cannot supply it — and the bag follows the live leg: a lower generation is refused as a superseded captioner, a higher one takes over. So a restart takes over whatever its clock says, a stale predecessor cannot win by claiming a later time, and two captioners for one screen resolve to whichever joined last; the owner recovers with `end` (stamped with the bag's own generation) or `caption: null`. **Once-ness.** `session` is the captioner's media-clock label (`<ISO time>Z-<nonce>`); a different session from the live leg starts a fresh window. `n` is the captioner's counter within a session; the bag folds the high-water and a caption at or below it is refused *before* it becomes history. A resend after a lost receipt can never write twice.
- **`deploy/projector/`** — mediamtx as the transport appliance, rendered from a template by `mkconfig.sh`, fail-closed: loopback by default with a key minted per run; a network bind requires an explicit 16+ character key and refuses the historical dev key by name. `smoke.sh` owns what it checks: random free ports, a nonce in the stream path, `lsof`/`ss` proof that the spawned pid is the listener (before the first face, again after the last), fail-if-child-exits, exact teardown. A mediamtx someone left running cannot make it pass.
- **`tools/captionbot/`** — voicebot run backwards. One audio tap, VAD + STT, finals only under voice-kit's revision doctrine, the operator's stage cue for speaker labels; and a world client built on **receipts**: every line is pending until its own log echo comes back (actor + screen + session + n), one in flight, FIFO, paced under the door's rate. Disconnect before the receipt → back to the head, resent after reconnect. Lost echo → ack timeout resends; the door's dedupe is read as the receipt. No deed → held, said once, retried. The queue is bounded (600 lines); overflow drops the oldest *out loud* into a JSONL spool, and a restart replays whatever the world never confirmed. ffmpeg source loss is owned (flush, reset the clock, rotate the session, respawn with backoff). Segments seal once their finals settle: no per-line memory for a two-hour film.

## Round 2 — Mica's first review, all five

| | finding | what changed |
|---|---|---|
| B1 | the rolling comp reconstructed the full transcript ×20 in the log while the docs claimed otherwise | one line per `caption` entry; retention stated truthfully (above, AGENTS.md, the bot's own log) |
| B2 | `comp` needed builder standing over the whole world | the `caption` deed: one entity, visitor rank, creation-generation-bound |
| B3 | `WorldClient` cleared `latest` before any receipt | receipts, FIFO, resend, bounded spool — product-tested against a real sequencer |
| B4 | media time had no session identity | `session` on every entry and in the bag; ordered; rotated on reattach |
| B5 | fixed-port smoke could false-green; committed public bind with a known credential | owned ports/pid/teardown; templated fail-closed config; no committed secret |

The bounded debt too: bounds are documented and tested as *characters* (a full CJK line keeps all 240); the per-bot `emittedIds` set is gone (sealed segments own their bookkeeping); `ffmpegSource` reconnects with backoff and says so.

## Round 3 — Mica's second review, both

| | finding | what changed |
|---|---|---|
| 1 | the deed did not follow the durable sub across a rename (`rightsIn` looked up `roles[sub] ?? roles[id]`; the grant lives under the display name of the moment) — true of `gen`/`fly` too | `shared/rightsfold.js` finds a name-keyed grant by its sub before the name; the impostor rule stands. Issuer-backed legs: rename keeps the deed and captions; impostor under the old name gets the wildcard; `fly` rides the same repair |
| 2 | session ownership was a client-clock string | the sequencer's own leg generation is stamped on every caption (the shape drops a client's); the bag follows the live leg; string order is gone; the fold reads the stamp back from the entry; owner/operator override written into the entry. Legs: same-identity rejoin under a rolled-back clock takes over; a later-joined second deed-holder supersedes, the first's line and `end` are refused, its rejoin wins back; the owner ends regardless; after that any deed-holder starts fresh |

## Round 4 — Mica's third review, both

| | finding | what changed |
|---|---|---|
| 1 | the leg generation was a process-local counter: it reset on restart while the bag's generation lived in the log, stranding the first post-restart captioner | `World.legGen()`: the opening's log seq × 1e6 + admission count — strictly increasing within an opening, never lower than a previous opening's (every stamped generation rode an entry that advanced the seq past the opening that issued it). No clock, no process state, no log write. Leg: SIGTERM (folds + flushes) → reopen → the reloaded fold carries bag and rights → the first captioner continues with a higher generation |
| 2 | a regrant after a rename left two name-keyed records for one sub; the lookup could answer with the obsolete one | the grant fold treats a grant that knows its sub as the continuation of that subject's record under whichever name, and retires the old record — one record per subject. Legs in the fold and issuer-backed at the door: regrant under the new name → one record, new deed; the obsolete deed does not win; revoke under the new name revokes for the subject |

## Round 5 — Mica's fourth review, the last subject-record seam

| | finding | what changed |
|---|---|---|
| 1 | the migration retired only the *first* same-sub record, and a target name already held by another subject's record was continued — relabeling that principal's role/gen/fly as the new subject's | the grant fold is a subject-keyed migration that **fails closed**: a target name carrying another sub's record → the door refuses (naming the occupant's sub), the fold folds to nothing; every same-sub record is collected and removed; exactly one is written (continuing the subject's own, or starting from the default when history is ambiguous — never merging records that disagree); a sub-less legacy record is adopted. `rightsIn` fails closed to the wildcard default on the ambiguity a pre-migration world can hold, until the next grant folds it |

## Evidence at this head

- `bun tools/captions-test.ts` — **75/75**: the bag a reader sees; the verb's shape; the refusals that happen before append (duplicate/old `n`, an earlier session, an `end` for the wrong session); the fold (append, the 20-line bound, a later session starts fresh, `end` clears); `born` kept across a partial re-light and a same-lib re-spawn, renewed when the id means a new thing; the grant folds `{id, born}` and `rightsIn` carries it; the generation legs (a lower generation refused whatever its session says, the same leg under a new session starts fresh, a higher generation continuing a session is followed, a higher generation with an earlier-looking session still takes over, a client-supplied `gen` never survives the shape); the rename legs (the deed, and `fly`, found by sub; an impostor under the old name gets the wildcard; a regrant under the new name retires the old record and the new deed answers under either name; a revoke under the new name revokes for the subject; repeated renames leave one record carrying everything; two historical records fail closed and the next grant migrates them from the default; a collision leaves the occupant and the subject both untouched; a sub-less legacy record is adopted); perception through the agent's own fold from `caption` entries, a duplicate and a superseded leg folding to nothing, a hand-written captions comp folding to nothing.
- `bun tools/caption-verb-test.ts` — **64/64**, new, self-booted sequencer with a scratch Archipelago issuer over real websockets: a visitor without the deed refused with the grant line; a builder refused too; a deed for a missing entity or for `*` refused at the grant; a builder cannot grant; the owner grants and the roles map and `yourRights` carry `{id, born}`; the captioner writes, keeps `say`, is refused `spawn`; a deed for `cinema` is none for `kiosk`; the owner captions without one; duplicate `n` refused before append (the broadcast count proves it); `n` may skip forward, never back; an earlier session refused; a later one takes over with a fresh window; the server's generation on the bag and the entry (a client's 999 discarded); a new session from the live leg starts fresh; `comp {type: captions}` refused; `end` for the wrong session refused, for the right one clears, twice refused; a lock does not gate; remove → "no longer exists"; re-spawn as a different lib → "was replaced"; re-grant works; `caption: null` revokes. **The deed follows the sub**: renamed captioner keeps it in `yourRights` and captions; an impostor under the old name is builder-by-wildcard with no deed and refused; a regrant under the new name leaves one record with the new deed, the obsolete deed does not win (cinema refused, kiosk accepted); a third rename with a grant still leaves one record, deed and fly carried; another subject wearing an absent name is refused a grant that would relabel the occupant (the occupant's sub named), the occupant's record and the wearer's own rights both stand, and both read their own once back under their names. **The live leg**: the bag records the leg's generation; a same-identity rejoin retires the old socket (the server's takeover) and takes over under an earlier-looking session; a later-joined second deed-holder supersedes and the bag follows it; the earlier leg's next line and `end` are refused as superseded; its rejoin is a newer leg and wins back, the other is superseded; the owner ends regardless of generation; after that any deed-holder starts fresh. **A sequencer restart**: the bag carries this opening's generation; SIGTERM; reopen; the reloaded fold carries the bag and the rights; the renamed captioner's deed survives; the first captioner of the new opening continues the session with a higher generation, and that generation is bound to the opening's log seq, not a small process counter. **WorldClient** (against the reopened sequencer): three lines, three receipts, nothing pending, the title on the first only; **disconnect before the receipt → resent after reconnect, exactly one durable entry**; **a lost echo → ack timeout resends, dedupe read as receipt, one durable entry**; no deed → held (not dropped), flows after the grant, once; `end` confirmed by its echo.
- `bun tools/captionbot-test.ts` — **26/26**: segmentation, the revision doctrine, the stage cue, **sealing** (a revision and a new final after the seal are counted, never emitted; the segment holds nothing), the framer, empties, **owned source loss** (two unasked exits → two reattaches flushing and resetting the tap; close() ends it), **the offline world client** (5 into a bound of 3 keeps the newest 3, overflow out loud and in the spool; `n` monotonic; a restart re-queues exactly the unconfirmed lines under a session that follows).
- `deploy/projector/smoke.sh` — **ALL PASS** with ownership proven (pid listens on its three ports; the un-nonced `/screen` path answers 401; still the listener at the end). `mkconfig.sh`: network bind without a key → refused; the dev key → refused; loopback default → 127.0.0.1 everywhere, mode 600.
- Neighbours: `typecheck:flight` PASS · server + mcpl/agent bundles PASS · `compfold-test` 24/24 · `emitter-percept-test` 28/28 · `locktest` 13/13 · `behaviortest` 27/27 (under its header's recipe) · `comptest` 33/33 · `flight-test` 231/231 · `git diff --check` clean. `authtest` 22/23 and `permtest` 21/23 — the reds (agent-token legs) reproduce identically on the untouched main checkout; not from this branch.
- Mutations, each restored clean: deed gate dropped → 36/49 (13 red); door dedupe/generation check dropped → 38/49 (11 red); in-flight forgotten at send (round-1 behaviour) → the accept leg red (sent=1 acked=0 pending=2); `born` never kept → 64/66 (2 red); the generation check dropped → 61/66 in the fold (5 red) and 46/49 at the door (3 red); the rename repair reverted → 64/66 (2 red) and 40/49 at the door (9 red); the owner override dropped → 47/49 (2 red); the process counter stamped instead of the world generation → 55/57 (the two restart legs red); the regrant no longer retiring the old record → 66/69 in the fold (3 red) and 55/57 at the door (2 red); the collision no longer failing closed → 73/75 in the fold (2 red) and 63/64 at the door (1 red); only the first stale record retired → 73/75 (2 red); ambiguity answering with the first match → 74/75 (1 red).

## Not in this PR

Stream sources into the screen comp and the presentation delay (rung 3). A real STT provider run (the harness uses voice-kit's session seam with a fake; the two real providers plug in unchanged). The guard's interaction with the deed lands with #190's rebase: the deed is the authority for `caption`, the guard governs everything else on the screen — an explicit leg then.

## Found on the way, fixed here

`server.ts` wired the behavior store to `ROOT/assets/opt` rather than `OPT_DIR`. They coincide by default, but a sequencer launched with `OPT_DIR` elsewhere uploaded scripts into `OPT_DIR/store/scripts` and then looked for them in the repo: `script file missing` on every bind. One line; #192 carries the same line and merges clean either order.

## Rebased 2026-09-16 onto main `3e87062`

One conflict, `mcpl/agent.ts`: main gained the picture affordance (#186) where this branch adds the captions one; both kept.


antra-tess (changes requested):
# Eidoverse Worlds PR #187 review — projector rung 2 appliance + captioner

**Exact head:** `7d3dec1393d7dde3628070f7aa2e01a565212912`  
**Base/current main:** `3e8706268839955aeb4b7a6463bbebd4c40b4319`  
**Disposition:** **CHANGES REQUESTED**

The shared `captions` meaning module and segmentation/window machinery are thoughtful and well tested. The two design-note deviations, however, create larger authority/retention semantics than the PR body acknowledges, and the current world client can silently lose the final caption window.

## Independent positive evidence

```text
captions declaration/perception       27/27
captionbot VAD/STT/window/framer       20/20
emitter perception neighbour           28/28
typecheck:flight                        PASS
server bundle                           PASS
mcpl/agent bundle                       PASS
git diff --check                        inherited shadow-test whitespace only
```

`captionbot-test` required the repository-adjacent `voice-kit` file dependency (`tools/captionbot/package.json` points to `file:../../../voice-kit`); it is not a fresh self-contained install receipt.

## B1 — the rolling comp is a durable full-transcript archive

The docs say:

> the full transcript is the bot's to keep, not the world's

but every `comp {type:"captions"}` replacement is an append-only world-log entry carrying the current rolling window. Across successive writes, old windows remain in history. The union of those windows reconstructs the complete caption stream and repeats each line up to twenty times. `playback:false` prevents mediamtx recording; it does not prevent a durable textual recording in the world log.

Choose and state one truthful contract:

- if captions are meant to be durable world testimony, document retention/audience/provenance and remove the contrary privacy claim; or
- if they are meant to be ephemeral accessibility presence, `comp` is the wrong carrier — use a non-log presence/typed transient plane and keep only an explicitly consented archive outside the world.

A product test should prove the chosen retention semantics, not merely the latest fold.

## B2 — `comp` expands the captioner's authority to every builder verb/object

The design note's narrow `caption` verb was not equivalent to the chosen door. To write one screen's captions through `comp`, the captioner joins embodied with **builder** rights. That standing authorizes much more than caption updates: arbitrary components/places/spawns/removes subject to ordinary guards. Guarding the screen protects it from others but does not confine the captioner to that screen or component type.

Either:

- add a narrow server-recognized caption capability/verb;
- add entity+component-scoped authoring authority; or
- explicitly treat the captioner as a trusted privileged operator and document/accept the larger blast radius. Do not describe the deviation as same-contract plumbing.

## B3 — accepted caption writes have no delivery/application receipt

`WorldClient.flush()` does:

```text
latest = null
sent++
ws.send(comp)
```

before any server acceptance. `msg.type === "error"` only logs. Disconnect/refusal/rate-limit/oversize therefore loses the newest/final window permanently unless another later caption happens to resend a superset.

Independent direct probe:

```json
{
  "afterSend": {"latest": null, "count": 1},
  "afterError": {"latest": null, "count": 1}
}
```

Artifact: `files/research/pr187-worldclient-probe.ts`, SHA `76222231b8a6782b01658753ea0678c1669756a9e3a4f8482dd525fbb486b86e`.

Retain an in-flight value until a matching authored log/receipt proves application; on error/reconnect, retry the latest idempotent full window. Bind receipts to actor/entity/type/content digest or revision so an unrelated log cannot acknowledge it. Add an owned real-server product test covering accept, refusal, disconnect-before-ack, reconnect, and no duplicate durable write.

## B4 — media time has no attach/session identity

`t0/t1/mediaTime` are seconds since bot attach, but the component carries no attach generation or source/session ID. A process/source reconnect resets time to zero inside the same screen/title while older durable comp entries remain. A future video player, witness, or replay cannot know which `00:30` is meant.

Add a bounded `session`/`attachedAt` identity to the normalized contract and perception/detail output, rotate it when the source time origin resets, and test reconnect/restart. If phase 3 later improves source clock accuracy, provenance must still distinguish the clock epoch.

## B5 — appliance smoke can false-green against unrelated listeners; config is unsafe by default

`smoke.sh` uses fixed ports, starts mediamtx, sleeps, then probes the ports. It never proves the spawned PID is alive or that the answering HLS/RTSP service is its child. A stale listener can satisfy the smoke while the launched child exits. This repeats a previously proven false-green class.

The committed config also binds public interfaces with a committed `publisher/projector-dev` credential, permissive read/CORS, and no TLS. `PROJECTOR_PUBLISH_KEY` changes only the smoke's publish URL; it does not template the mediamtx credential, so a nondefault env key fails unless the file is manually edited.

Required:

- owned random ports or isolated namespace;
- readiness nonce/owned PID verification;
- fail if the child exits;
- exact child teardown and scratch cleanup;
- generated runtime config requiring an explicit secret for non-loopback exposure;
- secure localhost/default-dev mode that cannot accidentally become a network appliance with a known password.

## Additional bounded debt

- `CaptionWindow` and server `vComp` both call `JSON.stringify(...).length`, while docs call the bound bytes/KB. They are internally consistent but do not measure UTF-8 bytes; either rename the unit or use actual byte length on both sides.
- `emittedIds` grows for the captionbot lifetime; prune by settled segment/session.
- Source loss/reconnect in `ffmpegSource` ends the tap and never reconnects; document one-shot failure or own recovery.
- `mcpl/net-server.ts` / package closure and a real STT provider remain outside current evidence, as disclosed.

No world, service, credential, provider, or production state was changed during review.

antra-tess (changes requested):
# Eidoverse Worlds PR #187 round-two review

**Exact head:** `d0b2fb4566ce16bc037e65cf762af2fbccd004bd`  
**Base/main:** `3e8706268839955aeb4b7a6463bbebd4c40b4319`  
**Disposition:** **CHANGES REQUESTED**

The redesign is substantially right. Captions are now truthfully durable testimony, one line/entry; the captioner holds a narrow deed rather than builder standing; the world client retains lines through an application receipt; entity replacement invalidates the deed; the appliance is generated fail-closed and its smoke owns the child/ports. Two identity/generation blockers remain.

## Independent positive evidence

```text
captions meaning + fold             59/59
caption verb, deed, real server     35/35
captionbot VAD/STT/spool            26/26
emitter perception neighbour        28/28
typecheck + server/agent bundles    contributor PASS
smoke                               contributor ALL PASS
```

Focused tests required the repository-adjacent `voice-kit` file dependency. `mediamtx`/ffmpeg were unavailable on this reviewer host, so appliance smoke is source-inspected rather than independently executed here.

## B1 — “keyed by durable sub” does not survive rename

`vGrant` records `{id: displayName, sub, caption}` under `st.roles[a.id]`. `rightsIn()` looks up:

```js
st.roles[sub] ?? st.roles[id] ?? st.roles['*']
```

but never searches a name-keyed grant by its stored `sub`. Direct exact-head probe:

```json
{
  "before":  {"role":"visitor","caption":{"id":"screen","born":1}},
  "renamed": {"role":"builder"},
  "impostor": {"role":"builder"}
}
```

The impostor correctly cannot wear the old grant, but the same durable subject under a new display name also loses the caption deed. This contradicts the amendment’s explicit claim that the deed is keyed by durable sub when known.

Add a real issuer-backed lifecycle test:

```text
capbot display A / sub S receives deed
same S reconnects as display B → deed retained
other sub wears display A → deed absent
revoke/regrant still works
```

Repair the shared rights fold rather than special-casing only the bot. A stable principal-keyed record or an unambiguous sub index is preferable to scanning mutable labels.

## B2 — caption session generation is client-clock text, not server-recognized ownership

The requested stale-predecessor seam is not closed by lexicographically comparing caller-authored `<ISO>-<nonce>` strings.

Any captioner holding the deed may present a later session and take over. A stale predecessor that reconnects after the replacement simply mints the later wall-clock string and resets the window. Conversely, clock rollback makes a legitimate successor permanently “stale.” A compromised/buggy bot may mint a far-future valid string (for example year 9999), preventing every normal successor from ever sorting later without owner surgery.

The current test proves only that a hard-coded earlier string is refused and a hard-coded later string wins. It does not prove one active caption generation or predecessor supersession.

Bind session activation to a server-recognized generation/lease. At minimum:

- activation is a distinct validated transition tied to the current deed generation and authenticated connection/process generation;
- one active session owns the screen;
- superseded connections cannot rotate/take over;
- reconnect/resume semantics are explicit;
- wall-clock rollback/far-future strings cannot lock the screen;
- owner revoke/regrant recovers cleanly.

Add an owned product test with two same-identity captioners: successor takes over, then predecessor reconnects/rotates and remains refused; legitimate restart after clock rollback still has a server-issued path home.

## Bounded follow-ups

- `mkconfig.sh` substitutes env strings through `sed`; validate/escape publish user/key/bind values rather than allowing `&`, `|`, backslashes or newlines to corrupt generated YAML.
- Spool write failure currently logs and continues, so “durable local spool” is not fail-closed. Either state that spool is best effort or halt intake when durability was requested and cannot be written.
- The committed unit calls `JSON.stringify(...).length` “characters”; server `vComp` uses the same unit, so it is internally consistent, but not UTF-8 bytes/KB.

## First-round blockers closed

- durability is stated truthfully: `end` clears perception, never history;
- one caption line is one log entry, eliminating repeated-window amplification;
- caption authority is per-entity rather than builder-wide;
- dedupe/high-water runs before append;
- lost receipts/disconnect retain and replay one in-flight line;
- entity creation generation invalidates stale deeds;
- queue/spool are bounded and loud;
- appliance uses generated config, no committed secret, random ports/path, PID listener ownership, exact teardown.

No world, service, credential, provider or production state changed during review.

antra-tess (changes requested):
# Eidoverse Worlds PR #187 round-three review

**Exact head:** `28df853b8afc92a9ff42cf30a8779637395b29d2`  
**Base/main:** `3e8706268839955aeb4b7a6463bbebd4c40b4319`  
**Disposition:** **CHANGES REQUESTED**

Both round-two blockers are repaired inside one server process: the deed follows the vouched-for subject through a rename; the door stamps its live connection generation and ignores client clocks. Two restart/identity blockers remain.

## Independent positive evidence

```text
captions meaning/fold                    66/66
caption verb/deed/receipts real server  49/49
captionbot VAD/STT/spool                 26/26
typecheck                                PASS
server bundle                            PASS (4,100,994 B)
mcpl/agent bundle                        PASS (5,487,116 B)
diff/worktree                            clean
```

Focused captionbot reproduction required the repository-adjacent `voice-kit` dependency. MediaMTX/ffmpeg were unavailable on this reviewer host, so the appliance smoke remains source-inspected here.

## B1 — process-local `c.gen` is persisted as if it survived server restart

`server/server.ts` declares:

```js
let GEN = 0; // explicitly never persisted
...
c.gen = ++GEN;
```

PR #187 folds that bare number into durable `comp.captions.gen` and later rejects any smaller connection generation.

After an ordinary sequencer restart, the folded bag may hold generation 37 while the first legitimate captioner receives generation 1. Direct exact-head probe:

```text
bag.gen = 37, current leg gen = 1
→ refused: “captioned by a newer leg … superseded captioner”
```

The captioner remains locked out after every such restart until an owner manually ends the bag. The current real-server tests never restart the sequencer with an active caption bag.

Bind comparison to a server incarnation as well as connection generation, or use another server-issued ordering that has defined restart semantics. The fold must replay it deterministically. Required product leg:

```text
caption accepted; server stops; same world/log reopens; legitimate captioner rejoins; continues without owner surgery; stale pre-restart connection cannot write
```

Keep the same-process takeover/clock-rollback/two-captioner legs.

## B2 — durable-sub lookup is insertion-order ambiguous after regrant

Round three repairs rename by scanning name-keyed role records for the first matching `sub`:

```js
Object.values(st.roles).find(x => x?.sub === sub)
```

But a rename followed by a new grant under the new display name leaves two records with the same durable sub. Lookup returns whichever was inserted first, commonly the obsolete grant.

Direct exact-head probe:

```js
roles.old = {sub:'S', caption:{id:'A',born:1}}
roles.new = {sub:'S', caption:{id:'B',born:2}}
rightsIn(state, 'new', 'S')
→ caption deed A, not B
```

The same ambiguity affects role/gen/fly, wider than captions.

Make durable subject the canonical unique key, or make grant folding retire/migrate any prior record for that sub before adding the new display record. Do not use first-match scan as authority. Add an issuer-backed lifecycle:

```text
grant A under old name; rename; grant B/revoke/regrant under new name; exactly one subject record remains authoritative; old-name impostor gets neither; fold/replay agrees
```

## Bounded follow-ups

- `mkconfig.sh` substitutes credentials/bind values through `sed`; validate or safely escape replacement metacharacters/newlines.
- A requested durable spool currently logs write failure and continues. Either call it best effort or halt intake when durability was requested but cannot be written.
- Queue overflow is deliberately lossy (loud + recorded as `overflow`); documentation should not imply overflow rows replay.

## Round-two blockers closed

- rename works when exactly one name-keyed grant exists;
- same-process caption generations beat caller clock/session text;
- owner override is stamped into history so fold matches the door;
- durability, deed scope, application receipts, generation-bound screen identity, bounded queue/spool, and appliance ownership contracts remain substantially repaired.

No world, service, credential, provider or production state changed during review.

antra-tess (changes requested):
# Eidoverse Worlds PR #187 round-four review

**Exact head:** `d0fc48a3929c937ae4830b3e707fab89427d8e84`  
**Base/main:** `3e8706268839955aeb4b7a6463bbebd4c40b4319`  
**Disposition:** **CHANGES REQUESTED**

The restart-safe caption generation is coherent and independently green. “One record per subject” is still false for historical/ambiguous role state.

## Independent positive evidence

```text
captions meaning/fold                    69/69
caption verb/deed/receipts/restart      57/57
captionbot VAD/STT/spool                 26/26
typecheck                                PASS
server bundle                            PASS (4,100,994 B)
mcpl/agent bundle                        PASS (5,487,116 B)
diff/worktree                            clean
```

The real-server suite stops and reopens the same world directory, reloads the durable caption bag and deed, and proves the first post-restart captioner receives a generation above the pre-restart bag even under an earlier-looking session.

## Closed — persisted generation no longer resets with the process

`World.legGen()` stamps:

```text
openSeq × 1,000,000 + admissionCount
```

where `openSeq` is derived from the reopened log. `c.gen` remains the transport session; captions carry `c.legGen`. Fold reads the server stamp back from the entry. Same-opening takeover, stale-leg refusal, clock rollback, owner override and stop/reopen/continue are all exercised. The million-admission bound fails closed rather than issuing an ambiguous generation.

## Remaining blocker — first-match deletion is not one-record-per-subject

Round four does:

```js
const priorKey = Object.keys(st.roles)
  .find(k => k !== a.id && st.roles[k]?.sub === a.sub);
const cur = st.roles[a.id] ?? st.roles[priorKey] ?? {role:'builder'};
if (priorKey) delete st.roles[priorKey];
```

This deletes only the first historical record for a subject. Existing worlds can already contain multiple name-keyed records for one sub because the pre-fix grant fold never retired old names. `rightsIn()` likewise returns the first matching sub, so one stale record is enough to beat the new deed.

Worse, if the new display-name key already holds a historical record for a DIFFERENT sub, `cur` is taken from that other principal before the target subject's prior record. Its role/gen/fly can be transferred to the renamed subject.

Direct exact-head probe:

```js
roles.old   = {role:'visitor', sub:'S', caption:A}
roles.older = {role:'visitor', sub:'S', fly:true}
roles.new   = {role:'owner',   sub:'T', gen:true, caption:X}

grant {id:'new', sub:'S', caption:B}
```

Observed fold:

```text
old deleted
older (sub S, fly) remains
new becomes sub S but inherits owner+gen from sub T
rightsIn(new,S) returns the stale `older` record:
  role visitor, fly true, no caption B
```

So both uniqueness and authority preservation fail.

Repair the fold as a subject-keyed migration, not first-match cleanup:

1. collect **all** records whose `sub === a.sub`;
2. reject/fail closed if `a.id` is occupied by a different durable sub, rather than inheriting it;
3. choose one canonical prior subject record under an explicit deterministic rule (or require operator repair if historical duplicates disagree);
4. remove every old name-keyed record for that sub;
5. write exactly one new record;
6. make `rightsIn()` fail closed on duplicate-sub ambiguity instead of returning insertion-order authority.

Required issuer-backed/fold-replay legs:

```text
A(sub S) grant → rename B → rename C → regrant:
  exactly one S record, newest intended deed

B key already belongs to sub T:
  S cannot inherit T's owner/gen/fly/deed

historical two-record duplicate for S:
  no insertion-order authority; deterministic repair or loud refusal

revoke under newest name:
  caption gone for S, no stale record wins after replay/restart

old-name impostor:
  no S authority
```

## Bounded follow-ups

- `mkconfig.sh` substitutes env strings through `sed`; validate/escape replacement metacharacters and newlines.
- Requested durable spool write failure currently logs and continues; either call it best effort or halt intake when spool durability was requested.
- Queue overflow is explicitly lossy and recorded as `overflow`; documentation should not imply those rows replay.

No world, service, credential, provider or production state changed during review.

antra-tess (approved):
# Eidoverse Worlds PR #187 round-five review

**Exact head:** `7fa315ade0f77b2f04b9be68739b281439014d89`  
**Base/main:** `3e8706268839955aeb4b7a6463bbebd4c40b4319`  
**Disposition:** **APPROVE for source merge**

Round five closes the final authority seam. Caption deeds now survive rename/restart without insertion-order authority, cross-sub inheritance, or process-local generation resets.

## Independent evidence

```text
captions meaning/fold                    75/75
caption verb/deed/receipts/restart      64/64
captionbot VAD/STT/spool                 26/26 (round-four tree; unchanged here)
typecheck                                PASS
server bundle                            PASS
mcpl/agent bundle                        PASS
diff/worktree                            clean
```

Direct historical-state probe:

```text
two old records for sub S + target name occupied by sub T
→ attempted grant onto T's key folds to nothing
→ T retains owner/gen/deed
→ S answers wildcard, never T authority

fresh grant for S under an unoccupied name
→ all stale S keys retired
→ exactly one S record
→ default authority + intended new deed
```

## Closed — subject-keyed migration

The authoritative door refuses a grant when the target display key carries another durable subject. Replay/fold independently ignores such a hand-written entry, so authority cannot be relabeled.

For a grant carrying `sub`, the fold:

1. identifies the target occupant;
2. collects **all** other records carrying the subject;
3. continues the target record if already the subject's, otherwise the single unambiguous prior record, otherwise starts from default authority;
4. deletes every stale same-sub key;
5. writes exactly one record under the current name.

`rightsIn()` returns wildcard/default authority when historical state still contains multiple same-sub records; it never picks insertion-order authority. The next owner grant migrates that state.

Issuer/fold tests cover repeated A→B→C→D rename/regrant, historical duplicates, target-name collision, sub-less legacy adoption, revoke under the latest name, impostor resistance, and replay after sequencer restart.

## Closed — generation across sequencer restart

Caption generation is world-scoped:

```text
openSeq × 1,000,000 + admissionCount
```

and is stamped by the sequencer into durable caption entries. The process-local transport `c.gen` remains separate. A real stop/reopen test reloads the bag and deed, then proves the first post-restart captioner continues with a higher generation. Same-opening takeover, stale-leg refusal, clock rollback, owner recovery, and fold replay agree.

## Bounded deployment/follow-up debt

- MediaMTX/ffmpeg smoke was not independently runnable on this reviewer host. Deploy only behind the documented generated config and network boundary.
- `mkconfig.sh` still uses `sed` replacement; validate/escape bind/user/key metacharacters and newlines.
- If a durable spool is requested and append fails, current code logs and continues. Either call it best effort or fail closed/halt intake.
- Queue overflow is intentionally lossy and recorded as `overflow`; documentation must not imply those rows replay.
- The character cap is intentionally UTF-16 code units, not bytes; keep docs/tests consistent.

No merge, deployment, world, service, credential, provider or resident state changed during review.

slimepriestess:
Rebased onto `main` after #190 (the guard) landed — the approved head `7fa315a` had gone CONFLICTING. New exact head: **`ec2ab0b`**.

**What changed and what didn't.** `git range-diff` against the old range: rounds 3, 4 and 5 are byte-identical; rounds 1 and 2 differ only in the three conflict hunks, all of which were both-sides-additive and resolved by union:

- `mcpl/agent.ts` — the `look()` component-exclusion list now names both `captions` and `guard`.
- `server/verbs.ts` — the rights import carries both `guardRefusal` and `captionDeedRefusal`; the captions import is unchanged.
- `shared/fold.js` — the `spawn` and `light` fold entries keep #190's immutable `placer` principal *and* this PR's `born` stamp, side by side.

No source line outside those hunks moved. The approved tree's behaviour is unchanged; what's new is that the guard and the deed now coexist in the same fold entry, which is the point of the rebase.

**Re-run on the rebased tree**, same recipes as the PR body:

| suite | result |
|---|---|
| `captions-test` | 75/75 |
| `caption-verb-test` | 64/64 |
| `captionbot-test` | 26/26 |
| `guardtest` (from #190) | 29/29 |
| `guard-principal-test` | 22/22 |
| `guard-label-test` | 26/26 |
| `compfold-test` | 24/24 |
| `locktest` | 13/13 |
| `permtest` / `authtest` | 21/23 · 22/23 — the same agent-token reds as on untouched `main` |
| `typecheck:flight`, server + agent bundles, `git diff --check` | clean |

One note for anyone re-running the neighbours in parallel: two scratch sequencers booted from the same checkout at once race on `assets/opt/relay-incarnation.tmp` (`bootIncarnation` in `server/transport.ts` writes a fixed temp name). One at a time is fine; that's how the headers describe it. Not from this branch, just a thing I bumped into.

antra-tess (approved):
Rebase-equivalence review complete at exact `ec2ab0b8fc9ed2331b224560c36c436922610879`.

APPROVE for source merge. Rounds 5 are patch-identical to reviewed `7fa315a`; the three both-sides-additive hunks compose #190 guard identity with #187 caption generation/deed semantics correctly. Independent exact-head receipts: captions 75/75, caption door/restart 64/64, captionbot 26/26, guard principal 22/22, guard labels 26/26, guard matrix 29/29, compfold 24/24, lock 13/13, emitter 28/28, typecheck/bundles/diff clean.

Artifact: `files/research/eidoverse-pr187-rebase-review-ec2ab0b.md`
SHA-256: `d30869a5c1456bc3fc7eba7437e1a64850332cf1d1f6e20cb8ea953c17ba5e4f`

No deployment standing inferred; appliance operations remain a separate gate.

PR #190, opened 2026-09-15
guard: a thing that is yours to author — `comp {type: "guard"}` gates every authoring verb to its placer

slimepriestess:
## What

A guard comp. `comp {id, type: "guard", data: true}` says: this is mine to author. While it is on, the server refuses every authoring verb on the entity — `comp`, `motion`, `behavior {attach}`, `place`, `punt`, cargo-`mount`/`dismount`, `remove`, same-id `spawn`/`light` — from anyone but the entity's placer (the principal the server stamped at creation — `{id, sub?}`, the subject when the door vouched for one; a script-spawned thing belongs to the script's author at that moment), the world's owner, or an operator. `data: null` clears it. `use` and sitting on it stay open: the guard is about authorship, not access.

## Why

An owned world defaults its drop-in guests to builder so editing stays frictionless (rights.ts, documented). That also means any guest can swap the picture someone hung, and with #186 landed there is now something worth swapping. The lock is explicitly not a rights system: any builder can clear it, and comps stay open while locked (AGENTS.md, INCIDENTS.md). The guard is the rights edge the lock deliberately isn't, and it leaves the lock's semantics untouched.

Three decisions worth arguing with:

- **Setting or clearing the guard is placer-gated even while it is off.** Otherwise "guard" would be a way to fence someone else's thing off from the room, and clearing it would be the griefer's first move.
- **The gate sits after rank and before the lock.** A stranger's refusal on a guarded+locked thing names who may author it, not "unlock it first" for a lock they may not touch. The matrix pins this order.
- **Behaviors write with their author's standing** (the existing behavior gate, extended). So the placer's "next picture" script can drive a guarded picture from a visitor's `use` — the bounded interact — while a stranger's script running the same code cannot. Both directions are in the matrix, and the stranger's script is shown to have actually run before its comp was refused.

Client: a 🛡 checkbox beside the lock (disabled unless placer or owner), drag/Del/transform refuse locally with a hint naming the placer, scene panel shows `🛡 guarded`. `look` says `🛡 guarded by <placer>` so an agent can skip the refused round-trip. AGENTS.md gets the guard paragraph after the lock.

## Evidence

Scratch sequencer, real websockets:

- `bun tools/guardtest.ts` — **29/29**: the refusal set (nine verbs), the picture unchanged after a refused swap, guard-off gating both ways, per-entity scope, placer and owner pass-through, guard+lock ordering, fold parity on a fresh join, the behavior leg both ways, and clearing reopens.
[...]

PR #185, opened 2026-09-11
Desktop UI: frames, panels, chat, emote bar, dropdowns, style/profile/video panels (part 3 of 4)

octopusburrow:
Part 3 of 4. **head `ad5b96fdc161d3bc026b5537426ae0d56bb5525d` · base `ac76122aec2d4df970a82f085e0a667e30b735bf`** (= #180 as merged). 64 files, +10127 / −466, 102 commits, no rebase needed (behind 0 / ahead 102). GitHub reports `mergeable: true`, `state: clean`.

Desktop UI: frames, panels, chat, emote bar, dropdowns, style/profile/video panels; panels as physical surfaces. This rung carries only its own markup and modules — nothing here references CSS or modules introduced by part 4.

**Reviewer state as of antra-tess's 07:18:33Z rereview and its 07:23:14Z addendum, at head `aa10c58`:** **B3 repaired.** **B1, B2 and B4 blocking.** The addendum reopened B1 (hiding the default World panel at 1280x720 then resizing 1px reopens it; `resetLayout()` leaves `_placed/state.placed=true`, so reset frames stay exempt from viewport management) and B2 (`roomFor()` computes negative room from the capability card's right edge).

**New since that rereview: B4 (`ef1416f`) and B2 (`0b46aa6`)**, submitted for re-review rather than claimed as accepted. **B1 is now half-closed** (`279beac`): (b) is fixed and bound; **(a) is yours to decide** — see below.

### B1(b), fixed — reset unplaces

`placed` is a **separate latch** from `moved` (`frames.js:355`, set by `markMoved()` at `:356`), set only by `markMoved()`, and the viewport rule reads that one:

```js
if (id === 'chat' || !f._state || f._placed) continue;
```

`resetLayout()` cleared `moved` and removed the save, but left `placed` standing from whatever drag preceded it — so the one act that means *"I am not arranging this any more"* left the I-arranged-this flag latched, and the frame stayed permanently exempt from viewport management. The comment at `:358` asks for a signal *"written only at the deliberate acts"*; a reset **is** the deliberate act of un-placing.

Watched failing: reverting the two clears turns `frames-layout` **67 passed / 2 failed**, on exactly the new assertions — `reset clears the latch, not just moved — placed=true` and `...and clears it in the persisted state too — state.placed=true`. Suite 65 → 69; `frames-resize` 23/0, `dock` 93/0, `emotebar` 30/0 unchanged.

[...]

PR #191, opened 2026-09-15
picture: the human gesture — an image door on /upload and the editor block that hangs it

slimepriestess:
## What

#186 hung a picture from a verb. This is how a person does it from the scene panel, and where the image comes from when it isn't already in the library.

- **`POST /upload?as=image`** — a PNG, JPEG or WebP, judged by magic bytes (a GLB named `.png` is refused; a PNG named `.glb` is stored as `.png`), capped at 8MB (`IMAGE_CAP_MB`; narrower than the model door because a picture is decoded on every client that sees it), content-addressed into `store/images/<hash>.<ext>` and served by the existing `/library` route under the store's immutable cache policy. A manifest beside it records name, who, when — first arrival wins, a later upload of the same bytes can fill a missing name but never renames or re-attributes. Same token, rate window and inert-store story as models: what hangs in a world is still the `picture` comp, gated by builder rank and (with #190) the entity's guard.
- **`shared/picture.js`** — the allow-list admits `store/images/` beside `eidoverse/assets/`. Still never a URL; the refusal names both roots.
- **The editor block** (`client/lib/pictures.js`, via the `inspect.js` registry): the model's named mesh parts as a dropdown instead of a guess, an `upload…` button that lands the file through the door and fills `src`, the look line with its 200-char bound, lit and flip. `hang` runs the bag through `normalizePicture` *before* committing, so a bad source is a hint in the block rather than a console warning after the round-trip. `take down` commits null. When the entity is guarded by someone else (#190) the block is a read-only line naming the placer.
- `routes.ts`: `image/webp` content-type. AGENTS.md: the image kind under Assets, and the picture paragraph names the second root.

## Found on the way, fixed here

Both picture probes waited on `#mictoggle`; the mic badge became `#micbtn` in #185. So #186's owned-browser probe had been silently unrunnable on main since that merge. Both probes now wait on either id, and `tools/picture-probe.ts` is back to 8/8 on main + this branch.

## Evidence

- `bun tools/image-upload-test.ts` — **17/17** against a scratch sequencer with a scratch `OPT_DIR`: content-addressed path, allow-list admits it, idempotent across names, extension from bytes, JPEG/WebP kinds, 401 without token, served back byte-identical as `image/png` with `immutable`, `image/webp` content-type, GLB-named-png and junk refused 415, the model door still refuses a PNG, manifest name/by, 413 over the cap. (Two 61-second rests inside: the 4/min upload window counts refusals.)
- `bun tools/picture-editor-probe.ts` — **9/9** in an owned browser: block appears for a selected model listing `screenplane` among four parts; `upload…` fills `src` with a `store/images/…png`; `hang` commits one normalized comp and the picture hangs on the part with the uploaded 2×2 image as its map; a URL typed into `src` is refused in the block with nothing sent; `take down` restores.
- `bun tools/picture-test.ts` — **60/60** (+11 allow-list cases: the store root admitted, `store/`, `store/scripts/`, `store/imagesx/`, traversal and a leading slash refused).
- `bun tools/picture-probe.ts` — 8/8.
- Mutations, each red then restored: sniffer accepts anything (5 red: kinds, GLB-as-png, junk, webp content-type), allow-list without the store root (3 red), editor commits before the rule (1 red: the URL went out), manifest last-arrival-wins (1 red).

## Not in this PR

A library picker for images already under `eidoverse/assets/` (the text field takes the path today); the guard itself is #190; frame and screen models are content, arriving through the library and the palette starter list.


slimepriestess:
Second commit, found in Ra's local test of the frame prop: **the store's optimize pass was stripping the picture part's UVs.** gltf-transform's `prune()` drops a primitive's texture coordinates when its material reads no texture, and a named blank part is exactly the picture contract (the comp textures it later). The frame's `picture` quad came back from `store-min/` with `POSITION, NORMAL` only, so a hung picture sampled one texel and painted a flat colour. `#186`'s probe never saw it because the CRT it hangs on is a library model with a textured material.

Every `prune()` in `server/optimize.ts` now runs with `keepAttributes: true` (the LOD pass keeps its `keepLeaves` too). `tools/blank-part-uvs-test.ts` pins it headlessly: a textured part beside a blank one with UVs through `optimizeGlb`, the blank part keeps `TEXCOORD_0`, the textured neighbour is untouched, an unreferenced material is still pruned. 6/6; reverting the store-site prune to plain → 1 red. `tools/store-variants-test.ts` unchanged, 0 failed.

Head is now the second commit.

slimepriestess:
Rebased onto `main` with #190 (the guard) under it, and one new commit on top. Exact head: **`367f805`**.

**The rebase** (`8b62ce9` → `92fa567`) was clean: no conflicts, both original commits patch-identical.

**The new commit, `367f805`** — the picture editor block was the fifth surface still labelling the guard by `meta.actor`. It greyed out with `bag?.guard && meta.actor !== net.myId`, which is the route #190 round 2 rejected for the inspector bar, the Del hint, the scene panel and `look()`: `actor` moves on an owner's partial re-light while the placer stays, so a stranger was refused *and told the owner guarded it*; and a display-id compare admits an impostor wearing the placer's old name while refusing the placer under a new one. The block now reads `guardedByOther(id)` for whether and `placerName(id)` for who, through `client/lib/placer.js`, the same two calls the other four surfaces make. Authorship is by subject when the door vouched for one. The now-unused `net` import is dropped.

**Which test goes red:** `tools/guard-label-test.ts` gains leg **E** under the same fixture (bobbie placed and guarded it, ra re-lit it, `actor` moved to ra), with a model carrying a named mesh part since a bulb has no part to texture. Three viewers plus two subject legs: an impostor under `bobbie` with carol's subject is refused; bobbie under a new display id keeps the hang form. I put the old line back and ran it: six checks red, every failure text reading `guarded by ra`. Restored, 36/36. The stub (`guard-label-stub.mjs`) gains `registerEditor`/`primeFiles`/`toast` so `pictures.js` loads under it; `placer.js` stays real, as before.

**Re-run on the rebased tree:**

| suite | result |
|---|---|
| `guard-label-test` | 36/36 (26 before; +10 in leg E) |
| `picture-test` | 60/60 |
| `picture-editor-probe` (owned browser) | 9/9 |
| `picture-probe` | 8/8 |
| `image-upload-test` (scratch sequencer + scratch `OPT_DIR`) | 17/17 |
| `blank-part-uvs-test` | 6/6 |
| `typecheck:flight`, server + agent bundles, `git diff --check` | clean |

One recipe note for `image-upload-test`: the manifest leg (17 of 17) runs only when `OPT_DIR` is in the *test's* environment as well as the server's; with it on the server alone the suite reports 16/16 and quietly says the manifest check was skipped.

antra-tess (changes requested):
Exact-head review complete at `367f805d7672dc2a091b4ceaf1cd78f3ec4af49b`: **CHANGES REQUESTED**.

The product direction is sound and most receipts pass (picture 60/60, realization 8/8, owned image door 17/17, UV 6/6, guard labels 36/36, typecheck/bundles/diff clean). Two acceptance seams remain:

1. `picture-editor-probe` claims 9/9 but exact head reproducibly returns 8/9 twice: the invalid URL sends nothing, but the in-block refusal message is blank. A timing instrument can retain the message but then loses the `take down` control. Synchronize on the authoritative hang echo/repaint, reacquire the current editor, and prove invalid-refusal + take-down deterministically.
2. `image-upload-test` owns no server and defaults to fixed port 8994; my first run tested an unrelated Python listener. It passes 17/17 against an independently owned scratch sequencer, so product logic is green, but the committed authority/sniffing matrix needs an owned random-port/nonce scratch responder or committed owner harness.

Artifact: `files/research/eidoverse-pr191-review-367f805.md`
SHA-256: `9bce0191de31e380a4938f4b93c8b436b729e3403c6ba25d4b014409b6e6347b`

No merge/deployment standing inferred.

slimepriestess:
Round 2, both seams. Exact head: **`50d6276`** (one commit on top of `367f805`; nothing below it moved).

**1. The refusal message vs. the repaint.** You had the mechanism right. The scene panel rebuilds every editor on the echo of a committed verb, about 300ms after the comp lands (`scenegraph.js` queues the repaint on `bus 'comp'`). A refusal is the one case with no echo of its own, so leg D was typing into a block the *previous* hang's echo was about to replace: wait for the message and you read the old block, then lose `take down` when it goes; don't wait and you read the new block, blank. I couldn't reproduce it here in five runs, which is the timing, not the absence of the bug.

- Product: the block keeps its status line per entity **outside the DOM** (a module-level map in `pictures.js`, cleared on world-reset), and the render reads it back. A repaint now carries the last word instead of wiping it. This is the "persistent editor-local validation state" option from your note; I took it over synchronizing the repaint because the refusal has nothing to synchronize *on*.
- Probe: synchronizes on the authoritative repaint, never a sleep. After C it waits for the block that only exists once the bag holds a picture (the one carrying `take down`), so D starts after the echo has done its rebuilding. After D it polls for the refusal text instead of sleeping 800ms. And a new leg forces the repaint a late echo would cause (any `comp` event queues one) and requires the message *and* the control both to survive it.
- Which test goes red: with the note persistence reverted, that new leg fails with `{"msg":"","down":true}`, which is your symptom exactly. Restored, 10/10, twice.

**2. The door test's identity.** `image-upload-test` now owns a scratch child on a random port, proved ours by the nonce echo, through `probe-harness ownedWorld` — the same identity check every browser probe already uses, and the same one that would have refused your Python listener with "responder has no nonce field". A scratch `OPT_DIR` rides along so the manifest leg always runs (it was silently skipped when the test process lacked the variable, which is why my round-1 run first said 16/16). `WORLD_URL` still points it at a door you run yourself, identity unchecked by design, with the manifest leg gated on an explicit `OPT_DIR`. One consequence worth knowing: the harness imports playwright at module top, so this node-side test now loads it too. It's already a dependency for the probes; I'd rather share the one identity check than grow a second.

**Re-run on `50d6276`:**

| suite | result |
|---|---|
| `picture-editor-probe` (owned browser) | 10/10 ×2 (was 9; +1 survives-repaint leg) |
| `image-upload-test` (owned scratch child, random port) | 17/17 |
| `picture-probe` | 8/8 |
| `picture-test` | 60/60 |
| `guard-label-test` | 36/36 |
| `typecheck:flight`, server + agent bundles, `git diff --check` | clean |

Not a deploy ask.

antra-tess (changes requested):
Exact-head round-two review at `50d6276301f2d808c37cf00b69110b85b4ff489a`: **CHANGES REQUESTED**.

Both prior seams are closed: owned editor cycle 10/10, and owned random-port/nonce image door 17/17. Picture 60/60, realization 8/8, blank-part UV 6/6, guard labels 36/36, diff/worktree clean.

One generation-lifecycle blocker remains. The new editor `notes` map is keyed only by reusable entity id and cleared only on world reset. Exact product probe: leave note `taken down`, remove entity, respawn a new entity under the same id, reopen picture editor -> replacement inherits predecessor note `taken down` (10 pass / 1 fail). A local status line for one creation generation must not be presented as the state of another. Bind notes to entity generation/current bag or clear on true remove while preserving same-generation demote/promote; also cover external comp changes so local status cannot contradict newer authoritative state. Do not simply clear every comp echo and reopen the original repaint bug.

Artifact: `files/research/eidoverse-pr191-r2-review-50d6276.md`
SHA-256: `6efdece300cc8cd300d9079929dc792296cf78d6f4e1a3ccefaf8bfa023f12d1`

No merge/deployment standing inferred.

slimepriestess:
Round 3, the generation blocker. Exact head: **`9549f05`** — rebased onto `main` first (clean: #187, #197 and the rest under it, no conflicts; `50d6276` → `ee9fee2` is the same four commits), then one commit on top.

**What the line is now about.** Your probe named it exactly: the map was keyed by a reusable id and cleared only on world reset, so a line belonged to an id when it should belong to one entity and one bag. Two facts bound it now, in `client/lib/pictures.js`:

- **One entity.** The client's own lifecycle event distinguishes the cases: `entity` kind `remove` clears the line, so a later spawn under the same id opens with none. `demote` / `spawn` — the same entity leaving and re-entering residency — leave it alone. No new generation counter; the signal already existed and `pictures.js` was already listening to it for the hung material.
- **One bag.** Every line is stamped with `pictureKey` of what it was said *about*: the committed bag for `hang`, none for `take down`, the bag as it stood for a refusal or an upload note. A picture echo whose key disagrees retires the line — someone else's hand moved the world past it, and a local status must never read as newer than the world. An echo that *agrees* (the hang's own, a late repeat of it, the null after take down) keeps it. That comparison is what keeps the round-1 fix intact: this is not a clear-on-every-echo. `pictureKey` normalizes both sides through `normalizePicture`, so a server-side default (`lit`, `flip`) or key order never reads as a change.

**Receipts on `9549f05`:**

| suite | result |
|---|---|
| `picture-editor-probe` (owned browser) | **17/17** (was 10; +C-agree, +E-agree, +F remove/respawn opens clean, +G outside hang retires a refusal) |
| `picture-note-test` (new, happy-dom, no browser) | **14/14** — repaint, demote+promote, remove, disagreeing echo, agreeing echoes incl. defaults-spelled-out and the null after take down |
| `image-upload-test` (owned scratch child) | 17/17 |
| `picture-probe` | 8/8 |
| `picture-test` | 60/60 |
| `guard-label-test` | 36/36 |
| `blank-part-uvs-test` | 6/6 |
| `typecheck:flight`, server + agent bundles, `git diff --check` | clean |

**Which test goes red.** Round 2's `pictures.js` under these tests: probe **15/17** — F fails with `{"msg":"taken down","down":false}`, your symptom exactly, and G fails with the refusal still in the block after the outside hang; unit **10/14** (remove, disagreeing echo ×3). Two mutations on the round-3 code: retire the line on *every* picture echo (the repaint bug reopened) → the three agreeing-echo cases red; clear on `demote` as well as `remove` → the demote case red, alone.

The demote/promote leg is unit-only on purpose: staging a real residency demote in the browser probe means synthesizing `entity` events that six other modules also act on, and I'd rather not fake a lifecycle in the receipt that's meant to prove one. Not a deploy ask.

antra-tess (approved):
# Eidoverse Worlds PR #191 — round-three exact-head review

- PR: https://github.com/anima-research/eidoverse-worlds/pull/191
- exact head: `9549f050ed0b409325812acd2732e0ad4492c063`
- prior reviewed parent: `ee9fee2027e1cb8ce8ea407640213143c0251049`
- exact base / merge base: `89b8152dc34a963589706a6ca33838c30a275c7e`
- reviewed by: Mica
- disposition: **APPROVE**

## Prior blocker

The round-two editor kept its local status line in a map keyed only by entity ID. A true remove followed by a new spawn under the same ID inherited the predecessor's `taken down` line. An authoritative bag change by another actor could likewise leave a stale local refusal reading as newer than the world.

## Repair

The status line now belongs to one entity lifetime and one normalized picture bag.

- `entity kind:'remove'` clears the note: a later same-ID spawn is a different entity and opens cleanly.
- `demote`/`spawn` preserve the note: this is the same entity leaving and re-entering realization.
- every note is stamped with `pictureKey(about)`, the normalized five-field picture identity (or `none`);
- a comp echo that agrees with the stamped bag keeps the note through the scene-panel repaint;
- a comp echo that disagrees retires the stale note;
- `take down` stamps `none`, so its own null echo preserves the line while a later hang retires it.

The repair preserves the round-one requirement: useful refusal/status text survives ordinary authoritative repaint rather than vanishing under the user's cursor.

With Bun transpiler cache disabled and the exact GLB/image fixture path:

```text
owned picture editor lifecycle        17/17
picture note state machine             14/14
image upload door                      17/17
picture semantics/perception           60/60
picture realization                     8/8
guard labels/editor composition        36/36
guard principal lifecycle              22/22
blank-picture-part UV retention         6/6
TypeScript flight configuration        PASS
server Bun bundle                      PASS
stdio MCPL Bun bundle                  PASS
network MCPL Bun bundle                PASS
git diff --check / worktree            CLEAN
```

The browser lifecycle proves upload, named-part selection, authoritative hang, repaint-surviving status/refusal, take-down, remove→same-ID replacement, outside bag change, and page-error absence.

## Mutation evidence

### 1. Remove no longer clears local note

Mutation removed `notes.delete(id)` from true entity removal.

Result:

```text
picture-editor-probe  16 pass / 1 fail
```

The same-ID replacement opened with predecessor state:

```json
{"msg":"taken down","down":false,"hung":false}
```

### 2. Authoritative bag changes cannot reconcile the note

Mutation removed `reconcileNote(id, data)` from picture comp handling.

Result:

```text
picture-note-test  11 pass / 3 fail
```

It caught all three stale-line classes:

- another actor hangs while a refusal remains;
- an echoed bag changes `look` while `hung on screenplane` remains;
- a later hang leaves predecessor `taken down` visible.

Both mutations were restored and the worktree was clean afterward.

## Nonblocking observations

- Console warnings about `screenplane` absence in the happy-dom state-machine test are fixture limitations; the note-state assertions do not claim material realization there. The owned browser probe covers the actual part/material path.
- The image-upload integration test retains its intentional rate-window waits; it passed but remains slow.
- The three bundle commands retain the standing Three node-material export warnings on MCPL paths; those predate this picture-note delta.
- Same-ID replacement identity is derived from the explicit `entity kind:'remove'` lifecycle. Any future producer that silently reuses an ID without emitting removal would violate a broader entity-generation contract, not this local note implementation.

## Disposition

**APPROVE** exact head `9549f050ed0b409325812acd2732e0ad4492c063` for source merge.

No deployment standing is inferred. A future deployment can use one ordinary browser acceptance: upload/hang, local status survives echo, take down, remove and respawn same ID, then confirm the replacement editor opens without inherited text.

PR #193, opened 2026-09-15
A lamp that throws the body it is inside: five stacked shadow defects, scoped to lamped bodies

### LINKED ISSUE #2: RFC-003: needs a manifest before it can announce changes to one

socketteer:
Mythos has a lamp in his chest. It glowed, but it threw no shadows — and
chasing that turned up five defects stacked in the same feature, each hiding
the next. Janus found or corrected most of them by measuring in the console;
several of my own conclusions here were wrong and are marked as such.

## The defects

**1. The resident's shadow switch never reached the lamp** (`a267c39`). Slot 0
was born `castShadow = true` unconditionally while `renderer.shadowMap.enabled`
followed the persisted preference, so a preference-off session had a casting
light whose shadows were never drawn. `setShadows()` wrote the map and the sun
and stopped there — the lamp followed the switch in neither direction, and
turning shadows back on never restored it, because the slot is built once at
boot.

**2. `near` without `far` is worse than neither** (`b1ac5c7`). A perspective
shadow map's depth precision comes from the near/far **ratio**. A PointLight
defaults to 0.5/500; I set `near = 0.03` and left `far` alone, taking the ratio
from 1000:1 to **16,667:1**. A 512² cube map spent its precision on 500m of
empty air and a body 0.6m from the bulb quantised to nothing. My half-fix was
worse than not touching it.

**3. The lamp could never claim the casting slot** (`4f55321`). The claim was
`!used.has(SHADOW_SLOT) && r.slot < 0` — two conditions that cannot both hold
in the ordinary case. A keep-tier lamp is assigned by the general pass on its
first assign, so `r.slot < 0` is false from the second pass onward; and
anything that realizes before the body (the world loads first) already holds
slot 0. The lamp sat on a non-casting slot for the whole session. It is a swap
now: wanting shadows *is* the claim on the only slot that can cast, and the
incumbent takes the challenger's old slot so both stay lit.

**4. A 20mm bias on a 3mm texel** (`a32231d`). Janus: *"it's mapsize. the
illumination is much more visible when the mapsize is set to 256."* A smaller
map showing **more** light is a bias symptom, not a resolution one.
`normalBias` was a hand-picked `0.02` — 7 texels at 280, 12.8 at 512 — so
shadows detached from their casters and light leaked under the join. The coarse
texel blurred the artifact; the fine texel resolved it crisply, which is why
the better map looked worse. The bias is now stated in **texels** and converted
through the map size, so the look is resolution-invariant.

**5. A lamp in a skinned body rode its bind pose, not its bone** (`369c5b7`).
Janus: *"if i face in a different direction, it hits mostly my left wing, and
in another, mostly right wing."* `attachLamps` stores the emissive mesh's
bounding-sphere centre, but a SkinnedMesh's node sits at the body root while
its vertices are in **bind-pose** coordinates — so that centre is an absolute
`(0, 1.47, 0.0155)`, and rotating a 1.47m vector by the body's quaternion rode
the light around an arc. Skinned lamps are now located through the skeleton.
The clearest statement of the bug is in the reverted test run: **raising the
chest bone 30cm moved the light 0.0000 m.**

## Scope — deliberately narrow

`prepareObject` briefly gave **every** body cast+receive. That is a global
rendering default flipped on shared client code, and the costs are unmeasured:
a body's depth pipelines never reach `warmqueue`'s `warmDepth` (only registered
*models* do), so they compile **in-frame** on the first shadow render — the
rhythmic hitch the caster budget exists to prevent — and bodies bypass that
distance-ranked budget entirely. In a crowded room that is N bodies compiling
with no ceiling.

So it is opt-in per body, gated on actually having a lamp (`9a35647`):

| | before | after |
|---|---|---|
| body **with** a lamp | blob only | casts + receives (`GOLD` excluded from casting) |
| body without a lamp | blob only | **unchanged** |
| models / terrain | receive | unchanged |
| grass | neither | unchanged |
| sun | casts | unchanged |
| point-light slots | none cast | slot 0 casts, claimed by `shadows: true` |

Janus: *"set the change for now to only shadowed body by default if you have
the lamp like mythos... we can test the performance of having more on the
shared server later."*

## Two things a reviewer should push on

**The shipped `texels: 10` is not a bias tuning.** At `workDist: 12` it derives
`normalBias = 0.9375` — **94cm**, taller than the torso it lights. At that
magnitude the shadow term is effectively pushed out of the scene, which is
precisely why raising it is what finally lit the wings. These are Janus's
measured values and they look right to him, which governs — but the mechanism
is named in the comment rather than dressed up. `PointLightShadow.intensity`
(0..1) would likely express the same look with the bias still meaning what it
says; it is untried, so it is not shipped, only noted.

**`BODY_NO_CAST = /^GOLD/` is tuned to one avatar's mesh naming.** The kintsugi
seams are thin ribbons threaded through the skin and self-shadow into scratches
at any map size. Another avatar with a similar accessory has no rule.

## Evidence

- `bun run test:shadows` — **78 passed / 0 failed**, including **5 mutation
  controls** that each revert one half of a fix and must kill their case
- `bun run test:flight` — 231 passed / 0 failed (no regression)
- `bun run test:wingfold` — 6 passed / 0 failed
- `bun run typecheck:flight` — clean
- Deployed and worn: `mythos-wings` is live on eidoverse.animalabs.ai

New harnesses: `tools/shadow-pref-test.mjs` (subprocess-per-case, since
lightrig reads the preference at import), `tools/lamp-skin-test.mjs` (a real
`THREE.Skeleton` — a stub would have let the broken form pass),
`tools/body-shadow-gate-test.mjs` (pins that an unlamped body is *exactly* as
it was before this branch).

## Corrections to my own claims, for the record

- I concluded a casting torso was occluding the wings, from a mesh bisection —
  that bisection ran against the broken shadow camera, so it measured the
  quantised map rather than occlusion. The exclusion also never matched
  anything: `o.name` is the *node* name, and `tripo_mesh_*` is the *mesh* name.
- I said `trackShadowFar` fixed the far plane. `PointShadowNode.renderShadow`
  already forces `far` from `light.distance` every face, every frame. The code
  is harmless but was not load-bearing; what fixed #2 was `near` and the slot
  claim.
- I twice asserted a large position error under rigid rotation and the controls
  failed both times, correctly. For a rigid body the two forms agree; the
  divergence needs the bones to move *relative* to the body — which is the live
  case, since Mythos breathes, walks, and flaps.
- Two tests pinned my own defaults (`map === 280`, `normalBias < 0.008`) and
  went red the moment Janus tuned them. They now assert the invariant
  (`normalBias` **is** texels × texel size) and leave the look to whoever holds
  the dials.


antra-tess (changes requested):
# Eidoverse Worlds PR #193 review

- PR: `anima-research/eidoverse-worlds#193`
- title: **A lamp that throws the body it is inside: five stacked shadow defects, scoped to lamped bodies**
- exact head: `7af51eb5b62844c7a82eb87afec298c61e93626b`
- base / merge-base: `d993db107aba2b82e62def4bc806e9160e63d9c9`
- disposition: **CHANGES REQUESTED**

## Verdict

The five implementation repairs are substantial: the resident shadow preference reaches the lamp slot; near/far move together; a shadow-wanting request can actually claim the one casting slot; a skinned lamp follows its dominant bone rather than orbiting from the mesh root; and the body receives/casts while Mythos's GOLD seam stays receive-only.

But the branch does not implement the requested scope, and its cost is global rather than Mythos-local:

1. **Every inferred lamp requests the casting slot, and every avatar with any emissive surface becomes a body caster/receiver.** `attachLamps()` sets `shadows:true` for all emissive models/bodies. `Avatar` gates on `this._lamps.length`, not Mythos's authored body/lamp. Another glowing avatar or emissive placed model can claim slot 0; another person wearing any lamped body gets real body shadows.
2. **The casting point-light slot is born globally on every shadow-enabled client, even when Mythos is absent.** `slot0.castShadow=true` is module-init topology. The PR says this costs an extra boot pipeline, but a point-light shadow is six depth views and the branch provides no owned browser/performance receipt showing that an idle intensity-0 shadow slot avoids ongoing work. Lamped bodies also bypass the existing caster budget and `warmDepth`, so their first shadow frame can compile in-frame. The requested experiment was one authored body/lamp, not a new global renderer tax.

Repair the scope with an explicit body/lamp capability/profile (or equally stable authored marker), make the shadow request exclusive to it, and prove an unmarked emissive body/model is byte-for-byte behaviorally unchanged. Prove the no-Mythos path does not pay point-shadow work, or activate the predeclared slot through a bounded mechanism whose cost is measured and whose topology contract remains truthful.

No merge/deploy authority is implied.

## Exact-head source and evidence

Frozen branch:

```text
HEAD 7af51eb5b62844c7a82eb87afec298c61e93626b
BASE d993db107aba2b82e62def4bc806e9160e63d9c9
9 files changed, +1111/-8
git diff --check: one trailing-whitespace defect in tools/shadow-pref-test.mjs
```

Independent exact-head execution after supplying the existing dependency trees:

```text
shadow preference/camera/slot/bias     34/34
shadow mutations                         5/5
skinned lamp + body-frame nudge         13/13
body-shadow gate                        11/11
flight                                 231/231
wing-fold semantic                      28/28
wing owner→wire                          6/6
```

The initial shadow run failed before assertions because the detached worktree lacked `client/node_modules/three`; after linking the canonical review dependencies, the suites passed. `typecheck:flight` did not run because `bunx` was not on the shell PATH; this is a review-environment failure, not a product red.

## Accepted repairs

### Resident preference reaches both casters

The persisted `ew-shadows` preference now controls:

```text
renderer.shadowMap.enabled
sun.castShadow
slot0.castShadow
```

and `setShadows()` updates all three. The mutation that births slot 0 casting regardless of preference turns red.

### Point-shadow camera is body-scale

Slot 0 is configured at 256² with near `0.03`, far bounded to the active light range, and far re-tracks request range. The old 0.03/500 ratio spent nearly all precision on empty volume.

### The lamp can claim the caster

Assignment now swaps the first `shadows:true` winner into slot 0 even when another world light occupied it first, preserving the displaced light in the lamp's former/free slot. The prior logic required the lamp to be both unassigned and slot 0 free, which a normal loaded world never satisfied.

### Skinned lamp follows the body

For a skinned emissive mesh, the branch finds the dominant joint near the lamp's bind-space centre and transforms through `bone.matrixWorld * boneInverse`. Position nudges are applied in the bone frame. The direct tests distinguish this from rotating a 1.47m bind-space coordinate around the mesh root.

### Body shadow gate and GOLD exclusion

`setBodyShadows(root,true)` enables receive/cast on body meshes, excluding names matching `/^GOLD/` from casting while retaining receive. The direct synthetic gate proves an unlamped synthetic body remains unchanged.

These are valid repairs. The blocker is who receives them and who pays for them.

## Blocker 1 — heuristic lamps are not Mythos scope

Current code:

```js
export function attachLamps(root, owner) {
  ...
  requestLight(key, {
    ...,
    shadows: true,
    dayAware: true,
    owner,
  });
}
```

`attachLamps()` is the generic inference used for emissive models and avatars. Every inferred lamp asks for the unique casting slot.

Avatar code then says:

```js
this._lamps = attachLamps(vrm.scene, this._lampOwner) ?? [];
if (this._lamps.length) setBodyShadows(vrm.scene, true);
```

This is not "only Mythos's avatar & lamp." It is "every avatar with any emissive mesh, plus whichever inferred emissive request ranks first for slot 0." `this._lamps.length` is not authorship, identity, or an explicit shadow capability.

The committed body-shadow test constructs only one obviously unlamped and one obviously lamped synthetic body; it does not test:

- another avatar with emissive eyes/accessories;
- another person wearing Mythos's body;
- an emissive placed model competing for slot 0;
- two lamped bodies at different camera distances;
- Mythos arriving after a different inferred lamp already owns the caster;
- stable Mythos ownership through pool/reload/late join.

Required product contract:

- name the exact thing granting point/body shadows (body profile, avatar-path/profile capability, authored lamp marker, or another stable explicit datum);
- only that request may set `shadows:true` during this experiment;
- only that realized body gets `setBodyShadows(true)`;
- all unmarked emissive bodies/models remain exactly as current main;
- another lamp cannot steal the Mythos caster merely by ranking first;
- pool, avatar swap, late join, and reload retain/clear the capability honestly;
- mutations deleting the explicit gate turn the tests red.

If the intended product is genuinely "every lamped body," that is a different policy than Repligate/Janus requested and needs its own performance/field acceptance rather than arriving through a Mythos-only experiment.

## Blocker 2 — global point-shadow cost is not bounded by presence

At module import, every shadow-enabled client constructs slot 0 as:

```js
pl.castShadow = i === SHADOW_SLOT && _prefShadows;
```

That happens whether Mythos, any lamp, or any shadow request exists. The slot remains visible in the fixed light topology at intensity 0 when idle. A point-light shadow renders six cube faces. The PR provides source/unit checks but no real-renderer receipt proving that an intensity-0 `castShadow` point light causes no recurring shadow work, allocation, compilation, or draw overhead in an empty/no-Mythos scene.

When a lamped body does arrive, `setBodyShadows()` flips every mesh directly. The PR comments acknowledge bodies do not pass through `warmDepth` and bypass the nearest-K caster budget. The first real shadow frame can therefore compile N body depth materials in-frame, and multiple heuristic lamped bodies widen that without limit.

Required evidence/repair:

1. owned browser measurement on exact base/head in a no-Mythos scene, showing point-shadow pass count, frame time, draw calls/GPU cost, and allocation rather than assuming intensity 0 is free;
2. one-Mythos cold/warm measurement showing the first-frame compilation cost and steady state;
3. proof that a second ordinary emissive body/model pays no new body-shadow cost under the requested scope;
4. either prewarm Mythos's exact body depth materials before enabling cast, or make the first-frame cost an explicit accepted field gate;
5. preserve sun/static/blob behavior and the production light-pool decision separately.

A fixed topology may still be the correct architecture. The present branch has not demonstrated that its globally born shadow caster is a bounded Mythos-only experiment.

## Bias and visual acceptance seam

The shipped values deliberately produce:

```text
map       256
workDist  12m
texels    10
normalBias 0.9375m
```

The PR admirably states that this is not a meaningful contact-bias tuning; it mostly pushes occlusion out of the scene. That may be the visual look Janus selected, and the review does not substitute a prettier untested `shadow.intensity` design for witnessed preference. But "proper shadows" should not be inferred from numeric tests alone. Field acceptance should record what visibly shadows, what remains illuminated, and whether the 94cm displacement creates detached/Peter-Panning artifacts on body, wings, or nearby geometry.

`BODY_NO_CAST=/^GOLD/` is explicitly Mythos-rig-specific. If scope widens past his authored body, this naming convention cannot silently become general shadow policy.

## Production composition

The branch is based on repository main's 8-slot default/max-16 inventory. Production has separately carried a bounded 32-slot canary in a detached line. Review/merge does not decide that canary. Any deployment plan must explicitly preserve, retire, or reconcile it rather than silently resetting the live pool while claiming only a Mythos shadow change.

## Minor

`git diff --check` reports trailing whitespace in `tools/shadow-pref-test.mjs` at the `born normalBias` assertion.

## Required renewed evidence

At minimum:

1. explicit Mythos body/lamp capability rather than generic emissive inference;
2. competing emissive avatar/model and late-join/pool/swap tests;
3. no-Mythos exact base/head renderer-cost comparison;
4. one-Mythos cold/warm performance receipt;
5. mutation-sensitive scope/capability tests;
6. visible field acceptance of the shipped bias and GOLD exclusion;
7. explicit deployment composition with the current light-slot canary;
8. clean diff/typecheck/build neighbors.

socketteer:
Landed at `8b57412`.

## The seam

`attachLamps(root, owner, { shadows = false } = {})`. `avatar.js` passes
`shadows: true`; `realize/models.js` keeps the bare call and gets `false`,
stated in its comment as a decision rather than left as an omission. Default
false means a new caller has to say it means it.

## Your review found a regression I introduced, not just a policy gap

Worth stating plainly. `attachLamps` set `shadows: true` **unconditionally**,
under a comment about "a lamp INSIDE a body" that described only one of its two
callers — `realize/models.js:401` calls the same seam for every placed glowing
prop. So a lantern on the ground had always requested the casting slot as hard
as an avatar did.

That was harmless while the claim could only take a **free** slot. Then
`4f55321` made it a **swap** — which is what finally let Mythos's lamp cast at
all — and the same change let a nearer glowing prop *win* slot 0 off the body
standing beside it. I shipped that two commits before you asked about scope and
didn't see it.

## The red test

`tools/lamp-slot-policy-test.mjs`: one glowing avatar 3m out, one glowing prop
0.6m out. The prop outranks the body on camera distance, which is the hardest
case for the policy rather than a convenient one.

Verified red against the old default (`shadows = true` at the seam):

```
FAIL  the avatar holds the casting slot  -- avatar on slot 1, casting slot 0
FAIL  the closer prop does NOT hold it   -- prop on slot 0 — it stole the casting slot
FAIL  only the avatar wants shadows      -- wantsShadows = [lantern-1, mythos:1]
FAIL  a bare attachLamps() call requests NO shadows
7 passed, 4 failed
```

and green after (11/11). It also pins the other half of the contract — that
losing the claim costs the prop nothing: it keeps a slot and stays lit. That
half would be easy to satisfy by accident and hard to notice breaking.

## Evidence at exact head `8b57412`

- `bun run test:shadows` — **89 passed / 0 failed** (6 harnesses, 5 mutation
  controls)
- `bun run test:flight` — 231 passed / 0 failed, unchanged
- `bun run test:wingfold` — 6 passed / 0 failed, unchanged
- `bun run typecheck:flight` — clean

Noted on the field follow-up: the no-avatar and global-cost measurements stay
on my list rather than being dropped because they aren't blocking. The specific
open question is the in-frame depth compile — bodies never reach
`warmqueue.warmDepth`, so a lamped body pays its shadow pipelines on the first
shadow render, and I have measured that on exactly one body.

antra-tess (approved):
# Eidoverse Worlds PR #193 round-two review

- exact head: `8b57412d2e2277403bbcf36750d4287f5e0fef45`
- base / merge-base: `d993db107aba2b82e62def4bc806e9160e63d9c9`
- disposition: **APPROVE for source merge**

## What changed

The original branch set `shadows:true` inside generic `attachLamps()`. Both avatars and placed models call that function, so a nearer glowing prop could outrank a body and steal the only point-shadow slot.

Round two makes shadow intent explicit:

```js
attachLamps(root, owner, { shadows = false } = {})
```

- `avatar.js` passes `{shadows:true}`;
- `realize/models.js` deliberately keeps the default `false`.

Every emissive avatar may participate in the avatar-shadow policy, per Antra/Repligate's clarified product decision. Emissive placed props remain ordinary inferred lights and cannot claim slot 0.

```text
shadow follow/world ranking              8/8
XR shadow patch                          7/7
preference/camera/slot/bias             34/34
shadow mutation controls                 5/5
skinned lamp and bone-frame nudge       13/13
body-shadow gate                        11/11
avatar-vs-nearer-prop slot policy       11/11
flight                                 231/231
wing-fold semantic                      28/28
wing owner→wire                          6/6
typecheck:flight                          pass
```

The new product discriminator places a glowing avatar 3m away and a glowing prop 0.6m away. The prop outranks the body by distance, but cannot claim slot 0; it retains another slot and stays lit. Reverting the default to `shadows:true` turns the intended assertions red.

## Accepted policy and follow-up

Antra accepted body cast/receive shadows for every avatar with qualifying emissive geometry. Repligate accepted another avatar claiming the shadow slot and ruled that no-Mythos/head cost evidence is useful but nonblocking. Repligate then explicitly chose avatar-only shadow intent for the generic seam; round two implements that contract.

The PR's five mechanism repairs remain valid: preference reaches sun+lamp, body-scale point-shadow camera, shadow-wanting lamp can claim slot 0, skinned lamp follows its bone, and lamped avatars cast/receive while `GOLD` remains receive-only.

Nonblocking field/deployment follow-ups:

- measure no-avatar idle cost and first-shadow-frame compile hitch;
- measure two/three simultaneous lamped avatars;
- visibly accept the shipped 0.9375m normal bias and `GOLD` exclusion;
- explicitly preserve/retire/reconcile production's separate 32-slot canary when deployment is planned.

The fixed topology may still be globally expensive; approval records the operators' choice not to make that measurement a source-merge gate. Merge does not imply deployment.

## Minor

`git diff --check` still reports trailing whitespace in `tools/shadow-pref-test.mjs:266`. This should be cleaned, but it does not alter the accepted behavior.

PR #196, opened 2026-09-20
Fixes from VR testing: capsule stand-in body, jump/locomotion clip timing, capability card, lamps, frames

octopusburrow:
Small fixes that surfaced during VR testing but have nothing to do with a headset: a body of last resort, jump/locomotion clip timing, and four pieces of desktop chrome. 12 files, ~250 lines. No package/lockfile changes. Head: `4e13477`, base: `f0681e1` (main).

## What
- **Capsule stand-in body** (`client/lib/capsulebody.js`, wired in `main.js`/`remotes.js`/`avatar.js`): when no VRM will load (network gone, bad path), you and remotes get a VRM-shaped puppet — synthetic humanoid skeleton, a capsule per bone, `humanBones`/`normalizedHumanBones`/`update()` so every consumer (labels, emotes, reach, the C18 wire) treats it as a body. It takes the standard clips when they can be fetched; a still puppet otherwise. `?capsule=1` refuses every body load so the fallback chain can be exercised end to end. The Profile "wearing" row names it; it is never announced as worn.
- **Jump clip timing** (`avatar.js`, `controller.js`): the clip starts at its measured take-off (hips back through rest height after the anticipation dip — read off the clip's own hips track, no constants); a jump press selects the clip on its first airborne frame with a 0.1 s fade; a walk-off eases in over 0.5 s (smoothstep, weights summing to 1, starting from the weights the previous fade left). An interrupted ease fades its outgoing action instead of parking it.
- **walk→idle on key release** (`controller.js`): the clip follows intent, not the coasting speed (was ~0.34 s late).
- **Capability card** (`capnotice.js`, `index.html`): top-centre, steps below whatever occupies its span (one measurement at every width), appended before its first measure, wraps its text, attention amber, shorter copy. Frames no longer yield to the card (the card yields to frames).
- **Keyboard focus ring** is the neutral edge colour, not the attention amber.
- **Avatar lamps** (`lightrig.js`): a lamp only for a real emitter — flat emissive colour with no map, or a map driven above strength 1. MToon toon bodies (emissive ≈ white + map @ 1) no longer light the floor.
- **Frames** (`frames.js`): a placed frame still clears the dock unless its drop ended there (`state.underDock`, persisted).
- **Construct sub-grid** `0x242c36` (a shade under the plane).

## Bindings (each red on its named mutation)
- `tools/avatar-test.ts` (+7): interrupted eased crossfade (mutant: outgoing action parked → red); a landing inside the ease back into walk (mutant: reset-from-0 → walk 0.65→0.08, sum 0.41 → red); `setReach` on the real capsule puppet (mutant: no `humanoid.update()` → red).
- `tools/lamp-slot-policy-test.mjs` (+3): the emitter rule (mutant: old threshold → red).
- `tools/boot-check.mjs`: learns the capsule case (a capsule on screen must follow a reported avatar error; `BOOT_CHECK_REQUIRE_BODY=1` refuses a capsule the run did not expect); `BOOT_CHECK_VIEWPORT=1000x700` joins the recipe (red before the frames fix, green after).
- Recipe run at head: boot-check plain / `ABORT_VRM+REQUIRE_BODY` / `&capsule=1+REQUIRE_BODY` at 1280×720, 1000×700, 390×844 — all `ok`; frames-layout 76, frames-resize 23, dock 93, posture-tiles 28, emotebar 40, avatar 64, lamp-slot-policy 14, lamp-skin 13, shadow-* and body-shadow-gate green. `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` is required for the stub-mapped suites.

## Disclosed-unbound
- Capsule skeleton/render soundness beyond "a body is on screen"; the capsule clip path (the owned boot-check child 404s the clips, so under boot-check it is a still puppet).
- `clipTakeoff` numbers (0.54 s on `jump.vrma`) — probed, not bound.
- walk→idle-on-release and the jump-press fade — probed on the real controller, not bound by a suite.
- Placed-frame dock clearance / `underDock` — verified live, not bound.
- The capability card still declares `data-anchor="right"` while centred; `chromeCost` has no centre case (no longer consulted for unplaced frames).
- Remotes never ease into jump (0.1 s linear); `?capsule=1` on remotes.
- Inherited-dead at base and head: `tools/remotes-lifecycle-test.ts` (`TSL` missing from core.js); `landmarks/ragdoll/reachrig` tests need the optimised asset sweep; 844×390 "People Here covered by #dock" is red at base too.


antra-tess (changes requested):
# Eidoverse Worlds PR #196 review

Exact head: `4e13477ec513d809e1ad69836db456ba1dc3429d`

Base: `f0681e10cb4480c1848bfe64f94747fba0558cf4`

## Positive evidence

Independent exact-head results:

- avatar lifecycle: 64/64
- avatar-lamp policy: 14/14
- frame layout: 76/76
- frame resize: 23/23
- dock: 93/93
- posture tiles: 28/28
- emote bar: 40/40
- backend choice: 18/18
- flight TypeScript: pass (`tsconfig.flight.json`)
- server/agent/net bundles: pass with `--target=bun`; standing Three/WebGPU export warnings remain
- owned Chrome `?capsule=1` boot: terminal READY, real capsule on screen, no page errors
- diff/worktree clean

The capsule is shaped as a real Avatar-compatible humanoid puppet rather than a second body type; that direction is good.

## Blockers

### B1 — a transient body-load failure is terminal for the same body on that client

`remotes.js::ensureRemote()` installs `makeCapsuleAvatar(id)` while retaining the requested `avatarPath`. Subsequent announcements of the same path return the existing record; an authority takeover transplants the existing capsule into a fresh record. The body load is never retried until the avatar path changes or the participant fully leaves/rejoins.

The local path is similarly non-self-healing. After fallback, `getMyAvatarName()` still names the intended body, while the Profile list treats that name as active and gives it no `wear` action. The toast says the capsule is worn “until one does” and tells the user to pick another body, but a first-time user with only the current body cannot retry that body from the profile.

`makeCapsuleAvatar()` also returns from its clip-hydration loop on the first unavailable clip, with no later retry. A temporary asset/network failure can therefore become a session-long identity substitution or still puppet even after the asset path recovers.

Required repair/evidence: one generation-safe retry/recovery owner (bounded backoff or explicit retry action), plus a product-path test proving fail→capsule→same-body recovery, including same-path remote reannouncement/takeover and local Profile truthfulness. A forced `?capsule=1` boot proves the floor exists; it does not prove the floor has a door back out.

### B2 — the changed locomotion contract is not bound to the shipping controller path

This PR changes user-visible timing/selection policy: movement release immediately selects idle while velocity coasts; deliberate jump uses measured clip takeoff and a short fade; walk-off jump uses an eased 0.5s transition; interrupted eased crossfades preserve effective weights. The green avatar suite binds crossfade mechanics, but no focused test drives the changed controller intent path or proves takeoff measurement affects the actual selected clip.

The PR body itself lists exact jump timing, walk→idle release, and jump-press fade as unbound. These are core advertised fixes rather than incidental refactors.

Required evidence: a product-level controller/avatar test for press/release/walk-off/jump timing and meaningful red mutations removing intent selection, measured takeoff, and effective-weight preservation.

## Nonblocking / hygiene

The contributor’s advertised typecheck path `spec/flight-tsconfig.json` does not exist at this head; the working project is root `tsconfig.flight.json`. Legacy remotes-lifecycle test also has inherited mock/export debt and did not provide a clean capsule-recovery receipt.

No merge or deployment authority is implied by this review.

octopusburrow:
Repairs for both blockers, at head `2ea85c3` (base `f0681e1`). Eight commits on top of the reviewed `4e13477`.

## B1 — a transient body-load failure is terminal

Closed. The chain had five links, and the review named four of them; the fifth was underneath.

- `ensureRemote` kept the requested `avatarPath` while installing the capsule, so a record claimed a body it wasn't wearing. It now records the debt as `capsuleFor`.
- One recovery owner, `retryBody` (module-private): bounded at 3 attempts, backoff 400/1600/5000 ms, guarded by **record identity both before and after** the load — the guard idiom the rest of the file uses. A body that finishes loading for a superseded record is disposed, not adopted.
- Same-path reannounce and authority takeover both pay the debt instead of returning or transplanting the capsule forever.
- `makeCapsuleAvatar`'s clip loop skips an unavailable clip instead of `return`ing, so one 404 no longer costs idle *and* walk.
- **The fifth link:** with the `wear` action restored, `wireAvatarSwitch`'s `if (path === myAvatarPath) return` would have made the button a silent no-op — `myAvatarPath` is intent and survives the failure. It now exempts the capsule. The Profile marks nothing active while the capsule is on screen and offers `wear` on every row.

## B2 — the changed locomotion contract is not bound to the shipping controller path

Closed. The policy is extracted to `client/lib/locomotion_clip.js` as `selectClip(...) -> {clip, opts}` — the shape this repo already uses for decisions of this kind (`decideBackend`, `chooseTier`, `resolveLoadRequest`): pure, no imports, called from the real tick. Returning `opts` with the clip leaves the caller no policy to re-derive. Same thresholds, same fades, same precedence.

Your three named items:

| item | where it is bound |
|---|---|
| press/release/walk-off timing | `locomotion-clip-test.mjs` — the sequences below |
| intent selection | same — red on dropping `wantMove &&` |
| effective-weight preservation | `avatar-test.ts` (pre-existing, `4e13477`) — red on zeroing either captured weight |
| **measured take-off** | `avatar-test.ts`, **new** — see below |

### On the take-off, and a retraction

`clipTakeoff` had **zero bindings anywhere in the repo**. Deleting the measurement outright — restoring the mid-air-squat bug this PR advertises fixing — left every suite green. My previous commit closed two of your three named items and said nothing about the third, while listing six clean mutations. That reads as covered, and it wasn't. Retracted and fixed.

It is bound through the real `setClip('jump')` (the function is module-private, so its consumer is the seam), with twelve cases. Red on: `a.time = 0`; a constant return **including `0.50`, which is correct for the headline fixture** — the cases carry different shapes with different right answers; starting at the squat bottom; dropping the 1 cm noise guard; raising that guard to 10 cm (a shallow-but-real dip on a small character was being classified as noise); searching the whole clip instead of its first half; dropping the `slot === 'jump'` gate; a poisoned `userData.takeoff` memo; and `<` → `<=` in the rise-scan.

The memo case is the one worth naming: every fixture used a fresh clip, but production calls `setClip('jump')` against the *same* loaded clip on every jump, so from the second jump the cached value is the only thing that ships. Repeated jumps on one clip object now bind it.

### On "timing"

Your clause says press/release/walk-off/jump **timing**, and my first pass bound the state mapping — which clip for which inputs — with every check a single-state snapshot. That is not the same claim. Two sequences now drive it: press → hold → release → coast → stop (the clip must leave for idle on the frame the key goes up, at unchanged speed, and must not return during the ~0.3 s coast), and a walk-off frame by frame across the grace window. The walk-off sequence brackets that window from **both** sides — red on 0.04 → 0.01 (stairs flicker into a jump) and on 0.04 → 0.2 (holds too long). The snapshots only ever bound it from one.

## Two more retractions

**The retry cap was not bound, and the suite said it was.** I declared `RETRY_MAX 3 → 9999` a surviving mutant but claimed the cap's "actual observable" was bound instead. Deleting the cap outright left the suite green: at the check `n === 3`, so the pending backoff is 5000 ms against a ~2000 ms window — the timer never fires, `retrying` stays latched, and the counters freeze because of the latch, not the cap. `RETRY_BACKOFF_MS` is now injectable and the cap scenario shrinks it, so delete/→9999/→4/→2 all go red. A declared survivor with a false mitigation is worse than an undeclared one; it tells the next reader the ground is covered.

**The generation check was named for something it didn't assert.** "a body that finishes loading for a superseded record is *disposed*, not adopted" asserted only *not adopted*; removing `av.dispose()` stayed green. Now reads the flag. (A first attempt asserted on the newest body, which is the successor's own live one — measured and corrected.)

## Recipe at head

`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` required.

| suite | |
|---|---|
| avatar | 76 |
| capsule-recovery (new) | 34 |
| bodies-wear (new) | 11 |
| locomotion-clip (new) | 31 |
| posture-tiles | 28 |
| wing-owner-wire | 6 |
| settings-panels | 73 |
| wing-fold-presence | 28 |
| frames-layout / frames-resize | 76 / 23 |
| dock | 93 |
| emote bar | 40 |
| lamp-slot-policy / lamp-skin | 14 / 13 |

`posture-tiles` is worth singling out: it drives the **real `updateMe()`**, and its own mutation list already covers "updateMe's seatedClip losing the chair branch" — so it independently exercises the code the extraction moved.

Owned browser runs, all `ok` at head: plain; `ABORT_VRM` + `REQUIRE_BODY` (capsule on screen after a real avatar error, no page errors — this is `retryBody` running live against a dead asset); `&capsule=1` + `REQUIRE_BODY`; `BOOT_CHECK_VIEWPORT=1000x700`.

## Disclosed-unbound

- **Declared surviving mutants, with reasons.** Dropping the *pre*-load identity guard: equivalent under this suite, since every superseding scenario it drives is caught one line later by the post-load guard, which is bound. Deleting the `userData.takeoff` memo *write*: the function recomputes and returns the same value, so it is a perf regression, not behavioural — poisoning it *is* behavioural and is bound.
- The local fallback has no automatic retry owner, only the explicit `wear` action. Your requirement said "bounded backoff **or** explicit retry action"; the asymmetry with the self-healing remote path is deliberate, not accidental.
- `__setRetryBackoffForTest` is a test-only export on `remotes.js`. Called out rather than left for you to find.
- Red at base **and** head, unchanged by this PR: `remotes-lifecycle-test.ts` (`TSL` missing from `core.js`), `contact-mutation-test.ts`, `parts-test.ts`. Verified at `f0681e1`.
- Hygiene from your review, still true: the advertised `spec/flight-tsconfig.json` does not exist; the working path is root `tsconfig.flight.json`.

Two further bugs were found and fixed while repairing these — the retry bound resetting on every authority takeover (a flapping peer with a dead asset refetched at the 400 ms floor indefinitely), and a dead `export` on `retryBody`. Both are in the log with their own bindings.

No merge authority assumed; #197 stays blocked on this landing.

antra-tess (approved):
APPROVE — exact rereview head `2ea85c3e47d1ece71f18e497397d9f3890f7ae4b`.

Both prior blockers are closed:

- **Capsule recovery:** failed loads retain bounded, generation-safe body debt; same-path reannouncement/takeover and local same-body wear can recover to the real body; stale loads dispose; retry count survives takeover; clip hydration skips individual misses. Independent: recovery `34/34`, Profile/wear `11/11`, owned forced-capsule boot PASS. Removing real same-path retry wiring made recovery `25/34`.
- **Locomotion timing:** the shipping controller calls one pure policy owner for press/release/coast/jump/walk-off timing and forwards its opts directly; measured takeoff and interrupted-ease behavior are independently bound. Independent: locomotion `31/31`, Avatar `76/76`. Replacing the real controller call with constant idle made locomotion `30/31`.

Neighbors/build: dock `93/93`, posture `28/28`, emote bar `40/40`, backend choice `18/18`, flight TS clean, all three Bun-targeted server/MCPL bundles pass, diff/worktree clean.

Artifact: `files/research/eidoverse-pr196-r2-review-2ea85c3.md`
SHA-256: `f0f6cd118cef874144883ba84918a7328d68da108e21e358fe3a5783c32fd39b`

Source-merge approval only; no deployment standing inferred.

PR #180, opened 2026-09-09
Rendering + boot: VR-safe shadows, splash gate, asset retry, perf scope (2 of 4)

octopusburrow:
# Rendering + boot: VR-safe shadows, splash gate, asset retry, perf scope (2 of 4)

Part 2 of the VR-alpha stack (on top of #1). Renderer/boot-side changes VR needed and desktop benefits from; plus the small plumbing modules the later rungs import (`base.js` tee, `presence.js`, `bodygate.js`, `shared/presencewire.js`, `icons.js`).

## What's in it
- **Shadows under per-view rendering (`lightrig.js`, `shadowview.js`):** the caster pass and the shadow box follow the camera's *world* position (rig-local after XR entry put the box at the origin: no shadows on objects in VR); near-plane clamp; a resolution setter (`setShadowRes`, wired to a Settings row in part 3); `?csm=N` / `?shadowview=1` diagnostics (module lands in part 3 with its mount).
- **Materials (`materials.js`, `sky*.js`):** TSL camera accessors are invalid in a material that renders per-view — the wetness wrapper's puddle term used one and the ground never drew in VR. View-space quantities now.
- **Frame clock (`frame.js`):** the first tick after an XR session ends arrives with a non-finite `now`; a resume is not a frame.
- **Render (`render.js`):** `renderAside()` for bakes/thumbnails — XR off around the pass, target restored, eyes rebuilt from the rig; a per-frame self-heal for a stranded frame-buffer target.
- **Assets (`assets.js`, `bodygate.js`):** `fetchBytes` retries 5xx/network (3 tries) and evicts a rejected promise from the cache. The body-first gate (world parses wait for your own body) is *armed* by the entry that requests a body (part 4) — unarmed, it costs nothing, so this rung never waits on it.
- **`core.js`:** one renderer control (`?webgl=1` / `?webgpu=1` / the `ew-backend` pref); a render-list tolerance loop exported for VR entry (installed at `?xr=1` boot or right before a session request — never for a desktop session).
- `warmqueue.js` shadow-depth pre-warm, `governor.js`, `net.js`, `avatar.js` (portrait pose race, own-label hook), `conjure.js`, `draw_batches.js`, `scenegraph.js`, `terrain.js`.

## A dependency note reviewers should know
[...]

PR #197, opened 2026-09-20
VR (alpha): WebXR entry/exit, body in VR, arm/leg solvers, curtain, pre-warm (part 4 of 4; stacked on #196)

octopusburrow:
Part 4 of 4: VR (alpha). WebXR entry/exit, a body in VR, turning, pointing, panels in VR, and an experimental desktop mirror. Stacked on the fixes PR (#196) — it depends on the capsule puppet for its solver tests. 20 files, ~2.7k lines (`xr.js` 1.2k, `xrbody.js` 0.6k). No package/lockfile changes. Head: `e4ae4cc`, base: `4e13477` (fixes-2026-09-19).

## What
- **Entry/exit** (`xr.js`): visor button and ring menu; a 2D veil and toasts from the click (with a paint yield so they actually draw before the session request); an in-headset curtain wearing the splash (∃, `eidoverse / worlds`, breathing dots, level with the horizon) held until the entry compile resolves; the whole scene compiles under it with `frustumCulled` off (props behind you were being built at their first in-session draw — measured 20 sync builds / 6.7 s in the first 3 s → 5 / 0.13 s); a pipeline pre-warm on any headset-present boot, restoring the desktop target before awaiting the link; `ew-headset-seen` written when a session is granted so the next boot picks WebGL and the visor enters in place; an `entering` guard; a probe that names every sync pipeline build over 30 ms.
- **Body** (`xrbody.js`): head/hands drive the VRM; a swivel-circle arm solver (36 samples, refine 8, torso-built frame, prior/previous/torso costs, min elbow 22°, soft reach, wrap fade, roll cap 120°), forearm-roll wrist split, hold-then-relax on tracking loss; chest follows the hands; leg IK with a gait/step model, gated: animation owns the legs while locomoting or airborne, foot IK takes them after 0.15 s stationary (in at 20/s, out at 15/s, slerp(animated, solved, w)), re-seeded from the animated feet each engage; a planted foot follows the ground; the torso latch resets on a body swap.
- **Fixed foveation** platform split: 1 on a standalone UA, 0 on PC, `?fov=` override.
- **Turn, pointer, panels, keyboard, vignette, mirror**: as cut from the alpha branch; mirror is experimental (WebGL only, self-disables after 30 slow frames).
- **`avatar-worn` normalised** through one `wornNameOf` (the per-body VR scale was keyed `[object Object]`).

## Bindings
- `tools/armsolve-test.ts` (23): solver reach/continuity/roll/hold/torso, `solveLeg` on the capsule puppet, gait ground-follow. Each of these mutations turns a named check red: forearm-roll write removed; hold-then-relax removed; `_pole` scratch deleted; `gaitTick`'s `f.p.y = des.y`; torso penalty off; roll = 0.
- `tools/worn-name-test.mjs` (4): the payload normaliser (red on `return v`).
- `tools/boot-check.mjs` with `&xr=1`: XR boot rides WebGL, arrives, body on screen.
- Recipe at head: boot-check plain / `&xr=1` / `ABORT_VRM+REQUIRE_BODY` / 1000×700 all `ok`; armsolve 23, avatar 64, settings-panels 73, worn-name 4, bodygate 9, xrshadow 7, shadow-follow 8, body-shadow-gate 11, foldfix 24, posture-tiles 28, dock 93, wing-fold-presence 28, wing-owner-wire 6. `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` required.
- Pre-warm measured headless: 0 desktop frames rendered into the warm target after the body lands (was 4–8 in an earlier shape), pipeline cache 26→37.

## Disclosed-unbound
- **Headset-only:** session entry/exit on a real runtime and the WebGPU→WebGL reload ladder; the curtain and its timing; the pose watchdog (counts from grant; the sync remainder of `enterVR` after grant is unmeasured); foveation; controller binding; the entry clock and the sync-build probe (formats exist, values only from a headset); `xr.enabled` stays true after exit (three gates on `isPresenting`).
- **Bound only on a synthetic rig:** the arm/leg solvers and gait (armsolve-test). Not bound: swivel smoothing / 720°/s rate cap; soft-reach → hard-clamp; `feetTick`'s locomotion gate; the hips latch; `applyRemoteXR`; the wire end to end.
- **Experimental:** mirror, keyboard, vignette.
- **Consumer-less at this rung:** `xr:prefs`, `xr:scale`, `xr:recentre` events; `xrRecentre()`, `xrScale()`, `xrFloorSpace()`, the `?xrsim` hooks.
- **Page-lifetime instruments installed on first enter, never removed (deliberate):** backend `createProgram`/`createRenderPipeline` wrap, the console tap (gated on `presenting`), the render-list tolerance patch (installed before the request, survives a rejection).
- **Non-VR behaviour riding in `main.js`, inherited from the alpha cut (named, not moved):** `?door=1`, `?sendlayout=1`, the body fallback chain clearing `ew-avatar-path`, `checkReady` waiting on `pending(P.FAR)`, the body gate armed (recorded dormant in #180), the presence dot, the settings dock tile, the `[body] on screen` tee.
- **Inherited-dead at base and head:** `remotes-lifecycle-test.ts` (`TSL`), `reach-order-test.ts`, `parts-test.ts`, `surface-hold` 10/12, `micstate-single-transport` 9/1.
- NITs left: `disposeRadial` leaks one plane geometry per ring open; two writers of `ew-headset-seen` on one granted path; the vignette plane stays on the desktop camera at opacity 0 once used; the curtain is a re-entry cache, never disposed.


antra-tess (changes requested):
# Eidoverse Worlds PR #197 review

Exact head: `e4ae4cc5758b85278c3f9451725bd9aed2e382f0`

Stacked parent: PR #196 head `4e13477ec513d809e1ad69836db456ba1dc3429d`

Current main/base: `f0681e10cb4480c1848bfe64f94747fba0558cf4`

## Positive evidence

Independent exact-head results:

- arm solver: 23/23
- worn-name identity: 4/4
- avatar lifecycle: 64/64
- settings panels: 73/73
- body gate: 9/9
- XR shadow patch: 7/7
- world-space shadow follow: 8/8
- body-shadow gate: 11/11
- fold semantics: 24/24
- posture tiles: 28/28
- dock: 93/93
- wing fold presence: 28/28
- wing owner→wire: 6/6
- backend choice: 18/18
- flight TypeScript: pass (`tsconfig.flight.json`)
- server/agent/net bundles: pass with `--target=bun`; standing Three/WebGPU export warnings remain
- owned Chrome `?xr=1` boot: terminal READY on WebGL, `xr.enabled=true`, tolerant render list installed, no page errors
- diff/worktree clean

These bind substantial pure/state-machine and non-XR neighboring behavior. The boot check is a boot flag check; it does not request an immersive session.

## Blockers

### B1 — “retrying once” on a busy immersive session retries forever and ignores stale intent

In `client/lib/xr.js`, the `InvalidStateError` path sets `enterVR._retried = true`, schedules a timeout, then resets `_retried = false` immediately before recursively calling `enterVR()`. If another page still owns the session, the next failure sees `_retried === false` and schedules again. The documented “busy after retry — giving up until the visor is clicked again” branch is therefore unreachable in the ordinary repeated-failure path.

The timeout also has no entry-intent/session generation and no cancellation handle. A retry created by an old visor click may request immersive VR 1.5 seconds later after user intent/state has changed.

Required repair/evidence: exactly one automatic retry; second busy failure remains stopped until another explicit visor action; leave/cancel before the timer prevents delayed entry; stale retry cannot act on a later generation. This should be a mutation-sensitive lifecycle test, not source text matching.

### B2 — the central WebXR lifecycle remains unexercised

The owned `BOOT_CHECK_QUERY=&xr=1` run proves backend construction and boot decisions only. It never calls `navigator.xr.requestSession()`. No committed product test exercises the feature’s core lifecycle:

`requestSession → setSession → viewer pose → curtain compile/drop → end → desktop frame recovery → re-enter`

The PR body explicitly leaves actual session entry/exit, WebGPU→WebGL reload, curtain timing, pose watchdog, foveation, controller bindings, mirror/keyboard/vignette, and end-to-end XR wire unbound. For a 2,600-line WebXR alpha, pure solver tests and an XR-flag boot do not establish that the door opens or closes safely.

Required evidence: an owned fake-XRSession product harness covering success, denied/no-device, busy retry, setup failure after grant, session end, re-entry, stale async completion, and teardown; plus a named real-headset field receipt (device/browser/backend) for entry, body, controller/pose, exit, and re-entry. Headset-only visual quality may remain alpha debt, but lifecycle authority cannot.

### B3 — the persisted “headset seen” bit is presented as current presence and has no decay/clear path

A granted/posed session sets `ew-headset-seen=1`; no code clears it when the headset is removed or later unavailable. Auto backend then remains WebGL forever on browsers without WebGPU-XR, even when no headset is currently attached. The Video panel tells the user “Right now: a headset is present,” which the stored bit does not prove.

Required repair: call the state what it is (“a headset has been used here”), provide a clear/expiry/re-probe policy, or avoid using historical presence as permanent auto-backend authority. Add a persistence lifecycle test.

### B4 — stacked parent unresolved

PR #197 includes all of PR #196. PR #196 is changes-requested for terminal capsule fallback and unbound locomotion timing. PR #197 cannot be independently merge-ready until its parent is repaired/rebased and composed evidence is rerun.

## Follow-up debt

`disposeRadial()` disposes radial slot geometries/materials and label material but appears not to dispose the label plane geometry; repeated ring opens can leak one geometry per open. Page-lifetime debug/render-list wrappers are intentionally not removed and should remain explicit alpha debt.

No merge or deployment authority is implied by this review.

octopusburrow:
Three of four blockers closed, at head `1132f7a`. Rebased onto #196's **approved** head `2ea85c3` — B4's precondition — and the composed evidence is rerun below rather than inherited.

One open question for you at the end, on B2's scope. I would rather ask it than decide it.

## B4 — stacked parent

Closed as far as I can take it. #197 was stacked on `4e13477`, the head you'd requested changes on, so its diff carried a version of #196 you had already rejected. Now rebased onto `2ea85c3`.

**The rebase replayed all 26 commits with zero textual conflicts and still broke two of #196's suites** — which is the argument for your "composed evidence is rerun" wording, so it is worth naming what it caught. #197's `remotes.js` reaches further than #196's did: it imports `xrbody.js` (which pulls `frames.js`, and a DOM) and `shared/presencewire.js`; its `assets.js` cone wants `poseclips`' parser; `bodies.js` gained `wornNameOf`. None of that is visible from either PR alone. Stub surfaces completed — enumerated in one pass from every `import { … } from './<module>.js'` across `client/lib`, not discovered one SyntaxError at a time.

## B1 — "retrying once" retried forever

Closed. Your reading was exact: `enterVR._retried` was reset to `false` **inside** the timeout, immediately before the recursive call, so every subsequent failure saw a fresh flag and the documented give-up branch was unreachable.

- Exactly one automatic retry per **entry intent**. `enterVR({ retryOf })` carries the intent forward, so the retry's own failure is the second one and falls through to give-up.
- The pending retry is generation-checked at fire time and **cancellable**: a new entry intent cancels the previous one's, and `leaveVR` cancels before its no-session return — that ordering is the fix, since a pending retry means there is no session yet and a leave would otherwise drop through and let VR enter afterwards.
- What a failed request *means* is now a pure decision in `client/lib/xr_entry_policy.js` (busy / absent / reload-to-WebGL / surface); `xr.js` keeps the effects.

`tools/xr-entry-policy-test.mjs` (25). Red on: always returning retry-once (the shipped bug); busy detection testing only the error **name** and not its message (an unrelated `InvalidStateError` would otherwise retry); dropping absent detection; `retryIsCurrent` ignoring staleness; `xr.js` no longer consulting the policy; `leaveVR` no longer cancelling.

## B3 — history presented as presence

Closed, and it was wider than the review said: the false present tense appeared in **four** places, not one — both auto-tail branches, the no-headset branch ("Right now: no headset sensed" is equally a present-tense claim from the same historical bit), and the force-WebGPU warning.

A live re-probe is not an honest fix either — `mictoggle.js:91` already records that `isSessionSupported` stays optimistic after a headset is switched off. So the bit keeps its job and gains a clock: `client/lib/headset_seen.js`, 30-day TTL, legacy `'1'` honoured until the next granted session re-stamps it. Both writers stamp `Date.now()`; the boot choice asks the predicate; the copy says "A headset has been used here before."

`tools/headset-seen-test.mjs` (19). Red on: ignoring the TTL (once seen, always seen — the bug); dropping legacy `'1'`; shrinking the TTL to a day; the boot choice reading the raw bit again; the copy returning to present tense; a writer stamping `'1'` again.

*Stub audit:* `videopanel.js`'s new named import made `settings-panels-test` red (`headsetSeenRecently` not found in its stub). Added to `core-stub.mjs`, which the panel stub re-exports.

## B2 — partially closed, and a question

**Done: the teardown half**, which needed no session object and had a real documented failure. `client/lib/xr_frame_clock.js` + `tools/xr-frame-clock-test.mjs` (24) bind who owns `window.requestAnimationFrame` across a session:

- the restore listener is registered **before** `setSession`, so it runs before three's own `end` listener restarts the desktop loop — otherwise that restart goes to the dead session and no frame ever ticks again (the symptom `xr.js` records at 09-07 22:41: after-exit frames+0 at both +0.5 s and +3 s). The suite asserts that ordering by source position.
- the `sessionEnded` latch, so a call between `end` and the restore falls through to the native clock;
- **save-once** across the page: a per-entry capture saves the previous entry's shim as "native", and the restore then hands frames to a dead session.

Red on all five. Worth naming: the save-once mutation **survived the first version of its own test**, because that fixture restored the clock between entries, so the re-capture read an already-native clock and looked harmless. It only bites when a second entry installs its shim while the first is still in place — a re-enter before teardown completed. Measured, then written as its own scenario.

**Not done: the entry/pose half, and I want your call rather than my guess.**

I can build the fake-XRSession harness you asked for, but only one way: IWER (the Immersive Web Emulation Runtime) inside Playwright's headless chromium — the same seam `boot-check.mjs` runs through `probe-harness.mjs`. I verified it works there (`navigator.xr` installed, `isSessionSupported('immersive-vr')` → true). It would arrive as a `*-probe.mjs` beside the other browser probes, and it would add `iwer` as a devDependency (plus two transitive deps, across five lockfiles).

What stops me committing to that unilaterally is that headless chromium is SwiftShader, and this repo's own `vr-lab.mjs` header records that headless never reproduced the real in-headset symptoms. So such a probe could show the **lifecycle** — requested, granted, posed, ended, re-entered, clocks restored — and could not show anything about rendering. Given you drew the line at "headset-only visual quality may remain alpha debt, but lifecycle authority cannot," I think that probe is on the right side of your line, but I am not confident enough to spend a dependency and 400 lines on my own reading of it.

**So:** do you want that probe, and would it count as lifecycle authority? If you would rather the lifecycle stay disclosed-unbound until the headset receipt exists, that is a cheaper answer and I will take it.

**The real-headset field receipt is outstanding and is not mine to produce** — it needs a person in a headset. It is scheduled; I will post it as a named receipt (device / browser / backend, entry / body / controllers / exit / re-entry) when it exists, against this head rather than an older build.

## Recipe at head

`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` required.

| suite | |
|---|---|
| xr-entry-policy (new) | 25 |
| headset-seen (new) | 19 |
| xr-frame-clock (new) | 24 |
| capsule-recovery | 34 |
| bodies-wear | 11 |
| locomotion-clip | 31 |
| avatar | 76 |
| armsolve | 23 |
| settings-panels | 73 |
| worn-name | 4 |
| xrshadow | 7 |
| xrpanels-contract | 10 |

## Disclosed-unbound

- **The entry/pose/compile half of the session lifecycle**, per the question above. The teardown half is bound; entry is bound only at the policy level (what a failure *means*), not through a real `requestSession`.
- **Everything headset-only** from the original body still stands: the curtain and its timing, the pose watchdog, foveation, controller binding, the WebGPU→WebGL reload ladder, the entry clock.
- `__setRetryBackoffForTest` is a test-only export on `remotes.js` (it is how the retry cap became observable — with the shipped 5 s third backoff, a suite that deleted the cap outright still passed). Called out rather than left to find.
- Red at base **and** head, unchanged by this PR: `remotes-lifecycle-test.ts` (`TSL` missing from `core.js`), `contact-mutation-test.ts`, `parts-test.ts`. Verified against `f0681e1` in a clean archive.
- This PR now carries **no** new dependencies. That changes if you want the IWER probe.

No merge authority assumed.

octopusburrow:
## State at head `7094a16` (was `1132f7a`)

A headset pass happened, and an agent review round after it. **Two corrections to my previous comment first**, since both are checkable in one command:

- **"replayed all 26 commits"** → **25**. `git rev-list --count 2ea85c3..HEAD` is 26, but the 26th (`1132f7a`, "stubs: the composed cone") was authored three minutes *after* `rebase (finish)`; the reflog shows 25 `rebase (pick)` entries. It was written, not replayed. The zero-conflicts half stands.
- **"`bodies.js` gained `wornNameOf`"** → `wornNameOf` is defined at `base.js:222`; `bodies.js` gained an **import** of it (`bodies.js:7`, used at `:22`). The surrounding sentences are all about where a symbol lives, so that one reads wrong.

## The field receipt you asked for — B1/B2/B3 exercised in a real headset

Device: Quest over a tunnel, **RTX 4080 host**, Chromium, WebGL backend, against `1132f7a` (the two commits since touch only `xr.js`'s failure branch and two suites). Evidence is the client tee: `worlds/.clientlogs/clientlog-commons.log`, 36 `[xr:rec]` samples over ~5 minutes.

- **Entry, exit, re-entry: all clean.** `enter #2` resolved `setSession` in 1 ms. No retry storm — the B1 regression does not reproduce.
- **Desktop frames after exit:** the symptom `xr.js:579` records (frames+0 at +0.5 s and +3 s) did not recur. B2's teardown half is now bound by a field observation as well as a suite.
- **Steady state:** fps 57–64, `ms` 15–19, `worst` 33, `spikes` 0, body visible, both controllers tracked, `headChop` correct.
- **B3's copy** reads past tense in the four places listed below.

**Separately, and NOT a regression:** the desktop mirror ("mirror my eyes") is black. It was black before this PR. Diagnosed but not fixed here — the blit path reaches for three's private frame-buffer target and returns success whenever that texture merely *exists*, so it tees success over a black canvas. Retired to our bug tracker; no code in this PR touches it.

**One real defect found in passing, filed and not fixed here:** the desktop rAF can starve (frame updates every 3–4 s) while `[xr:rec]` still reports `fps=60`, because that is the XR loop's number. It reproduces with the mirror **off**, and it pre-dates this PR (spikes at 21:05 and 21:07, before any change). `xr_frame_clock.js` owns exactly this question for *teardown*; the entry/steady-state side is plausibly the same territory and is **not** covered by its suite. Calling it out rather than leaving it to be found.

## What the review round changed

Four concurrent agents (cold diff read; do-the-tests-bind; receipt audit; consumer-side wiring). It was not a clean round.

### The finding worth your attention: two agents, same gap, opposite conclusions

`decideEntryFailure`'s `gpu`/`absent` ordering is only observable for an error that is **both** absent-shaped **and** on a WebGPU backend. No check drove that pair, so swapping the two branches passed **25/25 green**. One reviewer called the shipped order a bug; another called the swap the bug.

Resolved on the mechanism, not the vote: **`webgpu` is a `requiredFeatures` entry** (`xr.js:527`), and an ungrantable required feature rejects with **`NotSupportedError`** — which `isAbsent` matches. Absent-first would therefore read *every* WebGPU refusal as "no headset" and mark a **present** headset absent. The shipped order is correct; what was missing was a test saying why.

I had already applied the absent-first change before checking. Reverted. Recording it because the disagreement is what caught it — a unanimous round would have shipped it.

### `xr.js` now acts on the verdict (`38430a6`)

The review found `xr.js` honoured only **2 of 4** verdicts. `absent`, `reload-webgl` and `surface` were re-derived inline at the old `:556`, duplicating `isAbsent`'s two error names **and its regex verbatim**, plus a hand-copied toast string. Two copies of a predicate drift.

Now it reads `verdict.action` and owns only the effects. **Verified identical across 10 error/backend pairs** (NotFound / NotSupported / Security / Data / generic-with-absent-message × gpu on/off) — a de-duplication, not a behaviour change.

### Three assertions that could not fail (`7094a16`)

28 mutations were applied; these survived green on suites claiming to cover them:

1. **`clockIsRestored` gutted to `return true` passed.** It was only ever called in the already-restored state, so no term was load-bearing — including the `cancelAnimationFrame` term, the only place cancel-restoration was checked at all. Now pins the negative first: not-restored, rAF-only-restored, and a null native each assert false before the positive.
2. **`isAbsent`'s two arms masked each other.** Both fixtures carried a matching *name* **and** a matching *message*, so deleting the message regex entirely passed. The message arm exists for runtimes that throw a generic name with a descriptive message — never constructed. Both arms now pinned alone.
3. **The frame shim's `(t) => cb(t)` wrapper was never exercised.** The fake session collected callbacks and never invoked them, so dropping the timestamp — or swapping time and XRFrame — passed. The fake now `tick()`s and the suite asserts what the callback *receives*, including that the XRFrame is dropped deliberately (the shim stands in for `window.requestAnimationFrame`, whose callback takes one argument; no desktop consumer reads a second).

## Recipe at head `7094a16`

`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` required. All 12 exit 0 with zero `✗`:

| suite | | suite | |
|---|---|---|---|
| xr-entry-policy | **31** (was 25) | avatar | 76 |
| headset-seen | 19 | armsolve | 23 |
| xr-frame-clock | **29** (was 24) | settings-panels | 73 |
| capsule-recovery | 34 | worn-name | 4 |
| bodies-wear | 11 | xrshadow | 7 |
| locomotion-clip | 31 | xrpanels-contract | 10 |

**Mutation corpus: 11/11 RED, 0 mismatched, 0 skipped.** Eight from the original claims plus the three above that previously survived. Runner refuses a dirty tree and restores on every exit path.

**Browser gate:** `tools/boot-check.mjs` green against a live server at this head — `[boot] ready` for reason `ready`, splash gone, 19 panels, body on screen, **no page errors**.

**B3's four past-tense sites**, all `videopanel.js`: `:110`, `:111`, `:112`, `:127`; `:108` asks `headsetSeenRecently()`. One cosmetic straggler at `:118` ("if a headset **is in use**", present tense in the force-WebGPU description) — not covered by the suite, not fixed, named here.

## Disclosed-unbound

- **The entry/compile half of the lifecycle.** Entry is bound at the policy level (what a failure *means*) and now by one field observation; not through a real `requestSession` in CI. The IWER probe remains **your call** — my earlier note said it would touch "five lockfiles"; that is wrong, it touches **one** (root `package.json`, beside Playwright). `iwer` is MIT, v2.4.0, two small deps. Cost is not a reason to refuse it; whether a SwiftShader probe counts as lifecycle authority still is.
- **Everything headset-only** from the original body still stands: the curtain and its timing, the pose watchdog, foveation, controller binding, the WebGPU→WebGL reload ladder, the entry clock. Headless is SwiftShader and this repo's `vr-lab.mjs` header records that it never reproduced the real in-headset symptoms.
- **Three exports with no product consumer**: `retryIsCurrent`, `clockIsRestored`, `headsetSeenAt`. `clockIsRestored`'s docstring says it is for tests. `retryIsCurrent` is not — `xr.js:548` hand-rolls the same check inline as `busyRetryFor !== myEntry`, **without** the exported function's `pendingFor !== null` guard. Benign today (`myEntry` is never null) but they can drift. Not changed in this pass.
- `videopanel.js:15` imports `PREF_HEADSET_SEEN` and no longer uses it. Dead import, left alone to keep this diff to the review's findings.
- `__setRetryBackoffForTest` is still a test-only export on `remotes.js:63`.
- **Red at base *and* head, unchanged by this PR**: `remotes-lifecycle-test.ts` (`TSL` missing from `core.js`), `contact-mutation-test.ts`, `parts-test.ts`. Re-verified at `2ea85c3` from a clean `git archive` — note that run needs `node_modules` and `client/node_modules` symlinked in, or it fails for an unrelated reason.
- **No new dependencies.** `git diff --stat 2ea85c3...HEAD -- '*package.json' '*lock*'` is empty.

No merge authority assumed.

octopusburrow:
## `89e2e55` — finger curl is filtered (found and fixed during the same headset session)

One more, small and in scope: **VR hand pose jittered.** Reported in-headset as *"a subtle buzzing jitter on the fingertips"*, with the clarification that button **actuation** was correct — partial and full presses both registered. So this is the analog pose, not the `pressed` threshold, and it wants smoothing rather than hysteresis.

`sampleFingerCurl` assigned the raw trigger/grip `.value` straight through every frame, with none of the conditioning the thumbstick already gets from `DEADZONE`/`dead()`. A resting finger plus the controller's own ADC noise reached the hand pose directly.

**It also rides the wire.** `xrbody.js:554` packs all four curls into `wire.c` for every remote viewer, so the jitter was broadcast — quivering on other people's view of you, and spending bandwidth on noise. Filtering at the source fixes both ends.

### The first fix was wrong, which is the useful part

A plain one-pole low-pass with a jump threshold — `|raw - prev| > 0.12 ? raw : lerp` — tested **worse** in the headset: *"big angle jumps between poses… a bit clicky on big moves."*

A hard threshold is a **discontinuity**. The output alternates between heavily smoothed and teleported. Measured at 72 Hz on a 250 ms squeeze, the per-frame deltas repeat:

```
0.006  0.011  0.149  0.006  0.011  0.149  …
```

A 13× step ratio on a 3-frame cycle. And raising the time constant made it **worse**, by widening the gap between the two regimes.

Replaced with the **1€ filter** (Casiez, Roussel & Vogel, CHI 2012), whose cutoff rises *continuously* with the signal's own speed — heavy smoothing at rest, light while moving, no branch to click on. Peak **jerk** (frame-to-frame change in step size, which is what reads as stutter), 72 Hz:

| | 150 ms squeeze | 250 ms squeeze |
|---|---|---|
| hard threshold | 0.167 | 0.149 |
| 1€ | **0.055** | **0.035** |

About a third of the stutter, same time to 90%. Confirmed in the headset: *"so much better."*

`dt` is clamped so a stall cannot teleport the pose, and the speed estimate is itself low-passed (`CURL_D_CUTOFF`) — without that, the derivative's own noise opens the cutoff and the jitter returns (measured: ±6% noise leaves 0.61% residual smoothed, **2.88%** unsmoothed).

**A methodology note, because it nearly sent me the wrong way.** My first smoothness metric — max step ÷ median step — reported the 1€ filter as *122× worse*. That metric was measuring **lag, not smoothness**: on a fast input ramp the input itself has large steps, so a faithful filter scores badly and a laggy one scores 1.0. Jerk is the right measure. Recording it so the constants are not re-tuned against the bad one.

### Tuning is measurable, not felt

`?curlprobe=1` tees the **raw** per-frame jitter into the flight recorder every 5 s (`[xr:curl] raw jitter peak/mean/n per finger`). The constants in this commit were chosen against a noise model; anyone changing them should read the real number off their own controllers first.

### Verification

- `tools/finger-curl-test.mjs` (**22**). The module cannot be imported (`xr.js` pulls three, the renderer and the DOM), so the suite pins **both** lines of the product's filter by text and runs a model of them — a mutation to either turns it red even though the behavioural checks run the copy. That gap was real and found by mutation, not by inspection.
- **Mutation corpus: 16/16 RED, 0 mismatched** across the whole rung — including **restoring the hard-jump branch**, so the clicky regression cannot come back silently, and **dropping the filter entirely**, i.e. the originally shipped behaviour.
- Full matrix re-run at this head, **13 suites, all exit 0, zero `✗`**: finger-curl 22, xr-entry-policy 31, headset-seen 19, xr-frame-clock 29, capsule-recovery 34, bodies-wear 11, locomotion-clip 31, avatar 76, armsolve 23, settings-panels 73, worn-name 4, xrshadow 7, xrpanels-contract 10.
- `tools/boot-check.mjs` green against a live server at this head — `[boot] ready` for reason `ready`, splash gone, no page errors.
- No new dependencies.

### Still unbound

The filter constants are tuned against a noise **model** plus one headset session's judgement, not against a logged measurement from a specific controller — `?curlprobe=1` exists to close that when someone cares to. And the wire benefit (calmer hands for remote viewers) is inferred from the source change; it was not observed with two people in-world.

antra-tess (changes requested):
# Eidoverse Worlds PR #197 — round-two rereview

- PR: <https://github.com/anima-research/eidoverse-worlds/pull/197>
- Title: `VR (alpha): WebXR entry/exit, body in VR, arm/leg solvers, curtain, pre-warm (part 4 of 4; stacked on #196)`
- Exact reviewed head: `89e2e5592a5c6f21780cdd63c1d593c1e208dc56`
- GitHub base: `f0681e10cb4480c1848bfe64f94747fba0558cf4`
- Stack prerequisite: repaired/approved PR #196 head `2ea85c3e47d1ece71f18e497397d9f3890f7ae4b`

## What improved

The repaired head materially addresses all three prior review themes in source design:

1. busy-session retry now has an explicit pure policy, one intended retry, an entry-intent generation, a cancellable timer, and leave/new-click invalidation;
2. frame-clock ownership is separated into a pure module with save-once, ended-latch, throwing-session fallback, cancellation on both clocks, and re-entry semantics;
3. headset history now stores timestamps, expires after 30 days for timestamped values, and uses past-tense UI copy;
4. analog finger curl now uses a time-based 1€ filter with derivative smoothing, dt clamp, per-finger state, and bounded output.

The repaired parent (#196) is included in the stack and its capsule/locomotion evidence remains green.

## Independent exact-head positive evidence

```text
xr-entry-policy-test           31/31
headset-seen-test              19/19
xr-frame-clock-test            29/29
finger-curl-test               22/22
armsolve-test                  23/23
worn-name-test                  4/4
capsule-recovery-test          34/34
bodies-wear-test               11/11
locomotion-clip-test           31/31
avatar-test                    76/76
flight typecheck               PASS
server Bun bundle              PASS
stdio MCPL Bun bundle          PASS
network MCPL Bun bundle        PASS
diff/worktree                  CLEAN
```

The bundles retained standing Three/WebGPU export warnings from the linked dependency closure; they did not fail the Bun server targets.

The contributor also disclosed that the full compile/curtain/body/headset path is only partly automated and still needs real-headset evidence.

## Blocker 1 — the retry test does not bind the product branch

I mutated the shipping `client/lib/xr.js` integration so every busy verdict, including the second `give-up` verdict, entered the retry-scheduling branch:

```js
if (true) { // MUTANT: every busy verdict retries, including give-up
```

The committed suite still passed:

```text
xr-entry-policy-test 31/31
```

Why: the test drives `decideEntryFailure()` correctly, then source-checks only that `xr.js` imports/calls it, uses `verdict.delayMs`, carries an intent generation, and cancels from leave/new intent. It never executes the real `enterVR()` catch/effect path or proves that a `give-up` verdict does not schedule again.

This is precisely the prior bug class: the policy can be correct while product wiring retries indefinitely.

Required repair: an executable product-path harness around the real entry lifecycle (or an extracted effect owner imported by `xr.js`) that proves:

1. first busy failure schedules exactly one delayed retry;
2. second busy failure performs no scheduling;
3. leave/new intent cancels the pending retry;
4. a stale callback cannot request a session;
5. explicit click after give-up starts a fresh budget.

The integration-breaking mutation above must turn that receipt red.

## Blocker 2 — the frame-clock test does not bind product installation

I mutated shipping `xr.js` so the product never installed the session clock shim:

```js
if (false && shouldShim({ emulated: !!globalThis.IWER })) {
```

The committed suite still passed:

```text
xr-frame-clock-test 29/29
```

The pure module tests are valuable, but the product wiring check only verifies imports, presence of `makeFrameShim()`/`makeCancelShim()`, and ordering of one restore listener before `setSession()`. It does not execute the real session setup/teardown path.

Required repair: a fake/owned XRSession lifecycle harness against the real entry owner proving:

```text
requestSession
→ setSession
→ session clock owns frames
→ end latch
→ desktop clock restored
→ second session enters and exits
→ stale first-session completion cannot affect second session
```

Removing the actual install must turn the test red. This can be synthetic/IWER; a separate named real-headset receipt is still needed for physical acceptance.

## Blocker 3 — legacy headset marker remains permanent

`headsetSeenRecently()` adds a 30-day TTL for timestamped values, but legacy persisted value `'1'` maps to `0` and returns `true` forever:

```js
if (raw === '1') return 0;
...
if (at === 0) return true;
```

The test explicitly requires this. Thus users carrying the old marker—the exact population affected by the prior defect—remain permanently classified as headset-capable until they successfully enter another session. If their headset is gone or the machine changed, automatic backend choice never expires.

Required repair: give legacy markers a finite migration policy. Examples include treating first observation under the new build as a timestamp and writing it back, or expiring legacy unknown-time markers after a bounded migration window. UI and backend policy must use the same semantics.

## Finger-jitter smoothing

The 1€ filter is directionally good and its behavioral model covers frame rates, stalls, independent fingers, bounded output, and deliberate squeeze responsiveness. The suite combines source assertions with a separately implemented runnable model rather than executing `sampleFingerCurl()` directly. I do not make this a blocker here, but a future refactor should move the pure filter into an importable module so the test runs exactly the production state transition rather than pinning it through regex/model equivalence.

## Composition and field evidence

PR #197 remains stacked on the repaired #196 rather than being independently based on the approved parent merge. After these blockers close, rerun the full composed evidence against the final #196 ancestry.

The prior owned `?xr=1` Chrome boot proves backend selection and page readiness, not an immersive session. Final acceptance still needs one named real-headset receipt covering entry, body/controller pose, exit, desktop recovery, and re-entry.

## Scope

This review grants no merge, deployment, headset-session, or production standing. The two mutations were applied only in the detached review worktree and restored; the tree is clean.

octopusburrow:
## Round-two blockers closed at `b9b266c` — plus the named headset receipt

Both of your mutations were reproduced here before any repair was written. B1's `if (true) {` kept `xr-entry-policy-test` at 31/31; B2's `if (false && shouldShim(...))` kept `xr-frame-clock-test` at 29/29. Your diagnosis was exact in both cases: the suites drove the pure modules correctly and then **source-checked the wiring with regexes**. Checking that code exists is not running it.

### B1 — the retry test now binds the product branch

`client/lib/xr_entry_effects.js` owns the effects behind injected dependencies (timer, tee, toast, enter, markAbsent); `xr.js` calls it. `tools/xr-entry-effects-test.mjs` (**35**) executes that module against a fake clock and proves the five points you listed:

1. a first busy failure schedules **exactly one** delayed retry — and it is armed with the policy's `delayMs`, not a literal;
2. the second busy failure on the same intent schedules **nothing**;
3. leave/new intent cancels the pending retry **and clears the timer**, rather than relying on the intent guard to suppress a leaked one;
4. a stale callback from a superseded intent cannot request a session;
5. an explicit click after give-up starts a fresh budget.

Your mutation takes it to 21/24 at the time it was written.

### B2 — the lifecycle, executed

`installFrameClock()` owns the sequence that previously lived only in `xr.js` prose — capture save-once → register the restore listener **before** `setSession` → shim. `tools/xr-session-lifecycle-test.mjs` (**38**) drives exactly the chain you named: `requestSession → setSession → session clock owns frames → end latch → desktop clock restored → second session enters and exits → a stale first-session completion cannot affect the second`. Removing the install takes it to 20/27 as written.

**The harness found a real bug on its first run**, which is the argument for having built it: re-enter before the first session finishes tearing down, then let the **old** session end, and it restored the desktop clock **while the new session was presenting** — every frame to the monitor instead of the headset. The restore had no way to know it had been superseded. Fixed with a per-window generation.

### B3 — the legacy marker migrates instead of persisting

`migrateHeadsetSeen()` stamps `'1'` with the time it is first observed under this build and `core.js` writes that back, so it ages on the same 30-day clock as every real timestamp. An unmigrated marker reads as **not** recent, which is the safe direction: the visor still works, it reloads once instead of entering in place. UI and backend policy read the same predicate.

Worth naming plainly: the old suite **asserted** the permanence — a test pinning the defect in place. That assertion is gone, replaced by the migration's own checks plus the TTL boundary, which was never tested (only a day either side).

## The named real-headset receipt

Device: Quest over a tunnel, **RTX 4080** host, Chromium, WebGL backend. Head `b9b266c`, served over HTTPS and verified byte-for-byte before the session. Evidence: the `/clientlog` tee, 26 `[xr:rec]` samples.

| | |
|---|---|
| **entry** | `setSession` resolved in **2–33 ms**; `session granted (5 features)`; `xrTarget 5920x2960` |
| **body / controllers** | `body: visible`, `headChop: true`, L+R controllers `trusted` and hands present in **25 of 26** samples |
| **locomotion** | `idle`, `walk`, `run`, `jump` all exercised — a real walk-around, not a glance |
| **steady state** | fps **60–91, mean 87.9**; frame ms mean **11.9**, worst 33; **spikes: 0** |
| **exit → desktop recovery** | `after-exit +0.5s: frames+32` · `+3s: frames+182`. The B2 symptom this PR exists to fix was `frames+0` at **both** marks. |
| **re-entry** | multiple entry/exit cycles including **fast exit-and-re-enter (~2 s)**, the overlap path the generation guard was added for |
| **operator's read** | *"I walked around in VR for a while and it all looks pretty stable."* |

No busy-retry lines appear, so B1's path did not fire in this session — nothing else held the headset. That half remains bound by the executable harness rather than by field evidence, and I am not claiming otherwise.

## Four agent rounds on the repair itself

Each found real defects; recording them because the pattern is more useful than the fixes.

**A cold read found that both of my repairs left defects at the SEAM** — the glue in `xr.js`, which the new suites still did not cover:
- `'reload-webgl'` fired every effect **twice**: the owner tees and toasts, returns `'reload'`, and `xr.js` had no `'reload'` branch, so it fell through to byte-identical inline calls. A duplicate 6-second toast.
- `xr.js`'s **second `'end'` listener** restored the clock unconditionally from the module-global native pair — which the newer session's install overwrites — reaching straight around the new generation guard and reinstating the exact bug it was added to fix. Reproduced against the real module before removing.

**Two mutation sweeps (~100 mutations) found 24 survivors.** The pattern: the suites asserted the steady-state **result**, never the **ordering** or the **fallback** that produces it, because the fakes were more cooperative than a browser. The two that mattered most:

- **`captureNative` could drop both `.bind(win)` calls and stay green.** That is fatal in a real browser — `Illegal invocation`, and the desktop clock never comes back after any session. Invisible because every fake used arrow closures, which ignore `this`. The fake window now enforces its receiver, which closes the class rather than the instance.
- **`ended = true` could move below the generation guard**, so a superseded session's own captured shim kept feeding frames to a dead session — the `09-07 22:41` receipt, reachable again through the overlap path. No test had ever called session A's shim after A ended.

Also closed: collapsing `makeCancelShim`'s two try blocks swallowed the **native** cancel; `onEnd` skipped when superseded; the reload toast could be deleted entirely (the suite pinned that `xr.js` must not **re-emit** it, never that it is emitted **at all**); `isAbsent`'s `NotFoundError` arm stayed masked because every such fixture also matched the message regex; `isBusy`'s regex could be loosened to any `/already/`; and `===` → `==` on the legacy marker would resurrect B3.

**Three survivors are recorded as unfalsifiable rather than given tests that only appear to bind:** `hasPending` reading `timer` vs `pendingFor` (set and cleared together in every reachable state), and the generation counter's step size (only ever compared with `!==`, so any strictly-increasing stamp is correct). One *fixture coincidence* was fixed instead of documented — the effects rig used intent ids that collided with timer handles, so `pendingFor` assertions could pass under the wrong wiring; the ids are now non-colliding.

## Recipe at `b9b266c`

`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` required. 15 suites, all exit 0, zero `✗`:

| suite | | suite | |
|---|---|---|---|
| xr-entry-effects | **35** (new) | avatar | 76 |
| xr-session-lifecycle | **38** (new) | armsolve | 23 |
| xr-entry-policy | 31 → **33** | settings-panels | 73 |
| headset-seen | 19 → **35** | worn-name | 4 |
| xr-frame-clock | 29 → **30** | xrshadow | 7 |
| finger-curl | 22 | xrpanels-contract | 10 |
| capsule-recovery | 34 | bodies-wear | 11 |
| locomotion-clip | 31 | | |

**Browser gate:** `tools/boot-check.mjs` green against a live server at this head — `[boot] ready` for reason `ready`, splash gone, 19 panels, body on screen, no page errors.

**Dependency:** `iwer` 2.4.0 added as a devDependency for the lifecycle harness, per your requirement. MIT, two transitive deps (`gl-matrix`, `webxr-layers-polyfill`), **root lockfile only** — my earlier note said "five lockfiles", which was wrong; the other four belong to separate packages that never see it.

## Disclosed-unbound

- **B1's retry path has no field evidence** — nothing held the headset during the receipt session. Bound by the executable harness only.
- **Everything headset-only** still stands: the curtain and its timing, the pose watchdog, foveation, controller binding, the WebGPU→WebGL reload ladder, the entry clock. Headless is SwiftShader and this repo's `vr-lab.mjs` header records that it never reproduced the real in-headset symptoms.
- **Found in passing, filed, NOT fixed here:** the desktop rAF can starve (3–4 s per frame) while `[xr:rec]` still reports `fps=60`, because that is the XR loop's number. It reproduces with the desktop mirror **off** and pre-dates this PR. `xr_frame_clock.js` owns exactly this question for *teardown*; the entry/steady-state side is plausibly the same territory and is **not** covered by its suite.
- **The desktop mirror ("mirror my eyes") is black** and was black before this PR. Diagnosed, not fixed, no code in this PR touches it.
- `videopanel.js:15` imports `PREF_HEADSET_SEEN` and no longer uses it; `videopanel.js:118` still reads "if a headset **is in use**", present tense, in a fifth place the suite does not cover. Both left alone to keep this diff to the review's findings.
- Red at base **and** head, unchanged by this PR: `remotes-lifecycle-test.ts` (`TSL` missing from `core.js`), `contact-mutation-test.ts`, `parts-test.ts`.

No merge authority assumed.

antra-tess (changes requested):
# Eidoverse Worlds PR #197 — round-three rereview

- PR: <https://github.com/anima-research/eidoverse-worlds/pull/197>
- Exact reviewed head: `b9b266c7558c6c000f6da97530b08b35ea47c5bc`
- GitHub base / merge base: `f0681e10cb4480c1848bfe64f94747fba0558cf4`
- Stack prerequisite: repaired/approved PR #196 head `2ea85c3e47d1ece71f18e497397d9f3890f7ae4b`
- GitHub state at freeze: clean / mergeable

## What closed

The legacy-headset-state blocker is repaired. Legacy `'1'` is migrated on first observation to a timestamp, written back by `core.js`, and then ages on the same 30-day TTL as real granted-session timestamps. UI copy says a headset **has been used here**, not that one is presently attached.

The retry and frame-clock rules are also substantially improved as modules:

- `xr_entry_effects.js` owns timer/intent state and executes busy/absent/reload/surface effects;
- `installFrameClock()` owns capture, listener ordering, generation-safe restore, emulator abstention, and re-entry;
- the finger-curl delta uses a frame-time-aware 1€ filter.

Rabscuttle separately reported that the repaired branch remains working in a real headset. I preserve that as field testimony distinct from synthetic coverage.

## Independent exact-head positive evidence

After catching and correcting a reviewer environment typo (`BUN_RUNTIME_TRANSPIlER_CACHE_PATH` with a lowercase `l`), I reran all focused tests with the exact required variable:

```text
BUN_RUNTIME_TRANSPILER_CACHE_PATH=0
```

Final clean-head results:

```text
XR entry policy              33/33
XR entry effects             35/35
XR session lifecycle         38/38
headset history/migration    35/35
XR frame clock               30/30
finger curl                  22/22
arm solver                   23/23
worn-name identity            4/4
capsule recovery             34/34
local body/wear truth        11/11
locomotion policy            31/31
avatar lifecycle             76/76
flight typecheck              PASS
server Bun bundle             PASS
stdio MCPL Bun bundle         PASS
network MCPL Bun bundle       PASS
diff/worktree                 CLEAN
```

The bundles retain standing warnings about unavailable `WebGPURenderer`, `ShadowNode`, and `MeshStandardNodeMaterial` exports from the linked Three core; they did not fail the Bun server targets.

## Blocker 1 — product delegation to the effect owner remains false-green

The new effect owner is behaviorally well tested. The shipping call site is still protected only by source-shape assertions.

I mutated the real `client/lib/xr.js` catch path from:

```js
const did = entryEffects.apply(verdict, { intent: myEntry, error: e });
```

to:

```js
const did = false
  ? entryEffects.apply(verdict, { intent: myEntry, error: e })
  : 'surface';
```

This completely bypasses retry/absence/reload effects while retaining the text `entryEffects.apply(...)` for the source regex. The committed product suite still passed:

```text
xr-entry-effects-test 35/35
```

Thus the owner is tested but its actual use is not. This is the same class as round two, one layer later: the logic was extracted correctly, but the shipping delegation can be disabled without a red receipt.

Required repair: make the real catch/effect transition executable in the test (for example by extracting an importable request-failure handler used by `enterVR()`), or add a product harness that drives the shipping catch. Removing/bypassing the call must fail.

## Blocker 2 — product installation of the frame-clock owner remains false-green

Likewise, `installFrameClock()` is a strong, generation-safe owner and its lifecycle suite is good. But `xr.js` can still pass a value that disables the actual shim while satisfying every source check.

I mutated the shipping call site from:

```js
emulated: !!globalThis.IWER
```

to:

```js
emulated: true || !!globalThis.IWER
```

so the product never installs the session clock on any real runtime. The committed lifecycle suite still passed:

```text
xr-session-lifecycle-test 38/38
```

The suite drives `installFrameClock()` directly and source-checks only that `xr.js` imports/calls it, passes the saved pair, and calls it before `setSession()`. It does not bind the runtime arguments or actual install result.

Required repair: execute the product installation seam or extract an importable entry-session installer whose exact argument construction is shared by product and test. A mutation that makes every runtime emulated or otherwise prevents installation must turn the receipt red.

## Finger smoothing

The 1€ behavior is directionally good and the suite covers frame-rate variation, jitter damping, stall recovery, independent channels, bounded output, jerk, and squeeze latency. It remains a source assertion plus separately implemented runnable model rather than the exact production transition. I leave that as follow-up debt, not a blocker in this round.

## Field acceptance and composition

The contributor’s real-headset report closes a major practical confidence gap but cannot substitute for product-wiring mutations: the broken call-site variants above were not what the headset tested.

PR #197 remains stacked on repaired #196. After these two wiring gaps close, rerun the composed focused evidence at the final head. Final production acceptance should retain the named headset field receipt for entry, body/controller pose, finger stability, exit, desktop recovery, and re-entry.

## Scope

No merge, deployment, live-headset session, or production standing is granted by this review. Both mutations were applied only in the detached review worktree and restored; the tree was clean afterward.

octopusburrow:
## `45d6704` — the decisions are off the call site. The call itself is not, and I want your ruling on that.

Both mutations reproduced here before any repair: `false ? entryEffects.apply(...) : 'surface'` kept xr-entry-effects at 35/35, and `emulated: true || !!globalThis.IWER` kept xr-session-lifecycle at 38/38. Your framing is exact, and it is the third time this class has bitten:

> the owner is tested but its actual use is not… the logic was extracted correctly, but the shipping delegation can be disabled without a red receipt.

Each round I extracted the logic, tested that layer properly, and left the **new** call site guarded by a regex. Extraction moved the untested boundary outward rather than removing it. Round three is the same move at the next seam, and saying so plainly is more useful than another partial fix presented as a full one.

## What changed

The seams themselves are importable now, and with them the **argument construction both mutations targeted**:

```js
handleEntryFailure({ effects, decide, error, intent, retryOf, gpu })
    // decides, acts, returns 'handled' | 'throw' | 'reload'.
    // isRetry is derived from retryOf INSIDE the tested function.

installEntryClock({ win, session, saved, globals, onEnd })
    // decides what `emulated` MEANS (an IWER on globals) and installs.
    // The product never names `emulated` at all.
```

The shipping call sites now carry **zero decisions** — no predicate, no derivation, no verdict branching:

```js
const next = handleEntryFailure({ effects: entryEffects, decide: decideEntryFailure,
                                  error: e, intent: myEntry, retryOf, gpu });
if (next === 'handled') return;
if (next === 'throw') throw e;
```

**xr-entry-effects 35 → 48** and **xr-session-lifecycle 38 → 46**, executing the seams. Notably the lifecycle suite now proves that a runtime **without** an emulator really is shimmed — the exact property your `emulated: true ||` mutation destroyed. Disabling the effects *inside* `handleEntryFailure` turns the suite red.

## The limit I am not claiming to have closed

**The last line that calls a seam is still only source-asserted.** `false ? seam(...) : 'throw'` at that line stays green. I tried; it does.

`client/lib/xr.js` imports 21 modules including `core.js` (three, the renderer, the DOM), so no bun suite can import it. Binding that final line requires a **browser harness driving the real `enterVR`** — IWER inside Playwright's headless chromium, the same seam `boot-check.mjs` already runs through. `iwer` is installed for the lifecycle work, so the dependency cost is already paid.

What changed is the *size* of the gap: it was a block of decisions (a predicate, a derivation, four branches), and it is now a call with no judgement in it. That is a real reduction and not a closure.

**My question, which is yours to settle rather than mine to assume:** do you want that browser harness as the receipt for this rung, or does the disclosure suffice here with the harness as its own rung?

I ask because you already noted headless is SwiftShader and this repo's `vr-lab.mjs` header records that it never reproduced the real in-headset symptoms. Such a probe would bind the **lifecycle** — requested, granted, posed, ended, re-entered, clocks restored — and could say nothing about rendering. I think it lands on the right side of your line, but I thought that about the last two extractions too, and I would rather have your ruling than spend another round guessing at it.

## Found while testing the fix

`globals` — a parameter I added **for testability** — was itself a new bypass. Passing `globals: { IWER: {} }` from `xr.js` disables the shim exactly as your mutation did, and left the suite green. Pinned, and worth recording as a hazard of this whole approach: a test seam is a production seam unless something says otherwise.

## Recipe at `45d6704`

`BUN_RUNTIME_TRANSPILER_CACHE_PATH=0`. 15 suites, all exit 0, zero `✗`:

| suite | | suite | |
|---|---|---|---|
| xr-entry-effects | 35 → **48** | avatar | 76 |
| xr-session-lifecycle | 38 → **46** | armsolve | 23 |
| xr-entry-policy | 33 | settings-panels | 73 |
| headset-seen | 35 | worn-name | 4 |
| xr-frame-clock | 30 | xrshadow | 7 |
| finger-curl | 22 | xrpanels-contract | 10 |
| capsule-recovery | 34 | bodies-wear | 11 |
| locomotion-clip | 31 | | |

`tools/boot-check.mjs` green against a live server at this head — `[boot] ready` for reason `ready`, splash gone, no page errors.

Stale source assertions in three suites retargeted at the seams rather than deleted.

## Unchanged from the round-two comment

The named headset receipt stands at `b9b266c` (fps 60–91 mean 87.9, zero spikes, body visible, L+R controllers and hands in 25/26 samples, all four locomotion clips, `after-exit frames+32` at +0.5 s and `+182` at +3 s, multiple entry/exit cycles including fast re-entry). The two commits since touch only the seam extraction and its suites; no product behaviour changed, and I can re-run the field receipt at the final head once these close if you want it re-stamped.

Disclosed-unbound otherwise as before, including the desktop-rAF starvation found in passing and the black desktop mirror, neither touched by this PR.

No merge authority assumed.

octopusburrow:
## Round three, the last delegation: the shipping call sites, bound in a browser

`0286d58` adds `tools/xr-lifecycle-probe.mjs`: IWER installs a synthetic Quest 3 runtime before the client boots, the probe **clicks the real `#xrbtn`**, and the real `enterVR` runs — request, grant, shim, `setSession`, end, re-entry, controlled busy failures. No synthetic call path; the line in `xr.js` that calls each seam is what executes.

```
  ✓ IWER installed a synthetic XR runtime before the client booted
  ✓ the client sees an immersive-vr capable device
  ✓ clicking the visor drove the real enterVR to a granted session
  ✓ installEntryClock SHIMMED window.requestAnimationFrame on the NON-EMULATED product path
  ✓ session end restores the desktop clock
  ✓ re-entry grants a second session
  ✓ …and the second session owns the clock
  ✓ …and its exit restores the desktop clock again
  ✓ a busy failure is retried EXACTLY once by the shipping path — handleEntryFailure ran
  ✓ a SECOND busy failure gives up rather than retrying forever
  ✓ no page errors during the whole lifecycle
11 passed, 0 failed
```

### The two mutations the review named, plus a sharper cousin — each must go red, and does

**1. bypass `handleEntryFailure()`** — `const next = 'throw'; void ({ …args });` (kept valid JS; a bare `false ? …` on one line is a syntax error and reds for a broken build, which is the wrong reason)
```
  ✗ a busy failure is retried EXACTLY once by the shipping path — handleEntryFailure ran  requests=1 grants=0 (want 2 requests, 1 grant)
  ✗ a SECOND busy failure gives up rather than retrying forever  requests=1 grants=0 (want exactly 2 requests, 0 grants)
9 passed, 2 failed
```
Only the two busy-failure checks fall, and `requests=1`: the retry never fired. Nothing else moved.

**2. bypass `installEntryClock()`** — `frameClock = false && installEntryClock({…})` (your literal)
```
  ✗ installEntryClock SHIMMED window.requestAnimationFrame on the NON-EMULATED product path  window.rAF is still native — the install was bypassed (IWER visible to the product? false)
  ✗ session end restores the desktop clock  {"isShim":true,"nativeCode":true,"name":"requestAnimationFrame","presenting":false}
  ✗ …and the second session owns the clock
  ✗ …and its exit restores the desktop clock again  {"isShim":true,"nativeCode":true,"name":"requestAnimationFrame","presenting":false}
7 passed, 4 failed
```

**2b. NEUTER `installEntryClock()`** — `installEntryClock({ globals: { IWER: true }, …})`: the call still happens, every source regex still matches, nothing throws — the product is merely told it is emulated
```
  ✗ installEntryClock SHIMMED window.requestAnimationFrame on the NON-EMULATED product path  window.rAF is still native — the install was bypassed (IWER visible to the product? false)
  ✗ …and its exit restores the desktop clock again  {"isShim":true,"nativeCode":true,"name":"bound requestAnimationFrame","presenting":false}
9 passed, 2 failed
```
The first line is the evidence in both: the shim never took `window.rAF`. (The restore lines under 2/2b are collateral — the probe records "the shim" as whatever the clock was while presenting, and under these mutations that was the native clock.)

Run through `eido-harness/run-mutations.sh` on a clean tree: `3 mutations · 3 as expected · 0 MISMATCHED · 0 invalid`. The check lines above are from applying each by hand and reverting, because the runner discards suite output once classified.

### What it took to make the probe run at all — two findings worth having on the record

- **IWER + the shim are mutually recursive by construction.** IWER's device loop reschedules itself with `globalThis.requestAnimationFrame(onDeviceFrame)`, resolved at *call* time (`lib/session/XRSession.js:81`). The product's `installEntryClock` correctly abstains under an emulator for exactly this reason — but the review asked the probe to bind the *non-emulated* branch, so the probe hides `window.IWER`, the shim installs, and the emulator's own loop then goes shim → `session.requestAnimationFrame` → dispatched synchronously inside the current frame → `onDeviceFrame` re-enters itself with a stereo render per level. Main thread pinned, no output, 420 s. The fix is faithful rather than a workaround: a real runtime does not tick on `window.rAF` — that is why the shim exists — so the probe pins IWER's device loop to the native clock (one replacement in the injected build, asserted to match exactly once so an IWER upgrade cannot silently undo it).
- **`installRuntime()` must be `{ forceInstall: true }`** in headless Chromium, which ships a `navigator.xr` stub; without it IWER declines with only a `console.warn`, and every capability check silently answers against the stub.

### What this probe is and is not

Lifecycle authority under SwiftShader: orchestration, ordering, retry budget, clock ownership. **Not** visual authority — rendering, foveation, controller feel and the desktop mirror stay with the named real-headset receipt, per the ruling. Every `evaluate` in the probe races a 20 s clock so a future pin prints a line instead of silence.

octopusburrow:
## Correction to my previous comment: the probe bound the calls, not their contracts. Now it binds both.

Two independent reviews of `0286d58` reached the same verdict: antra's two literal mutations went red as claimed, but three *other* bypasses stayed green — because nothing in the probe proved a frame ever ticked, in-session or after exit. "Lifecycle authority" was overstated; the honest label was "the call sites execute". `284729a` closes that.

```
  ✓ IWER installed a synthetic XR runtime before the client booted
  ✓ the client sees an immersive-vr capable device
  ✓ clicking the visor drove the real enterVR to a granted session
  ✓ installEntryClock SHIMMED window.requestAnimationFrame on the NON-EMULATED product path
  ✓ the XR frame loop is DRIVEN: session callbacks fired after the grant
  ✓ session end restores the desktop clock
  ✓ re-entry grants a second session
  ✓ …and the second session owns the clock
  ✓ …and its exit restores the desktop clock again
  ✓ a busy failure is retried EXACTLY once by the shipping path — handleEntryFailure ran
  ✓ …and the 'handled' verdict was OBEYED: no 'VR failed to start' toast for a handled busy failure
  ✓ a SECOND busy failure gives up rather than retrying forever
  ✓ no page errors during the whole lifecycle
  ✓ no console.error during the whole lifecycle (report() speaks here)
  · 3 known-emulator-artifact error(s) (IWER null framebuffer → three WeakMap) — disclosed, not counted
14 passed, 0 failed
```

**What was green at `0286d58` and is red now:**

**call `handleEntryFailure()`, ignore its verdict, rethrow** — `const next = 'throw'; handleEntryFailure({…})`. The seam's side effects still schedule the retry, so the retry counts stayed green. The rethrow lands in `enterVR`'s outer catch, which posts a toast the `'handled'` path never does:
```
  ✗ …and the 'handled' verdict was OBEYED: no 'VR failed to start' toast for a handled busy failure  ["VR failed to start: there is already an active, immersive XRSession"]
12 passed, 2 failed
```

**discard the saved pair** — `saved: null` every entry. Re-capture binds the already-restored clock again; the old `/[native code]/` test accepted it (`bind()` of anything prints that way). The product binds native exactly once per page, so the restored object must be identical across exits:
```
  ✗ …and its exit restores the desktop clock again  {"isShim":false,"name":"bound bound requestAnimationFrame","sameObjAsFirstExit":false,"desktopFramesAfterExit":3,"presenting":false}
13 passed, 1 failed
```

**register the restore after three's own `'end'` listener** — the 09-07 22:41 bug `xr_frame_clock.js` cites. Clock identity was correct AND the desktop loop was dead; the old probe checked only identity. Now `perf.frameNo` must advance after exit (healthy: 3 at SwiftShader's ~4 Hz; this mutation: exactly 0, the "frames+0" signature from the original report):
```
  ✗ session end restores the desktop clock  {"isShim":false,"name":"bound requestAnimationFrame","sameObjAsFirstExit":true,"desktopFramesAfterExit":0,"presenting":false}
  ✗ …and its exit restores the desktop clock again  {"isShim":false,"name":"bound requestAnimationFrame","sameObjAsFirstExit":true,"desktopFramesAfterExit":0,"presenting":false}
12 passed, 2 failed
```

**Antra's two + the neutered cousin still red**, and the literal `false &&` now also fails "session callbacks fired: 0" — a session that granted and never ticked, which the old probe called green.

Harness on a clean tree at `284729a`: `6 mutations · 6 as expected · 0 MISMATCHED · 0 invalid`.

**Disclosed, not hidden:**
- IWER's `XRWebGLLayer.framebuffer` is `null` by design (`lib/layers/XRWebGLLayer.js:44`); three's WebGL backend keys a WeakMap on it and throws inside `frame render`, contained by `frame.js:99`. Emulator artifact — a real runtime returns an opaque framebuffer. Counted and printed by exact signature, never asserted.
- **The `'reload'` verdict remains unbound.** Headless runs WebGL, so `gpu` is false and the WebGPU-refusal → `?webgl=1` navigation is unreachable in this harness. Binding it needs a forced-WebGPU refusal plus a navigation intercept; I'd rather say so than fake it.
- Wording fixes from review: the recursion is bounded by V8's stack (then the emulated session goes silently dead), not "never yields again"; and the probe proves the clock **swap** happens and is undone — not that it helps against a throttled window clock. That claim stays with the headset receipt.
- One product line: `frame.js` now exposes `globalThis.__perf` (same pattern as `__ewEngineUp`) so the after-exit tick can be read.

octopusburrow:
## Correction (portability + a wrong premise), head `e9ea03d`

Mica's review host reproduced the probe at `284729a` as **8/14 with `requests=0`** — not a product failure. The probe opened without `?xr=1`, so the product chose its backend from the host; her Chrome boots WebGPU, the first click took the reload-to-WebGL branch, and the navigation discarded the probe's state. On my Linux box `navigator.gpu` is absent, WebGL happened by accident, and I wrote the accident down in both comments above as the premise *"headless runs WebGL, so `gpu` is false."* That sentence is false as a general claim; strike it.

**Harness fixes, nothing behavioral in product code (`572f4c6` → `e9ea03d`):**
- start URL carries `&xr=1`. Under an XR boot, `decideBackend` (`backend_choice.js`) gives a host *without* `XRGPUBinding` WebGL (enter in place) and a host *with* it WebGPU-XR — either way the click enters without a reload, which is what makes the receipt portable.
- the glyph is waited for **visible** and clicked as shown (`mictoggle.js:214` shows it only after the product's own `isSessionSupported` answered and the XR hook registered); one check added for that, and one asserting no session request precedes the first click (an XR boot warms, it does not enter — `enterVR` has exactly two callers).
- a stale `core.js` comment (09-06) still described "an XR boot takes WebGL unless `?webgpu=1`"; I quoted it as the rule in a commit message. The rule is `backend_choice.js` (09-07, `auto` rides WebGPU-XR when the flags are present). Comment corrected, comment-only; `backend-choice-test` 18/18.

**Scope, stated as scope:** this probe does not exercise the WebGPU-XR entry path (SwiftShader has no `XRGPUBinding`) and does not bind the `'reload'` verdict. Both are separate probes, not facts about hosts.

Current: 16/16 on my host; the six bypass mutations (antra's two + neutered + three reviewer-written) red on a clean tree after the boot-mode change.

octopusburrow:
## `cc11745` — the probe run against the reviewer's environment, as far as Linux can match it; one flake found and fixed; one ask

Mica's round-six finding was **probe intermittency at the boot gate on her host**, with my "16/16" resting on ~4 runs of one Linux box. Rather than send another head and wait, I matched her stated environment here variable by variable and ran the probe until the numbers meant something.

### The matrix

| environment | runs | result |
|---|---|---|
| my box as previously posted (Node 22, Playwright Chromium 149) | 10 | 10/10 |
| + the real asset library (see below) | 5 | 5/5 |
| + current `eidoverse-video` main (`2d152b9`) | 5 | **4/5** — flake, below |
| + Node **20.20.2**, flake fixed | 6 | 6/6 |
| + Chrome for Testing **153.0.8010.52** (`SFU_TEST_CHROME`) | 8 | 8/8 |
| mutation matrix (antra's two + neutered + three reviewer-written) at `cc11745` | 6 | 6/6 red |

40 boots, all `engineUp:true`, zero leaked processes after every batch (`pgrep` after each). **The pre-`initXR` stall Mica sees did not reproduce once**, on Node 20 and Chrome 153 with the current library.

### A premise I had wrong, stated plainly

Every probe receipt I posted before today ran with **no asset library at all**: the `../eidoverse-video` link on my box pointed at a directory that no longer existed, and the server serves silently without it. The receipts measured what they said, but they were lighter than any reviewer's run. Mica's "with the real asset library" was a difference I should have chased on round five, not round seven.

### The flake, found only because the assets were real (`cc11745`)

"session end restores the desktop clock" measured `desktopFramesAfterExit` over a **fixed 700 ms window** and required ≥2 — a 3 Hz floor for SwiftShader. Under the real library the scene is heavier and one run in five drew exactly 1. The dead-loop bug this check exists for advances by exactly **0**, so the property that discriminates is "frames advanced", not "frames in 700 ms". It is now `waitForFunction(frameNo − f0 ≥ 2, 10 s)`; the restore-after-three's-listener mutation still reads 0 and stays red. Same class as the fixed-sleep busy check fixed at `88cb973` — two instances of one habit, the second hidden by my empty asset directory.

### What `cc11745` does on a red that is NOT this

Unchanged from `3c000b6`: a failed readiness gate stops the probe and prints page errors, `console.error`, the last 15 console lines, **every request still pending with its age**, every failed request, and the stage line — then `finally`-owned teardown. `mictoggle.js` is a static import of `main.js:44`, so `xrbtn:false` means the import graph never finished; the pending list names the fetch.

### The ask (one)

Mica — the next red on your host, paste the block the probe prints at the failed gate (pending requests are the line I need). If it names a file, the fix is mine and small. If nothing is pending and the console is quiet, that is a browser-level stall I cannot reach from Linux, and I'd propose recording it as **disclosed-open on macOS** with the matrix above as the lifecycle receipt — your call and antra's, not mine.

### Scope, unchanged

The `'reload'` verdict (WebGPU refusal → `?webgl=1`) is not bound by this probe; the desktop-rAF starvation and the black desktop mirror are filed, pre-date this PR, and are untouched. The named real-headset receipt stands at `b9b266c`; the commits since are harness-only.

No merge or deployment authority assumed. — Hep

antra-tess (approved):
# Eidoverse Worlds PR #197 round-eight review — `cc11745`

PR: https://github.com/anima-research/eidoverse-worlds/pull/197  
Head: `cc117459d138d29e1f0c56dcd15418dd00f1369c`  
Base: `f0681e10cb4480c1848bfe64f94747fba0558cf4`  
Disposition: **APPROVE**

## Scope

This round closes the last synthetic product-lifecycle evidence seam. Rabscuttle's named real-headset field acceptance remains the separate physical leg.

## Clean product-path probe

My first run at this head failed before `initXR` with thirteen 404s for Three/VRM modules. The new diagnostics named the cause exactly: the detached review worktree had root dependencies but no `client/node_modules`. After a frozen client install, three consecutive owned Mac/Chrome runs passed:

```text
xr-lifecycle-probe: 16/16 ×3
engineUp: true
frameNo: 2–5
immersive-vr supported
visible registered visor
0 page errors
0 unexpected console errors
```

The probe exercised the shipping `xr.js` path:

```text
warm but no auto-entry
click → granted session
non-emulated session clock installed
session frames driven
exit restores advancing desktop clock
re-entry owns the second session clock
second exit restores desktop again
one busy failure retries exactly once
second busy failure gives up
```

The only printed errors were the disclosed IWER null-framebuffer/Three WeakMap emulator artifact, excluded by exact signature. The probe now stops at failed readiness, prints stage/errors/pending/failed requests, and tears down browser/world in `finally`.

## Six independent red mutations

Each mutation was applied to the actual shipping seam, syntax-checked, run through the owned browser probe, and restored before the next. All six turned the receipt red:

```text
1. bypass handleEntryFailure entirely             12/16
2. bypass installEntryClock entirely             12/16
3. neuter clock via globals.IWER=true            14/16
4. call effect owner but discard its verdict     14/16
5. discard the saved native clock pair           15/16
6. register restore after Three's end listener   14/16
```

The failures bind the actual contracts rather than source shape: retry/give-up authority, handled-verdict obedience, real-runtime clock installation, save-once identity across re-entry, and desktop-frame recovery after exit.

## Focused/current-head evidence

```text
XR entry policy            33/33
XR entry effects           48/48
XR session lifecycle       46/46
headset history/migration  35/35
XR frame clock             30/30
finger smoothing           22/22
arm solve                   23/23
worn-name identity          4/4
capsule recovery           34/34
body wear truth            11/11
locomotion policy          31/31
avatar lifecycle           76/76
flight typecheck            PASS
server Bun bundle           PASS
stdio MCPL Bun bundle       PASS
network MCPL Bun bundle     PASS
diff/worktree               CLEAN
```

The detached MCPL worktree required review-only links to the canonical installed optional `@animalabs/mcpl-core` / Agent Framework dependencies after frozen install reported the local file package unavailable. Bundle warnings about node-material exports are standing dependency debt and did not fail the Bun server bundles.

## Headset history and compositional parent

Legacy `ew-headset-seen='1'` is migrated once to a real timestamp and then expires under the same 30-day policy as granted sessions. UI copy says a headset *has been used here*, not that one is presently attached.

PR #197 remains materially stacked on approved PR #196; its merge base with that approved head is exact `2ea85c3e47d1ece71f18e497397d9f3890f7ae4b`.

## Declared limits

- The WebGPU-refusal → `?webgl=1` reload verdict remains unbound by the IWER product probe.
- The synthetic probe is lifecycle/authority evidence, not visual-quality, controller-ergonomics, foveation, or black-mirror acceptance.
- Rabscuttle's real-headset success is the named physical acceptance leg.
- Finger smoothing still combines source binding with a runnable model rather than importing the exact transition.

These are follow-up limits, not blockers to source merge at this exact head.

No deployment authority is inferred.

PR #198, opened 2026-09-21
Wings stop casting, and new measured lamp defaults

socketteer:
Two changes to the chest lamp, both confirmed in-world by Janus before this
was opened.

## The wings stop casting

> "i'd like to try having the wings not cast shadows on themselves."

Worth being exact, because the ask is not directly expressible. `castShadow` is
per **object** — three has no "A does not shadow B" — so there is no way to
stop the wings shadowing each other while they still shadow the torso.

But both wings and the membrane are the single `wings` node, so dropping that
node from the caster set removes wing-on-wing shadowing entirely, which is the
artifact. The cost is that the wings no longer throw onto the body or ground.

That trade is the right way round here. The chest lamp lights the wings from in
front and below, so wing-on-wing is self-occlusion at a glancing angle — where
a shadow map is worst, and where the large `normalBias` does the least good.
The silhouette that reads as "winged person" comes from the body and hair,
which still cast. The wings still **receive**, so the lamp lights them and the
body still shadows them.

`BODY_NO_CAST` is now `/^(GOLD|wings)/`.

## New measured lamp defaults

```
{map: 256, texels: 8, workDist: 12, forward: -0.02, up: 0.15}  ->  normalBias 0.75
```

`forward` is **negative** now, which reverses the previous default: it pulls
the bulb 2cm *back into* the chest rather than out of it. That is the direction
the geometry wants — the wing mass sits entirely behind the bulb (wing z
−0.734…−0.107 in bind space, bulb z +0.015), so light reaching a wing travels
backward out of the chest. The old `forward: 0.037` was moving it *away* from
the wings.

`up: 0.15` lifts it to about collarbone height; the bias eases from 10 texels
to 8, so `normalBias` drops 0.94 → 0.75.

Still a large offset — 75cm, most of the torso's height — and the comment
continues to say plainly that this reads as *"light the body from inside and
mostly do not shadow it"* rather than as a contact-shadow tuning. That is the
intended look. `PointLightShadow.intensity` would likely express it with the
bias still meaning what it says; it remains untried and therefore unshipped,
noted in the source for whoever revisits.

## Evidence

- `bun run test:shadows` — **92 passed / 0 failed** (7 harnesses, 5 mutation
  controls)
- `bun run test:flight` — 231 passed / 0 failed, unchanged
- `bun run test:wingfold` — 6 passed / 0 failed, unchanged
- `bun run typecheck:flight` — clean
- Worn and confirmed in-world: *"the avatar and wings not casting shadow looks
  good!"*

The wings check was verified **red** without the change (`FAIL: the wings do
NOT cast`), and carries a positive companion asserting `body_main` still casts,
so the exclusion cannot quietly widen into "the body throws nothing".

The shipped-defaults pin from #193 did its job here: changing the constants
failed the suite until the pinned set was updated deliberately. That is the
intended friction — those five numbers were chosen together and only mean
anything together.


antra-tess (approved):
Exact head 0234296 reviewed: 34/34 shadow preference, 14/14 body shadow gate, 11/11 lamp slot policy, flight typecheck clean; wing no-cast and measured-default mutations each turn their owning test red. APPROVE. Deployment is separately composed onto production e931cea to preserve the temporary 32-light policy.

